PR #10, opened 2018-03-24
Fixes tests to conform to Swift bool naming

machados:
Motivation:

Tests fail due to a change in swift-nio which renames
`fulfilled` to `isFulfilled` (1e7410aa).

Modifications:

Renamed the property

Result:

Tests will run and pass.

### DIFF
--- a/Tests/NIOOpenSSLTests/OpenSSLIntegrationTest.swift
+++ b/Tests/NIOOpenSSLTests/OpenSSLIntegrationTest.swift
@@ -487,7 +487,7 @@ class OpenSSLIntegrationTest: XCTestCase {
         serverClosed = true
 
         for promise in promises {
-            XCTAssert(promise.futureResult.fulfilled)
+            XCTAssert(promise.futureResult.isFulfilled)
         }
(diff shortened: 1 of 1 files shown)


machados:
I'm not sure I fully understand. I don't have much know how of the Swift PM/toolchain but my basic understanding is that someone building an application with this library will have swift-nio-ssl as a dependency and get swift-nio and other dependencies by resolution? As long as the dependency versions are compatible with each version of swift-nio-ssl then everything should work?

Lukasa:
@machados Correct, but right now this version of `swift-nio-ssl` provides a constraint on the swift-nio version to be at least 1.0.0. If we make this change, we will become incompatible with 1.0.0 (at least for testing purposes). I'd rather not do that, and the way we can avoid doing that is by using a different function. That allows users to grab a newer version of swift-nio-ssl without needing to rev the version of swift-nio, which could potentially force an inconsistent upgrade.

machados:
Ok no worries, feel free to close this since it will take a while before I'm be able to get to it.

PR #11, opened 2018-03-26
Remove @testable use of NIO.

Lukasa:
Motivation:

When this was in-tree with NIO we could safely use @testable imports to
get at the project internals, becuase we couldn't possibly regress against
them. This is no longer true, so we should avoid @testable imports.

Modifications:

Removed all @testable imports.

Changed NIO.Posix.open to Darwin/Glibc.open.

Replaced uses of fulfilled with either wait() or callbacks.

Result:

Less chance of randomly breaking due to NIO changes.

### DIFF
--- a/Tests/NIOOpenSSLTests/OpenSSLIntegrationTest.swift
+++ b/Tests/NIOOpenSSLTests/OpenSSLIntegrationTest.swift
@@ -14,7 +14,7 @@
 
 import XCTest
 import CNIOOpenSSL
-@testable import NIO
+import NIO
 @testable import NIOOpenSSL
 import NIOTLS
 import class Foundation.Process
@@ -298,8 +298,8 @@ class OpenSSLIntegrationTest: XCTestCase {
         guard let fileName = fileName else {
             fatalError("couldn't make temp file")
         }
-        let tempFile = try fileName.withCString { ptr in
-            return try Posix.open(file: ptr, oFlag: O_RDWR | O_CREAT | O_TRUNC | O_CLOEXEC, mode: 0o644)
+        let tempFile = fileName.withCString { ptr in
+            return open(ptr, O_RDWR | O_CREAT | O_TRUNC | O_CLOEXEC, 0o644)
         }
         precondition(tempFile > 1, String(cString: strerror(errno)))
         let fileBio = BIO_new_fp(fdopen(tempFile, "w+"), BIO_CLOSE)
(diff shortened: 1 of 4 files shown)


Tests/NIOOpenSSLTests/OpenSSLIntegrationTest.swift
```diff
@@ -14,11 +14,17 @@
 
 import XCTest
 import CNIOOpenSSL
-@testable import NIO
+import NIO
 @testable import NIOOpenSSL
 import NIOTLS
 import class Foundation.Process
 
+#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
```
weissi: I believe this shouldn't be necessary as the `NIO` should export them, no?

normanmaurer:
@Lukasa typo in the commit message: `becuase` -> because`

Tests/NIOOpenSSLTests/OpenSSLIntegrationTest.swift
```diff
@@ -298,8 +304,8 @@ class OpenSSLIntegrationTest: XCTestCase {
         guard let fileName = fileName else {
             fatalError("couldn't make temp file")
         }
-        let tempFile = try fileName.withCString { ptr in
-            return try Posix.open(file: ptr, oFlag: O_RDWR | O_CREAT | O_TRUNC | O_CLOEXEC, mode: 0o644)
+        let tempFile = fileName.withCString { ptr in
```
weissi: should we make the `Posix` methods `public`?

Lukasa:
Cool, that's fixed.

Tests/NIOOpenSSLTests/OpenSSLIntegrationTest.swift
```diff
@@ -754,10 +766,15 @@ class OpenSSLIntegrationTest: XCTestCase {
         // synchronization: all of this should succeed synchronously. If it doesn't, that's
         // a bug too!
         let closePromise = serverChannel.close()
-        XCTAssertFalse(closePromise.fulfilled)
+        closePromise.whenComplete {
```
weissi: why this instead of `XCTAssertFalse(closePromise.isFulfilled)`?

normanmaurer: because isFulFilled is internal ?

weissi: aaaah :)

Tests/NIOOpenSSLTests/OpenSSLIntegrationTest.swift
Lukasa: Oh yeah, good call.

Tests/NIOOpenSSLTests/OpenSSLIntegrationTest.swift
Lukasa: Nah, I don't think so. They're mostly an internal implementation detail tbh.

PR #13, opened 2018-04-29
Include md5 openssl Header

Yasumoto:
Motivation:

I'm switching [an AWS SDK](https://github.com/noppoMan/aws-sdk-swift) to use NIO under the hood, and there are some MD5 sum operations we need to complete related to requests + file operations.

Including [the `md5.h` header ](https://github.com/libressl/libressl/blob/master/src/crypto/md5/md5.h) will make it much easier to utilize and prevent requiring an additional package to get  at the functions already present here.

Modifications:

Included the `md5.h` header in `CNIOOpenSSL

Result:

This will allow md5 related functions to be used by packages including `CNIOOpenSSL`.

### DIFF
--- a/Sources/CNIOOpenSSL/include/c_nio_openssl.h
+++ b/Sources/CNIOOpenSSL/include/c_nio_openssl.h
@@ -20,6 +20,7 @@
 #include <openssl/bio.h>
 #include <openssl/ssl.h>
 #include <openssl/sha.h>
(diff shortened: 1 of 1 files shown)


Lukasa:
Sure thing, this seems totally reasonable to me. I'll tag a new release shortly, thanks!

Yasumoto:
Rad, thank you @Lukasa !

PR #20, opened 2018-06-29
Support OpenSSL 1.1 and later.

Lukasa:
Motivation:

Currently there are no Swift.org distributions of Swift for Linux that link against an OpenSSL 1.1 release, but with the recent release of Ubuntu 18.04, the next LTS release of Ubuntu, it seems reasonable that at some point such a release will exist.

Given that OpenSSL 1.1 is substantially better than OpenSSL 1.0, we should endeavour to allow our users to take advantage of this new support and compile against this newer OpenSSL.

*Unfortunately*, OpenSSL's specific source change is not well-tolerated by Swift code. There are a number of reasons for this, mostly explained in https://forums.swift.org/t/opaque-pointers-in-swift/6875. The end result of this is that we require substantial, wide-ranging changes to the implementation in order to successfully compile against both OpenSSL versions 1.0.X and 1.1.X.

This patch contains those wide ranging modifications. The most profound of them is to use the compiler to essentially convert between arbitrary pointer types without writing the names of those types down. Unfortunately this totally escapes the Swift type system, which is substantially less than ideal. However, in the current mode where opaque C structures are unityped in Swift it is really the best we can do.

Modifications:

- Supported converting all pointer types to OpaquePointer.
- Added conversions absolutely everywhere.
- Fixed a few problems with callbacks.
- Moved OpenSSL initialization to C to support both macros and
  functions.
- Made locking callbacks optional.
- Fixed tests that relied on OpenSSL 1.0.2 behaviour of ALPN failure
  to now check OpenSSL version before they assume an outcome.

Result:

Compile against OpenSSL 1.1.0 at the very least.

### DIFF
--- a/Tests/NIOOpenSSLTests/OpenSSLALPNTest.swift
+++ b/Tests/NIOOpenSSLTests/OpenSSLALPNTest.swift
@@ -18,6 +18,7 @@ import NIO
 import NIOTLS
 import NIOOpenSSL
 
+
 class OpenSSLALPNTest: XCTestCase {
     static var cert: OpenSSLCertificate!
     static var key: OpenSSLPrivateKey!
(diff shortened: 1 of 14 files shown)


tomerd:

Lukasa:
Ok, I think this is now ready for review.

Tests/NIOOpenSSLTests/OpenSSLALPNTest.swift
```diff
@@ -112,11 +154,17 @@ class OpenSSLALPNTest: XCTestCase {
...
         } catch OpenSSLError.failedToSetALPN {
-            XCTAssertTrue(SSLeay() < 0x010002000)
+            XCTAssertTrue(CNIOOpenSSL_OpenSSL_version_num() < 0x010002000)
             return
         }
 
-        try assertNegotiatedProtocol(protocol: nil, serverContext: serverCtx, clientContext: clientCtx)
+        print(CNIOOpenSSL_OpenSSL_version_num())
+        if (CNIOOpenSSL_OpenSSL_version_num() < 0x010100000) || (CNIOOpenSSL_is_libressl() == 1) {
+            print("less")
```
weissi: pinrt

Tests/NIOOpenSSLTests/OpenSSLTestHelpers.swift
```diff
@@ -243,40 +243,39 @@ func randomSerialNumber() -> ASN1_INTEGER {
 
     // Our 20-byte number needs to be converted into an integer. This is
     // too big for Swift's numbers, but OpenSSL can handle it fine.
-    var bn = BIGNUM()
+    let bn = BN_new()
```
weissi: needs free

Lukasa:

PR #21, opened 2018-07-06
Support private keys with passphrases.

Lukasa:
Motivation:

Many users want to use passphrases to secure their private keys. We have not
previously supported any mechanism to provide that passphrase other than
the default OpenSSL passphrase callback. This is not useful in all
circumstances, and is particularly troubling in the case of automated
systems. We should add better support.

Modifications:

Enhance OpenSSLPrivateKey to allow setting a passphrase callback.
Enhance SSLContext to allow setting a passphrase callback.

Result:

Users can set passphrase callbacks.


Sources/NIOOpenSSL/SSLContext.swift
```diff
@@ -112,8 +112,8 @@ public final class SSLContext {
 
     /// Initialize a context that will create multiple connections, all with the same
     /// configuration.
-    public init(configuration: TLSConfiguration) throws {
-        precondition(initialized)
+    internal init(configuration: TLSConfiguration, passphraseUserData: UnsafeMutableRawPointer?) throws {
+        guard openSSLIsInitialized else { fatalError() }
```
weissi: could we have a message here, just in case :)

Sources/NIOOpenSSL/SSLContext.swift
```diff
@@ -185,13 +185,19 @@ public final class SSLContext {
             case .some(.file(let f)):
                 try SSLContext.loadVerifyLocations(f, context: ctx)
             case .some(.certificates(let certs)):
-                    try certs.forEach { try SSLContext.addRootCertificate($0, context: ctx) }
+                try certs.forEach { try SSLContext.addRootCertificate($0, context: ctx) }
```
weissi: I'm assuming this is thread-safe?

Sources/NIOOpenSSL/SSLPrivateKey.swift
```diff
@@ -14,6 +14,95 @@
...
+
+
+/// Our global static OpenSSL passphrase callback. This is used as a thunk to dispatch out to
+/// the user-provided callback.
+func globalOpenSSLPassphraseCallback(buf: UnsafeMutablePointer<CChar>?,
+                                     size: CInt,
+                                     rwflag: CInt,
+                                     u: UnsafeMutableRawPointer?) -> CInt {
+    guard let buf = buf, let u = u else {
+        preconditionFailure("Invalid pointers passed to passphrase callback")
```
weissi: should we add the values to the string just so it's easier to debug just in case?

Sources/NIOOpenSSL/SSLPrivateKey.swift
```diff
@@ -68,7 +161,13 @@ public class OpenSSLPrivateKey {
...
-            key = PEM_read_bio_PrivateKey(bio, nil, nil, nil)
+            if let userData = pemUserData {
+                // This annoying conditional binding is used to work around the fact that I cannot pass
+                // a variable to a function pointer argument.
+                key = PEM_read_bio_PrivateKey(bio, nil, globalOpenSSLPassphraseCallback(buf:size:rwflag:u:), userData)
+            } else {
+                key = PEM_read_bio_PrivateKey(bio, nil, nil, nil)
+            }
         case .der:
             key = d2i_PrivateKey_bio(bio, nil)
```
weissi: should we assert this isn't happening on an event loop as this stuff will block?

Sources/NIOOpenSSL/SSLPrivateKey.swift
```diff
@@ -14,6 +14,95 @@
...
+/// the user-provided callback.
+func globalOpenSSLPassphraseCallback(buf: UnsafeMutablePointer<CChar>?,
+                                     size: CInt,
+                                     rwflag: CInt,
+                                     u: UnsafeMutableRawPointer?) -> CInt {
+    guard let buf = buf, let u = u else {
+        preconditionFailure("Invalid pointers passed to passphrase callback")
+    }
+    let bufferPointer = UnsafeMutableBufferPointer(start: buf, count: Int(size))
+    guard let cbManager = Unmanaged<AnyObject>.fromOpaque(u).takeRetainedValue() as? CallbackManagerProtocol else {
```
weissi: I don't think we should release this thing in here. I think we should have a clearly matched retain & release in code. Just like before ARC, a function doesn't consume references as it doesn't own the pointers that get passed into it. If it wants to own them, it needs to `retain` them.

Sources/NIOOpenSSL/SSLPrivateKey.swift
```diff
@@ -80,6 +179,56 @@ public class OpenSSLPrivateKey {
...
+    /// DER format, providing a passphrase callback.
+    ///
+    /// - parameters:
+    ///     - file: The path to the file to load.
+    ///     - format: The format of the key to load, either DER or PEM.
+    ///     - passphraseCallback: A callback to invoke to obtain the passphrase for
+    ///         encrypted keys.
+    public convenience init<T: Collection>(file: String, format: OpenSSLSerializationFormats, passphraseCallback: @escaping OpenSSLPassphraseCallback<T>) throws where T.Element == UInt8 {
+        let manager = OpenSSLPassphraseCallbackManager(userCallback: passphraseCallback)
+        let userData = Unmanaged.passRetained(manager).toOpaque()
```
weissi: this should be matched with a `release` _in this function_ just like in the days before ARC. You don't pass a pointer that you own to someone else who needs to consume it as that's to error prone. The rules are that passing a pointer isn't passing ownership. If the consumer needs to own it for longer than its stack frame it needs to retain it.

weissi: I'd actually probably do this:

```swift
withExtendedLifetime(manager) { manager in
    let managerPtr = Unmanaged.passUnretained(manager).toOpaque()
    ....(managerPtr)
}
```

Sources/NIOOpenSSL/SSLPrivateKey.swift
```diff
@@ -80,6 +179,56 @@ public class OpenSSLPrivateKey {
...
+    /// - parameters:
+    ///     - buffer: The key bytes.
+    ///     - format: The format of the key to load, either DER or PEM.
+    ///     - passphraseCallback: Optionally a callback to invoke to obtain the passphrase for
+    ///         encrypted keys. If not provided, or set to `nil`, the default OpenSSL
+    ///         behaviour will be used, which prints a prompt and requests the passphrase from
+    ///         stdin.
+    public convenience init<T: Collection>(buffer: [Int8], format: OpenSSLSerializationFormats, passphraseCallback: @escaping OpenSSLPassphraseCallback<T>) throws where T.Element == UInt8  {
+        let manager = OpenSSLPassphraseCallbackManager(userCallback: passphraseCallback)
+        let userData = Unmanaged.passRetained(manager).toOpaque()
```
weissi: same here, this should match the pre-ARC memory ownership rules.

Tests/NIOOpenSSLTests/OpenSSLIntegrationTest.swift
```diff
@@ -288,11 +295,35 @@ class OpenSSLIntegrationTest: XCTestCase {
         return try assertNoThrowWithValue(SSLContext(configuration: config), file: file, line: line)
     }
 
+    private func configuredSSLContext<T: Collection>(passphraseCallback: @escaping OpenSSLPassphraseCallback<T>,
+                                                     file: StaticString = #file, line: UInt = #line) throws -> NIOOpenSSL.SSLContext
+                                                     where T.Element == UInt8 {
+        let config = TLSConfiguration.forServer(certificateChain: [.certificate(OpenSSLIntegrationTest.cert)],
+                                                privateKey: .file(OpenSSLIntegrationTest.encryptedKeyPath),
+                                                trustRoots: .certificates([OpenSSLIntegrationTest.cert]))
+        return try assertNoThrowWithValue(SSLContext(configuration: config, passphraseCallback: passphraseCallback), file: file, line: line)
```
weissi: why doesn't the `assertNoThrowWithValue` not do `file: StaticString = #file, line: Int = #line` to have file and line taken automatically using default args?

Tests/NIOOpenSSLTests/SSLCertificateTest.swift
```diff
@@ -103,28 +103,28 @@ i5PCcPYi39q101UIxV/WokS0mqHx/XuTYTwhWYd/C49OnM8MLZOUJd8w0VvS0ItY
 -----END CERTIFICATE-----
 """
 
-func makeTemporaryFile() -> String {
-    let template = "/tmp/niotestXXXXXXX"
+func makeTemporaryFile(fileExtension: String = "") -> String {
+    let template = "/tmp/niotestXXXXXXX\(fileExtension)"
     var templateBytes = template.utf8 + [0]
     let templateBytesCount = templateBytes.count
```
weissi: shouldn't that be `ptr.count` in the block?

Sources/NIOOpenSSL/SSLContext.swift
```diff
@@ -18,7 +18,7 @@ import CNIOOpenSSL
 // This is a neat trick. Swift lazily initializes module-globals based on when they're first
 // used. This lets us defer OpenSSL intialization as late as possible and only do it if people
 // actually create any object that uses OpenSSL.
```
weissi: just in case, have you run with

- ASan in _release_ mode
- TSan

?

Sources/NIOOpenSSL/SSLContext.swift
Lukasa: Define "this"?

Tests/NIOOpenSSLTests/OpenSSLIntegrationTest.swift
Lukasa: It does, but here we are propagating the file and line from higher up so they have to be passed explicitly.

Sources/NIOOpenSSL/SSLPrivateKey.swift
Lukasa: I don't think so. This blocking is potential, and not necessarily wildly egregious, so I'm inclined to leave it in the user's hands. Regardless we certainly can't do it in this patch, as the assertion would be semver major.

Sources/NIOOpenSSL/SSLPrivateKey.swift
Lukasa: We can't actually do it that way, because the thing we have to put inside `withExtendedLifetime` is `self.init`, and if you try you get Swift complaining: "Initializer delegation ('self.init') cannot be nested in another expression"

Lukasa: So I will have to manually manage the retains/releases. Not the end of the world, just a thing.

Sources/NIOOpenSSL/SSLContext.swift
Lukasa: Yup, both. All good.

Lukasa:
Updated.

Sources/NIOOpenSSL/SSLContext.swift
weissi: `SSLContext.addRootCertificate`

Sources/NIOOpenSSL/SSLContext.swift
```diff
@@ -108,12 +108,13 @@ private func alpnCallback(ssl: UnsafeMutablePointer<SSL>?,
...
 public final class SSLContext {
     private let sslContext: UnsafeMutablePointer<SSL_CTX>
+    private let passphraseUserData: UnsafeMutableRawPointer?
     internal let configuration: TLSConfiguration
 
     /// Initialize a context that will create multiple connections, all with the same
     /// configuration.
-    public init(configuration: TLSConfiguration) throws {
-        precondition(initialized)
+    internal init(configuration: TLSConfiguration, passphraseUserData: UnsafeMutableRawPointer?) throws {
```
weissi: - let's take and store this `passphraseUserData` as `Unmanaged<thing>` and only make it an `UnsafeMutableRawPointer` when we actually need it
- also: the caller should pass us that thing however it likes, we should then `retain` it in `init` and `release` in `deinit`. It's non-standard to consume a reference (and even more odd to consume a reference that we obtain as `UnsafeMutableRawPointer`.

Sources/NIOOpenSSL/SSLPrivateKey.swift
```diff
@@ -43,7 +131,13 @@ public class OpenSSLPrivateKey {
         let key: UnsafeMutablePointer<EVP_PKEY>?
         switch format {
         case .pem:
-            key = PEM_read_PrivateKey(fileObject, nil, nil, nil)
+            // This annoying conditional binding is used to work around the fact that I cannot pass
+            // a variable to a function pointer argument.
+            if let userData = pemUserData {
+                key = PEM_read_PrivateKey(fileObject, nil, globalOpenSSLPassphraseCallback(buf:size:rwflag:u:), userData)
```
weissi: this is the point where we should go `Unmanaged<...>` to `UnsafeMutableRawPointer`

Sources/NIOOpenSSL/SSLContext.swift
```diff
@@ -224,12 +231,36 @@ public final class SSLContext {
...
+    /// - parameters:
+    ///     - configuration: The `TLSConfiguration` to use for all the connections with this
+    ///         `SSLContext`.
+    ///     - passphraseCallback: The callback to use to decrypt any private keys used by this
+    ///         `SSLContext`. For more details on this parameter see the documentation for
+    ///         `OpenSSLPassphraseCallback`.
+    public convenience init<T: Collection>(configuration: TLSConfiguration,
+                                           passphraseCallback: @escaping OpenSSLPassphraseCallback<T>) throws where T.Element == UInt8 {
+        let manager = OpenSSLPassphraseCallbackManager(userCallback: passphraseCallback)
+        let userData = Unmanaged.passRetained(manager).toOpaque()
```
weissi: this should release the pointer here, the callee shouldn't consume

Sources/NIOOpenSSL/SSLContext.swift
Lukasa: In this usage, sure is.

Sources/NIOOpenSSL/SSLContext.swift
Lukasa: I cannot take this as an Unmanaged without erasing its type by storing it in a protocol existential. If we're doing that, we may as well just take the protocol existential directly and create the unmanaged in the inner function.

Lukasa:
Ok, addressed this feedback. We now hold a protocol existential as much as possible, which reduces the amount of refcounting I do myself. Can only lead to the code being more correct. :smile:

Sources/NIOOpenSSL/SSLPrivateKey.swift
```diff
@@ -14,6 +14,95 @@
...
+/// That makes `OpenSSLPassphraseCallbackManager` a generic object, and now we're in *real* trouble, becuase
+/// `Unmanaged` requires us to specify the *complete* type of the object we want to unwrap. In this case, we
+/// don't know it, because it's generic!
+///
+/// Our way out is to note that while the class itself is generic, the only function we want to call in the
+/// `globalOpenSSLPassphraseCallback` is not. Thus, rather than try to hold the actual specific `OpenSSLPassphraseManager`,
+/// we can hold it inside a protocol existential instead, so long as that protocol existential gives us the correct
+/// function to call. Hence: `CallbackManagerProtocol`, a private protocol with a single conforming type.
+internal protocol CallbackManagerProtocol: class {
+    func invoke(buffer: UnsafeMutableBufferPointer<CChar>) -> CInt
```
weissi: invoke should have documentation.

weissi: nvm, it's `internal` actually

PR #22, opened 2018-07-06
Correctly keep buffers alive.

Lukasa:
Motivation:

BIO_new_mem_buf does not take a copy of its underlying memory, so it's
enormously important that the pointer we pass it remain valid for the
duration of the lifetime of the BIO. This was previously not actually
something we did. Very bad.

Modifications:

Extended the scope of the `withUnsafeBytes` call until the `BIO` created
from those bytes is supposed to go out of scope.

Result:

Better memory safety.

### DIFF
--- a/Sources/NIOOpenSSL/SSLCertificate.swift
+++ b/Sources/NIOOpenSSL/SSLCertificate.swift
@@ -71,26 +71,26 @@ public class OpenSSLCertificate {
     /// Create an OpenSSLCertificate from a buffer of bytes in either PEM or
     /// DER format.
     public convenience init (buffer: [Int8], format: OpenSSLSerializationFormats) throws  {
-        let bio = buffer.withUnsafeBytes {
-            return BIO_new_mem_buf(UnsafeMutableRawPointer(mutating: $0.baseAddress!), Int32($0.count))!
-        }
-        defer {
-            BIO_free(bio)
+        let ref = buffer.withUnsafeBytes { (ptr) -> UnsafeMutablePointer<X509>? in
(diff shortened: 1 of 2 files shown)


Sources/NIOOpenSSL/SSLCertificate.swift
```diff
@@ -71,26 +71,26 @@ public class OpenSSLCertificate {
     /// Create an OpenSSLCertificate from a buffer of bytes in either PEM or
     /// DER format.
     public convenience init (buffer: [Int8], format: OpenSSLSerializationFormats) throws  {
-        let bio = buffer.withUnsafeBytes {
-            return BIO_new_mem_buf(UnsafeMutableRawPointer(mutating: $0.baseAddress!), Int32($0.count))!
-        }
-        defer {
-            BIO_free(bio)
-        }
+        let ref = buffer.withUnsafeBytes { (ptr) -> UnsafeMutablePointer<X509>? in
```
weissi: how do we keep this `X509 *` alive?

Lukasa: The `X509 *` is passed into this initializer:

https://github.com/apple/swift-nio-ssl/blob/0adc938bc8de3d3829b842f9767d81c7480b8403/Sources/NIOOpenSSL/SSLCertificate.swift#L40-L42

It is freed in `deinit`:

https://github.com/apple/swift-nio-ssl/blob/0adc938bc8de3d3829b842f9767d81c7480b8403/Sources/NIOOpenSSL/SSLCertificate.swift#L163-L165

PR #23, opened 2018-07-10
Add support for PKCS#12 bundles.

### LINKED ISSUE #17: [enhancement]Support for PKCS#12 in TLSConfiguration
Kitura supports certificates in the PKCS#12 format on `macOS`. It'd be great to have that supported by `swift-nio-ssl`.

Lukasa:
Resolves #17.

Motivation:

As requested in #17, it's not uncommon to want to support PKCS#12 bundles.
This patch adds support for unpacking PKCS#12 bundles into useful data types
for NIOOpenSSL.

Modifications:

Added new OpenSSLPKCS12Bundle type.

Result:

PKCS#12 bundles can be used to distribute certs/keys for SwiftNIO.

/cc @pushkarnk


Sources/NIOOpenSSL/SSLPKCS12Bundle.swift
```diff
@@ -0,0 +1,218 @@
...
+///
+/// ### Working with TLSConfiguration
+///
+/// In many cases users will want to configure a `TLSConfiguration` with the data
+/// from a PKCS#12 bundle. This object assists in unpacking that bundle into its
+/// associated pieces.
+///
+/// If you have a PKCS12 bundle, you configure a `TLSConfiguration` like this:
+///
+///    let p12Bundle = OpenSSLPKCS12Bundle(file: pathToMyP12)
```
weissi: I think there's one space missing here, right? Needs 5 in total, one to separate it from `///` and then four for code indent I think

Sources/NIOOpenSSL/SSLPKCS12Bundle.swift
```diff
@@ -0,0 +1,218 @@
...
+        // and .withMemoryRebound(to:) are only added to the buffer pointers in 4.1.
+        let bufferSize = Int(self.count) + 1
+        let ptr = UnsafeMutablePointer<UInt8>.allocate(capacity: bufferSize)
+        let bufferPtr = UnsafeMutableBufferPointer(start: ptr, count: bufferSize)
+        defer {
+            ptr.deallocate()
+        }
+
+        if mlock(bufferPtr.baseAddress!, bufferPtr.count) != 0 {
+            throw IOError(errnoCode: errno, function: "mlock")
```
weissi: this isn't guaranteed to give you the right `errno`. Can we move into the `Posix` module?

Sources/NIOOpenSSL/SSLPKCS12Bundle.swift
```diff
@@ -0,0 +1,218 @@
...
+    /// memory region that is locked into memory (that is, which can never be backed by a file),
+    /// and which will be scrubbed and freed after use.
+    ///
+    /// This method should be used when it is necessary to take a secure copy of a collection of
+    /// bytes. Its implementation relies on OpenSSL directly.
+    internal func withSecureBytes<T>(_ block: (UnsafePointer<Int8>) throws -> T) throws -> T {
+        // We need to allocate some memory and prevent it being swapped to disk while we use it.
+        // For that reason we use mlock.
+        // Note that here we use UnsafePointer and UnsafeBufferPointer. Ideally we'd just use
+        // the buffer pointer, but 4.0 doesn't have the APIs we want: specifically, .allocate
```
weissi: I think we created an extension for that, no?

Sources/NIOOpenSSL/SSLPKCS12Bundle.swift
```diff
@@ -0,0 +1,218 @@
...
+            ptr.deallocate()
+        }
+
+        if mlock(bufferPtr.baseAddress!, bufferPtr.count) != 0 {
+            throw IOError(errnoCode: errno, function: "mlock")
+        }
+        defer {
+            // If munlock fails take out the process.
+            let rc = munlock(bufferPtr.baseAddress!, bufferPtr.count)
+            precondition(rc == 0, "munlock failed with \(rc), errno \(errno)")
```
weissi: also should go to `Posix`

Sources/NIOOpenSSL/SSLPKCS12Bundle.swift
```diff
@@ -0,0 +1,218 @@
...
+            X509_free(cert)
+            EVP_PKEY_free(pkey)
+            CNIOOpenSSL_sk_X509_free(caCerts)
+            throw NIOOpenSSLError.unableToAllocateOpenSSLObject
+        }
+
+        let certStackSize = caCerts.map(CNIOOpenSSL_sk_X509_num) ?? 0
+        var certs = [OpenSSLCertificate]()
+        certs.reserveCapacity(Int(certStackSize) + 1)
+        certs.append(OpenSSLCertificate.fromUnsafePointer(pointer: actualCert))
```
weissi: I'm assuming the `fromUnsafePointer(pointer: ...)` functions are consuming a reference? If yes, could we file an issue to rewrite them to something that doesn't consume the reference and copies the contents instead. If OpenSSL doesn't have copying functions for that then the label needs to become better and needs to read `fromUnsafePointer(transferringOwnership: actualCert)` etc...

weissi: and we need a comment at each invocation site that they're non-standard and just take ownership of the passed pointer.

Sources/NIOOpenSSL/SSLPKCS12Bundle.swift
```diff
@@ -0,0 +1,218 @@
...
+            defer {
+                BIO_free(bio)
+            }
+            return d2i_PKCS12_bio(bio, nil)
+        }
+        defer {
+            p12.map(PKCS12_free)
+        }
+
+        if let p12 = p12 {
```
weissi: similar issues here: the lifetime management of `p12` is unclear to me.

Sources/NIOOpenSSL/SSLPKCS12Bundle.swift
```diff
@@ -0,0 +1,218 @@
...
+    /// - parameters:
+    ///     - buffer: The bytes of the PKCS#12 bundle.
+    ///     - passphrase: The passphrase used for the bundle, as a sequence of UTF-8 bytes.
+    public init<Bytes: Collection>(buffer: [UInt8], passphrase: Bytes?) throws where Bytes.Element == UInt8 {
+        guard openSSLIsInitialized else { fatalError("Failed to initialize OpenSSL") }
+        
+        let p12 = buffer.withUnsafeBytes { pointer -> UnsafeMutablePointer<PKCS12>? in
+            let bio = BIO_new_mem_buf(UnsafeMutableRawPointer(mutating: pointer.baseAddress), CInt(pointer.count))!
+            defer {
+                BIO_free(bio)
```
weissi: so `bio` can be freed before the returned value?

Sources/NIOOpenSSL/SSLPKCS12Bundle.swift
```diff
@@ -0,0 +1,218 @@
...
+
+        let p12 = d2i_PKCS12_fp(fileObject, nil)
+        defer {
+            p12.map(PKCS12_free)
+        }
+
+        if let p12 = p12 {
+            try self.init(ref: p12, passphrase: passphrase)
+        } else {
+            throw OpenSSLError.unknownError(.init())
```
weissi: what's the `.init()` in there doing? Could we make it `ActualType()` ?

Sources/NIOOpenSSL/SSLPKCS12Bundle.swift
```diff
@@ -0,0 +1,218 @@
...
+
+    /// Create a `OpenSSLPKCS12Bundle` from the given bytes on disk,
+    /// assuming it has no passphrase.
+    ///
+    /// If the bundle does have a passphrase, call `init(file:passphrase:)` instead.
+    ///
+    /// - parameters:
+    ///     - file: The path to the PKCS#12 bundle on disk.
+    public init(file: String) throws {
+        try self.init(file: file, passphrase: nil as [UInt8]?)
```
weissi: not sure but I prefer `Optional<[UInt8]>.none` or `[UInt8]?.none` instead of `nil as [UInt8]?` but feel free to ignore

Sources/NIOOpenSSL/SSLPKCS12Bundle.swift
```diff
@@ -0,0 +1,218 @@
...
+    ///
+    /// This method should be used when it is necessary to take a secure copy of a collection of
+    /// bytes. Its implementation relies on OpenSSL directly.
+    internal func withSecureBytes<T>(_ block: (UnsafePointer<Int8>) throws -> T) throws -> T {
+        // We need to allocate some memory and prevent it being swapped to disk while we use it.
+        // For that reason we use mlock.
+        // Note that here we use UnsafePointer and UnsafeBufferPointer. Ideally we'd just use
+        // the buffer pointer, but 4.0 doesn't have the APIs we want: specifically, .allocate
+        // and .withMemoryRebound(to:) are only added to the buffer pointers in 4.1.
+        let bufferSize = Int(self.count) + 1
```
weissi: why does it allocate one larger and put a `NULL` in there? If that's necessary it should read `withSecureCString` or something maybe?

Sources/NIOOpenSSL/SSLPKCS12Bundle.swift
```diff
@@ -0,0 +1,218 @@
...
+        let ptr = UnsafeMutablePointer<UInt8>.allocate(capacity: bufferSize)
+        let bufferPtr = UnsafeMutableBufferPointer(start: ptr, count: bufferSize)
+        defer {
+            ptr.deallocate()
```
weissi: the memory needs to be `deinitialized` too before `deallocate`

Sources/NIOOpenSSL/SSLPKCS12Bundle.swift
```diff
@@ -0,0 +1,218 @@
...
+            // We use OPENSSL_cleanse here because the compiler can't optimize this away.
+            // .initialize(repeating: 0) can be, and empirically is, optimized away, bzero
+            // is deprecated, memset_s is not well supported cross-platform, and memset-to-zero
+            // is famously easily optimised away. This is our best bet.
+            OPENSSL_cleanse(bufferPtr.baseAddress!, bufferPtr.count)
+        }
+        assert(nextIndex == (bufferPtr.endIndex - 1))
+
+        // Add a null terminator.
+        bufferPtr[nextIndex] = 0
```
weissi: I think this is technically wrong as you're writing to unintialised memory here. I think it should be `bufferPtr.advanced(by: nextIndex).initialize(to: 0)`.

I realise that `UInt8` is a primitive type that doesn't have any ref counts but imagine you'd have saved a type that needs ref counts. In that case this would non-deterministically crash. Because `bufferPtr[nextIndex] = newValue` would first read out the old value, decrement the ref count, the increment the ref count of `newValue` and then write it in there. `initialize` on the other hand doesn't touch the ref count of the memory but just increases `newValue`'s ref count and then writes it into the memory (which then becomes initialised).

Yes, `UInt8` doesn't have a ref count of course but still the same rules apply. Memory needs to be 1) allocated 2) initialised 3) written as often as the user pleases 4) deinitialised 5) deallocated

You're doing (1), most of (2) (except for the last byte), (3) and (5). So parts of (1) and (4) are missing

Sources/NIOOpenSSL/SSLPKCS12Bundle.swift
Lukasa: Yes. The BIO is used to read the data in, but that's it: the PKCS12 object holds a structured version of the data.

Sources/NIOOpenSSL/SSLPKCS12Bundle.swift
Lukasa: If `p12` is `nil` then we have no lifecycle to manage: there is no object here to free, only the BIO, which was freed once we exited the block above.

Sources/NIOOpenSSL/SSLPKCS12Bundle.swift
Lukasa: It can't use `withSecureCString` because it operates on arbitrary collections, just as the passphrase code for private keys does.

Sources/NIOOpenSSL/SSLPKCS12Bundle.swift
Lukasa: No, just for `deallocate` without size, which in fact I use. ;)

Sources/NIOOpenSSL/SSLPKCS12Bundle.swift
```diff
@@ -0,0 +1,220 @@
...
+    /// Create a `OpenSSLPKCS12Bundle` from the given bytes on disk,
+    /// optionally decrypting the bundle with the given passphrase.
+    ///
+    /// - parameters:
+    ///     - file: The path to the PKCS#12 bundle on disk.
+    ///     - passphrase: The passphrase used for the bundle, as a sequence of UTF-8 bytes.
+    public init<Bytes: Collection>(file: String, passphrase: Bytes?) throws where Bytes.Element == UInt8 {
+        guard openSSLIsInitialized else { fatalError("Failed to initialize OpenSSL") }
+
+        let fileObject = fopen(file, "rb")
```
weissi: please handle `NULL` as at least `fclose` blows up on `NULL`

weissi: also: please add test case for `/dev/null/does/not/exist` or so

Sources/NIOOpenSSL/SSLPKCS12Bundle.swift
weissi: ok, as of our chat a minute ago: this should then have its own `Posix` wrapper following `NIO`'s pattern

Sources/NIOOpenSSL/SSLPKCS12Bundle.swift
weissi: here too, because as you pointed our: `EINTR` could be a real issue here ...

Sources/NIOOpenSSL/SSLPKCS12Bundle.swift
Lukasa: I agree, but are you open to me doing this as a follow-up PR? There are a few other sys calls in the package, I'd like to fix it all in one go.

weissi: @Lukasa yes!

pushkarnk:
@Lukasa Thanks a lot for your work on this! Much appreciated!

PR #24, opened 2018-07-12
Move syscalls to a port of Posix enum.

Lukasa:
Motivation:

Right now we make a bunch of naked syscalls that don't correctly
handle reporting errno or catching EINTR. We can't use the NIO Posix
enum to fix this for us because it's internal, and because it doesn't
have the function we need. Instead, we create a temporary port of those
functions.

Ideally we'd reconcile these at some time in the future.

Modifications:

Add a port of the Posix enum and its core functions.
Improve wrapErrorIsNullReturnCall to be more useful.
Adopt the posix functions.

Result:

Better safety in the face of EINTR.
Better reporting of errnos.


Tests/NIOOpenSSLTests/SSLCertificateTest.swift
```diff
@@ -251,8 +251,8 @@ class SSLCertificateTest: XCTestCase {
         do {
             _ = try OpenSSLCertificate(file: "/nonexistent/path", format: .pem)
             XCTFail("Did not throw")
-        } catch NIOOpenSSLError.noSuchFilesystemObject {
-            // ok
+        } catch let error as IOError {
+            XCTAssertEqual(error.errnoCode, ENOENT)
```
Lukasa: All of these test changes are a change in the thrown error, but they're semver patch because these functions never used to throw in this circumstance at all: they crashed. The error thrown before this patch was added in #23.

PR #27, opened 2018-07-31
Add custom ByteBufferBIO object.

Lukasa:
Motivation:

Currently we use OpenSSL with MemoryBIO objects. These objects are an
implementation of OpenSSL's pervasive BIO abstraction that wrap a simple
heap-allocated buffer.

The result of this is that we have the following data pipeline for
reads:

1. Ciphertext read from the socket into ByteBuffer
2. Ciphertext copied from ByteBuffer to memory BIO
3. Ciphertext copied from memory BIO to transient OpenSSL memory
4. Ciphertext decrypted
5. Plaintext copied from transient OpenSSL memory to ByteBuffer.

And for writes:

1. Plaintext copied from ByteBuffer to transient OpenSSL memory.
2. Plaintext encrypted.
3. Ciphertext copied from transient OpenSSL memory to memory BIO.
4. Plaintext copied from memory BIO to ByteBuffer.
5. Plaintext copied from ByteBuffer to kernel.

The thing is, we don't need all of these copies. Specifically, the
interposing of a memory BIO is totally unnnecessary. We can remove
that copy by implementing the BIO abstraction ourself on top of the
ByteBuffer objects we're using.

This patch does just that.

Modifications:

- Implement a new ByteBufferBIO object.
- Rewrite some core code to use ByteBufferBIO.
- Rewrite the read path to reduce overhead.
- Much testing.

Result:

Less memory copying, better efficiency.


ianpartridge:
Do you have any numbers for the difference this makes?

Lukasa:
This patch, combined with apple/swift-nio#539 and a minor change to the TLS echo server in this module, leads to a roughly 50% improvement in sheer TLS throughput. 

I can get you a more specific breakdown tomorrow if you’re interested.

ianpartridge:
Thanks, that's a big win!

Sources/NIOOpenSSL/ByteBufferBIO.swift
```diff
@@ -0,0 +1,289 @@
...
+//===----------------------------------------------------------------------===//
+
+import NIO
+import CNIOOpenSSL
+
+
+/// The OpenSSL entry point to write to the `ByteBufferBIO`. This thunk unwraps the user data
+/// and then passes the call on to the specific BIO reference.
+internal func openSSLBIOWriteFunc(bio: UnsafeMutablePointer<BIO>?, buf: UnsafePointer<CChar>?, len: CInt) -> CInt {
+    guard let concreteBIO = bio, let concreteBuf = buf else {
```
weissi: just a question, why not `guard let bio = bio, buf = buf else {` ? Leaves fewer useless variables in scope

Sources/NIOOpenSSL/ByteBufferBIO.swift
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    let swiftBIO: ByteBufferBIO = Unmanaged.fromOpaque(userPtr).takeUnretainedValue()
+    let bufferPtr = UnsafeBufferPointer(start: concreteBuf, count: Int(len))
+    return swiftBIO.sslWrite(buffer: bufferPtr)
+}
+
+/// The OpenSSL entry point to read from the `ByteBufferBIO`. This thunk unwraps the user data
+/// and then passes the call on to the specific BIO reference.
+internal func openSSLBIOReadFunc(bio: UnsafeMutablePointer<BIO>?, buf: UnsafeMutablePointer<CChar>?, len: CInt) -> CInt {
```
weissi: same questions as above

Sources/NIOOpenSSL/ByteBufferBIO.swift
```diff
@@ -0,0 +1,289 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+import CNIOOpenSSL
+
+
+/// The OpenSSL entry point to write to the `ByteBufferBIO`. This thunk unwraps the user data
+/// and then passes the call on to the specific BIO reference.
+internal func openSSLBIOWriteFunc(bio: UnsafeMutablePointer<BIO>?, buf: UnsafePointer<CChar>?, len: CInt) -> CInt {
```
weissi: - could you document that the types are silly (buf & len separate) because this function needs to be compatible with `@convention(c)`
- couldn't we accept an `UnsafeRawPointer` here?

Sources/NIOOpenSSL/ByteBufferBIO.swift
```diff
@@ -0,0 +1,289 @@
...
+    ///
+    /// Generally speaking OpenSSL wants to own the object initialization logic for a BIO.
+    /// This doesn't work for us, because we'd like to ensure that the `ByteBufferBIO` is
+    /// correctly initialized with all the state it needs. One of those bits of state is
+    /// a `ByteBuffer`, which OpenSSL cannot give us, so we need to build our `ByteBufferBIO`
+    /// *first* and then use that to drive `BIO` initialization.
+    ///
+    /// Because of this split initialization dance, we elect to initialize this data structure,
+    /// and have it own building an OpenSSL `BIO` structure.
+    private var bioPtr: UnsafeMutablePointer<BIO>
```
weissi: is the only reason we have this as a `var` that we can `0xdeadbeef` it?

Sources/NIOOpenSSL/ByteBufferBIO.swift
```diff
@@ -0,0 +1,289 @@
...
+        CNIOOpenSSL_BIO_set_init(self.bioPtr, 1)
+        CNIOOpenSSL_BIO_set_shutdown(self.bioPtr, 1)
+    }
+
+    deinit {
+        // On deinit we need to drop our reference to the BIO, and also ensure that it doesn't hold any
+        // pointers to this object anymore. We also want to dirty our own ref to the BIO.
+        CNIOOpenSSL_BIO_set_data(self.bioPtr, nil)
+        BIO_free(self.bioPtr)
+        self.bioPtr = UnsafeMutablePointer(bitPattern: 0xdeadbeef)!
```
weissi: consider using `0xdedbeef` or something for 32-bit systems

Sources/NIOOpenSSL/ByteBufferBIO.swift
```diff
@@ -0,0 +1,289 @@
...
+    /// Owned pointer to the backing OpenSSL BIO object.
+    ///
+    /// This pointer is safe to use elsewhere, as it has increased the reference to the backing
+    /// `BIO`. This makes it safe to use with OpenSSL functions that require an owned reference
+    /// (that is, that consume a reference count).
+    ///
+    /// Note that the BIO may not remain useful for long periods of time: if the `ByteBufferBIO`
+    /// object that owns the BIO goes out of scope, the BIO will have its pointers invalidated
+    /// and will no longer be able to send/receive data.
+    internal var bio: UnsafeMutablePointer<BIO> {
```
weissi: can we please make this a function called `ownedBIO` or `createBIO` or something? Not happy with either name suggestion but I'd like it to be a function so it's more clear it may have side-effects and also to stick more info there (ie. returns a +1 reference)

Sources/NIOOpenSSL/ByteBufferBIO.swift
```diff
@@ -0,0 +1,289 @@
...
+    /// - returns: The number of bytes we have copied.
+    fileprivate func sslRead(buffer: UnsafeMutableBufferPointer<CChar>) -> CInt {
+        guard var inboundBuffer = self.inboundBuffer else {
+            // We have no bytes to read. Mark this as "needs read retry".
+            CNIOOpenSSL_BIO_set_retry_read(self.bioPtr)
+            return -1
+        }
+
+        let bytesToCopy = min(buffer.count, inboundBuffer.readableBytes)
+        _ = inboundBuffer.readWithUnsafeReadableBytes { bytePointer in
```
weissi: does this function not allow to discard the return value? It should...

Sources/NIOOpenSSL/ByteBufferBIO.swift
```diff
@@ -0,0 +1,289 @@
...
+            assert(bytePointer.count >= bytesToCopy)
+            memcpy(buffer.baseAddress!, bytePointer.baseAddress!, bytesToCopy)
```
weissi: there's `buffer.baseAddress!.copyMemory(...)`

Sources/NIOOpenSSL/ByteBufferBIO.swift
```diff
@@ -0,0 +1,289 @@
...
+
+        return CInt(bytesToCopy)
+    }
+
+    /// OpenSSL has requested to write ciphertext bytes from the network.
+    ///
+    /// - parameters:
+    ///     - buffer: The buffer for NIO to copy bytes from.
+    /// - returns: The number of bytes we have copied.
+    fileprivate func sslWrite(buffer: UnsafeBufferPointer<CChar>) -> CInt {
```
weissi: why not URBP?

Sources/NIOOpenSSL/ByteBufferBIO.swift
```diff
@@ -0,0 +1,289 @@
...
+        if self.mustClearOutboundBuffer {
+            // We just flushed, and this is a new write. Let's clear the buffer now.
+            self.outboundBuffer.clear()
+            assert(!self.mustClearOutboundBuffer)
+        }
+
+        let byteCount = buffer.count
+        let writtenBytes = buffer.baseAddress!.withMemoryRebound(to: UInt8.self, capacity: byteCount) { (newPointer) -> Int in
```
weissi: I don't think `withMemoryRebound(to:)` is correct here because you don't know the original type

weissi: (in the new model)

Sources/NIOOpenSSL/OpenSSLHandler.swift
```diff
@@ -248,27 +264,46 @@ public class OpenSSLHandler : ChannelInboundHandler, ChannelOutboundHandler {
...
-                // TODO(cory): Should we coalesce these instead of dispatching multiple times? I think so!
-                // It'll also let us avoid this weird boolean flag, which is always good.
-                didDeliverData = true
-                ctx.fireChannelRead(self.wrapInboundOut(buf))
+            case .complete:
+                // Good read. We need to check how many bytes are still writable and, if it's
+                // too few, move them along. As a general heuristic we want room to write 1kB
+                // of data, though this number is utterly arbitrary.
+                if receiveBuffer.writableBytes < 1024 {
+                    receiveBuffer.changeCapacity(to: receiveBuffer.capacity * 2)
```
weissi: - this doesn't guarantee it's `>= 1024` still, right? Maybe `changeCapacity(to: 1025)`?
- also shouldn't we use `reserveCapacity` and just require NIO >= 1.9.0? People still have the old versions available if they want to stay on the old stuff

Sources/NIOOpenSSL/SSLConnection.swift
```diff
@@ -49,22 +48,10 @@ internal final class SSLConnection {
         return false
     }
     
-    init? (_ ssl: UnsafeMutablePointer<SSL>, parentContext: SSLContext) {
+    init (_ ssl: UnsafeMutablePointer<SSL>, parentContext: SSLContext) {
```
weissi: nit: extra space

Sources/NIOOpenSSL/SSLConnection.swift
```diff
@@ -49,22 +48,10 @@ internal final class SSLConnection {
...
         self.ssl = ssl
```
weissi: don't you need to openssl_up_ref that?

Sources/NIOOpenSSL/SSLConnection.swift
```diff
@@ -49,22 +48,10 @@ internal final class SSLConnection {
...
-        if fromNetwork == nil || toNetwork == nil {
-            // TODO(cory): I don't love having this cleanup here, it means I have to remember it.
-            SSL_free(ssl)
-            return nil
-        }
-        
-        self.fromNetwork = fromNetwork!
-        self.toNetwork = toNetwork!
-        SSL_set_bio(self.ssl, self.fromNetwork, self.toNetwork)
+
```
weissi: nit: extra line

Sources/NIOOpenSSL/SSLConnection.swift
```diff
@@ -81,6 +68,16 @@ internal final class SSLConnection {
...
     }
 
+    func setAllocator(_ allocator: ByteBufferAllocator) {
+        self.bio = ByteBufferBIO(allocator: allocator)
+
+        // This weird dance where we pass the *exact same* pointer in to both objects is because, weirdly,
+        // the OpenSSL docs claim that only one reference count will be consumed here. We therefore need to
+        // avoid calling BIO_up_ref too many times.
+        let bioPtr = self.bio!.bio
+        SSL_set_bio(self.ssl, bioPtr, bioPtr)
```
weissi: and the previous `bioPtr` will be `down_ref`'d automatically? what happens if old `bioPtr` and new `bioPtr` are equal, does openssl handle that too?

Tests/NIOOpenSSLTests/ByteBufferBIOTest.swift
```diff
@@ -0,0 +1,328 @@
...
+
+
+final class ByteBufferBIOTest: XCTestCase {
+    override func setUp() {
+        guard openSSLIsInitialized else {
+            fatalError("Cannot run tests without OpenSSL")
+        }
+    }
+
+    private func bioWithDroppedRef() -> UnsafeMutablePointer<BIO> {
```
weissi: this seems very unsafe, nothing is keeping `swiftBIO` alive

Sources/NIOOpenSSL/ByteBufferBIO.swift
Lukasa: It does not.

Sources/NIOOpenSSL/ByteBufferBIO.swift
Lukasa: Yes I do, it's `CChar`. This pointer came in from OpenSSL as a `const char *`.

Sources/NIOOpenSSL/OpenSSLHandler.swift
Lukasa: I think it does. The buffer is originally allocated at 16kB, so doubling the capacity will always yield a buffer at least twice as large.

Lukasa: I can assert to that effect if you want.

Sources/NIOOpenSSL/SSLConnection.swift
Lukasa: No, this function takes an owned ref. As it's purely internal, we can guarantee safe usage.

Sources/NIOOpenSSL/SSLConnection.swift
Lukasa: This function is some bullshit. Even the manual acknowledges it. From the man page:

> `SSL_set_bio()` does a similar job as `SSL_set0_rbio()` and `SSL_set0_wbio()` except that it connects both the `rbio` and the `wbio` at the same time. This function transfers the ownership of `rbio` and `wbio` to `ssl` except that the rules for this are much more complex. For this reason this function is considered a legacy function and `SSL_set0_rbio()` and `SSL_set0_wbio()` should be used in preference. The ownership rules are as follows:
>
> - If neither the `rbio` or `wbio` have changed from their previous values then nothing is done.
> - If the `rbio` and `wbio` parameters are different and both are different to their previously set values then one reference is consumed for the `rbio` and one reference is consumed for the `wbio`.
> - If the `rbio` and `wbio` parameters are the same and the `rbio` is not the same as the previously set value then one reference is consumed.
> - If the `rbio` and `wbio` parameters are the same and the `rbio` is the same as the previously set value, then no additional references are consumed.
> - If the `rbio` and `wbio` parameters are different and the `rbio` is the same as the previously set value then one reference is consumed for the `wbio` and no references are consumed for the `rbio`.
> - If the `rbio` and `wbio` parameters are different and the `wbio` is the same as the previously set value and the old `rbio` and `wbio` values were the same as each other then one reference is consumed for the `rbio` and no references are consumed for the `wbio`.
> - If the `rbio` and `wbio` parameters are different and the `wbio` is the same as the previously set value and the old `rbio` and `wbio` values were different to each other then one reference is consumed for the `rbio` and one reference is consumed for the `wbio`.

*Unfortunately*, the replacement functions are OpenSSL 1.1 only, and it is deeply non-trivial to try to use `SSL_set_bio` to re-implement the two replacement functions. In this case, we just rely on the documented OpenSSL behaviour and the fact that all of this code is internal.

Tests/NIOOpenSSLTests/ByteBufferBIOTest.swift
Lukasa: That's the purpose of the function: to demonstrate that this works ok.

Sources/NIOOpenSSL/ByteBufferBIO.swift
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    /// OpenSSL has requested to read ciphertext bytes from the network.
+    ///
+    /// This function is invoked whenever OpenSSL is looking to read data.
+    ///
+    /// - parameters:
+    ///     - buffer: The buffer for NIO to copy bytes into.
+    /// - returns: The number of bytes we have copied.
+    fileprivate func sslRead(buffer: UnsafeMutableBufferPointer<CChar>) -> CInt {
```
weissi: can we make this UMRBP

Sources/NIOOpenSSL/OpenSSLHandler.swift
weissi: after chatting to you I suggest `receiveBuffer.capacity + 1024`

Sources/NIOOpenSSL/SSLConnection.swift
weissi: new suggestions: 
- either take ownership by upping or
- name `owned` like in the older PR

Sources/NIOOpenSSL/SSLConnection.swift
weissi: ok cool, then we need an `assert(bioPtr != getOldBIOPtr())`

Sources/NIOOpenSSL/ByteBufferBIO.swift
Lukasa: As discussed in person, we want to use the old values in the body of the `else` block, which requires we bind to new names.

Sources/NIOOpenSSL/ByteBufferBIO.swift
Lukasa: As discussed in person, we want to use the old values in the body of the `else` block, which requires we bind to new names.

Lukasa:
Ok, feedback addressed.

Lukasa:
Ok, feedback addressed.

Sources/CNIOOpenSSL/helpers.c
```diff
@@ -0,0 +1,82 @@
...
+BIO_METHOD *CNIOOpenSSL_ByteBufferBIOMethod = NULL;
+
+/// This is the type of the ByteBufferBIO created by Swift.
+///
+/// This type is used to create the BIO in Swift code. It can also be used to gate
+/// initialization: if this is non-zero, we have already initialized ByteBufferBIOMethod
+/// and can safely use it.
+int CNIOOpenSSL_ByteBufferBIOType = 0;
+
+static const char *bio_name = "ByteBuffer BIO";
```
weissi: nit, you're missing one `const` here to make it a true constant: `static const char * const bio_name = "ByteBuffer BIO";`

Sources/CNIOOpenSSL/helpers.c
```diff
@@ -0,0 +1,82 @@
...
+#include <assert.h>
+#include <openssl/bio.h>
+
+/// This is a pointer to the BIO_METHOD structure for NIO's ByteBufferBIO.
+///
+/// This structure is always initialized at startup, and must be initialized in a
+/// thread-safe manner. That means it should be guarded by some kind of pthread_once
+/// setup behaviour, or a lock. For NIO, we use the initializeOpenSSL dance to do
+/// this construction.
+BIO_METHOD *CNIOOpenSSL_ByteBufferBIOMethod = NULL;
```
weissi: can we make this `const BIO_METHOD *` ? We need to reassign it, yes but we don't need to write the pointee, do we?

Sources/NIOOpenSSL/ByteBufferBIO.swift
```diff
@@ -0,0 +1,302 @@
...
+    fileprivate func sslRead(buffer: UnsafeMutableRawBufferPointer) -> CInt {
+        guard var inboundBuffer = self.inboundBuffer else {
+            // We have no bytes to read. Mark this as "needs read retry".
+            CNIOOpenSSL_BIO_set_retry_read(self.bioPtr)
+            return -1
+        }
+
+        let bytesToCopy = min(buffer.count, inboundBuffer.readableBytes)
+        _ = inboundBuffer.readWithUnsafeReadableBytes { bytePointer in
+            assert(bytePointer.count >= bytesToCopy)
```
weissi: feel free to ignore, in these conditions it's usually good to see the values should this ever actually trigger so maybe a string which shows the values?

Sources/NIOOpenSSL/SSLConnection.swift
```diff
@@ -49,22 +48,9 @@ internal final class SSLConnection {
         return false
     }
     
-    init? (_ ssl: UnsafeMutablePointer<SSL>, parentContext: SSLContext) {
-        self.ssl = ssl
+    init (_ ownedSSL: UnsafeMutablePointer<SSL>, parentContext: SSLContext) {
```
weissi: we should make the `ownedSSL` the name that is visible from the outside, otherwise it doesn't carry its meaning as well

weissi: also: extra space between `init` and `(`

Lukasa:
Ack, feedback addressed.

Issue #12, opened 2018-04-03
HTTPS example


weissi:
we should have an HTTPS example in here just so that people know how to get that going.

weissi:
this is now done, thanks @mike-neck

PR #32, opened 2018-09-01
Provide an http client example.

mike-neck:
Provide an HTTPS example. (#12)

Motivation:

To provide an example for people to use swift-nio-ssl at their projects with ease.

Modifications:

- Add NIOHTTP1Client module.

Result:

A good example for getting started with swift-nio-ssl.


Sources/NIOHTTP1Client/main.swift
```diff
@@ -0,0 +1,89 @@
...
+}
+
+let tlsConfiguration = TLSConfiguration.forClient()
+let sslContext = try! SSLContext(configuration: tlsConfiguration)
+let openSslHandler = try! OpenSSLClientHandler(context: sslContext, serverHostname: "httpbin.org")
+
+let bootstrap = ClientBootstrap(group: eventLoopGroup)
+        .channelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR), value: 1)
+        .channelInitializer { channel in
+            _ = channel.pipeline.add(handler: openSslHandler)
```
weissi: 1. the `openSslHandler` should be constructed inside of the `channelInitializer` closure as it might be run more than once (see the [warning in here](https://apple.github.io/swift-nio/docs/current/NIO/Classes/ClientBootstrap.html#/s:3NIO15ClientBootstrapC18channelInitializerACXDAA15EventLoopFutureCyytGAA7Channel_pcF) )
2. the futures should be properly chained:

```swift
[...]
.channelInitializer { channel in
    return channel.pipeline.add(handler: OpenSSLClientHandler(...)).then {
        channel.pipeline.addHTTPClientHandlers()
    }.then {
        channel.pipeline.add(handler: HTTPResponseHandler(promise))
    }
}
```

Sources/NIOHTTP1Client/main.swift
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+func sendRequest(_ channel: Channel) {
+    var request = HTTPRequestHead(version: HTTPVersion(major: 1, minor: 1), method: HTTPMethod.GET, uri: "https://httpbin.org/get?query=param")
+    request.headers = HTTPHeaders([
+        ("Host", "httpbin.org"),
+        ("User-Agent", "swift-nio"),
+        ("Accept", "application/json")
+    ])
+    _ = channel.write(HTTPClientRequestPart.head(request))
```
weissi: 1. please use

`channel.write(..., promise:nil)`

if you don't care about the result
2. the `sendRequest` function should probably return an `EventLoopFuture<Void>` which is the result of the `writeAndFlush`

Sources/NIOHTTP1Client/main.swift
```diff
@@ -0,0 +1,89 @@
...
+    _ = channel.writeAndFlush(HTTPClientRequestPart.end(nil))
+}
+
+bootstrap.connect(host: "httpbin.org", port: 443)
+        .whenSuccess { sendRequest($0) }
```
weissi: after changing `sendRequest` to return an `EventLoopFuture` you should use 

```swift
bootstrap.connect(...)
    .then {
        sendRequest($0)
    }
    .cascadeFailure(promise: promise)
```

the reason being that your program will hang until `promise` is fulfilled. The problem is if we can never write the request in `sendRequest` you'll also never get a response so `promise` won't ever be fulfilled, ie. we'll hang forever. Therefore in case of an error in `sendRequest` we want to cascade that failure into `promise` so that the program terminates.

Sources/NIOHTTP1Client/main.swift
```diff
@@ -0,0 +1,89 @@
...
+    request.headers = HTTPHeaders([
+        ("Host", "httpbin.org"),
+        ("User-Agent", "swift-nio"),
+        ("Accept", "application/json")
+    ])
+    _ = channel.write(HTTPClientRequestPart.head(request))
+    _ = channel.writeAndFlush(HTTPClientRequestPart.end(nil))
+}
+
+bootstrap.connect(host: "httpbin.org", port: 443)
```
weissi: we should probably accept hostname, port and target (possibly as a URL) from the command line.

weissi:
@mike-neck sorry for the massive delay, I totally missed your PR here.

Sources/NIOHTTP1Client/main.swift
mike-neck: Thanks! I fixed them. Please see 6e21fc7c597f7dd6aabf1669f28b07ab579d2121 .

Sources/NIOHTTP1Client/main.swift
mike-neck: I made `sendRequest` to return `EventLoopFuture<Void>` and cascade `promise`(via your advice on https://github.com/apple/swift-nio-ssl/pull/32#discussion_r217534442). Please see 09809a479c1c2fb76a11136abdd5ea4d73df1695.

Sources/NIOHTTP1Client/main.swift
mike-neck: I fixed it in the commit 09809a479c1c2fb76a11136abdd5ea4d73df1695.

Sources/NIOHTTP1Client/main.swift
```diff
@@ -0,0 +1,100 @@
...
+            print("\(httpResponseHeader.version) \(httpResponseHeader.status.code) \(httpResponseHeader.status.reasonPhrase)")
+            for (name, value) in httpResponseHeader.headers {
+                print("\(name): \(value)")
+            }
+        case .body(var byteBuffer):
+            if let responseBody = byteBuffer.readString(length: byteBuffer.readableBytes) {
+                print(responseBody)
+            }
+        case .end(_):
+            _ = ctx.channel.close()
```
weissi: `ctx.channel.close(promise: nil)` expresses what you want a little bit better and doesn't allocate

Sources/NIOHTTP1Client/main.swift
```diff
@@ -0,0 +1,100 @@
...
+            promise.succeed(result: ())
+        }
+    }
+
+    func errorCaught(ctx: ChannelHandlerContext, error: Error) {
+        print("Error: ", error)
+        _ = ctx.channel.close()
```
weissi: same here, `ctx.channel.close(promise: nil)`

Sources/NIOHTTP1Client/main.swift
```diff
@@ -0,0 +1,100 @@
...
+                channel.pipeline.add(handler: HTTPResponseHandler(promise))
+            }
+        }
+
+func sendRequest(_ channel: Channel) -> EventLoopFuture<Void> {
+    var request = HTTPRequestHead(version: HTTPVersion(major: 1, minor: 1), method: HTTPMethod.GET, uri: url.absoluteString)
+    request.headers = HTTPHeaders([
+        ("Host", url.host!),
+        ("User-Agent", "swift-nio"),
+        ("Accept", "application/json")
```
weissi: I think you should add `("Connection", "close")` as you're closing the connection after receiving the response.

Sources/NIOHTTP1Client/main.swift
```diff
@@ -0,0 +1,100 @@
...
+    func channelRead(ctx: ChannelHandlerContext, data: NIOAny) {
+        let httpResponsePart = unwrapInboundIn(data)
+        switch httpResponsePart {
+        case .head(let httpResponseHeader):
+            print("\(httpResponseHeader.version) \(httpResponseHeader.status.code) \(httpResponseHeader.status.reasonPhrase)")
+            for (name, value) in httpResponseHeader.headers {
+                print("\(name): \(value)")
+            }
+        case .body(var byteBuffer):
+            if let responseBody = byteBuffer.readString(length: byteBuffer.readableBytes) {
```
weissi: this can fail with UTF-8 strings (let's assume the server sends `"🎉"` in chunks of one byte, that would be `f0`, `9f`, `8e`,  `89` then this would print five times `⍰` instead of 🎉.)

but I might be fine with that, CC @Lukasa

Lukasa: Yeah, maybe we should skip `print` and instead write bytes to stdout?

weissi: sounds like a plan to me.

Sources/NIOHTTP1Client/main.swift
mike-neck: Right. But if I add this header, `HTTPResponseDecoder` signals `fireChannelInactive` earlier than `fireChannelRead` even though it has pending inputs.

Sources/NIOHTTP1Client/main.swift
mike-neck: I changed in 10c09d5.

Sources/NIOHTTP1Client/main.swift
mike-neck: I changed in 10c09d5.

Sources/NIOHTTP1Client/main.swift
mike-neck: I changed in 10c09d5 to use `Connection` header, but currently it seems that this handler's `channelRead` is not called.

Sources/NIOHTTP1Client/main.swift
mike-neck: Thanks! I removed `print` and use stdout in 39b519a.

Sources/NIOHTTP1Client/main.swift
Lukasa: Are you sure? Please try updating to the current master, because when I run this patch against current master it seems that channelRead is called fine.

Lukasa: Aha, the issue appears to be intermittent.

Lukasa: Ok, there's a straightforward logic bug in this module. I'll fix it up.

weissi: awesome job @mike-neck and @Lukasa for finding this issue and fixing it!

mike-neck: @Lukasa Thanks it's working fine!

weissi:
@mike-neck so sorry, I totally didn't see the github notifications. If you wouldn't mind fixing the conflicts and making it work for NIO 2 (mostly renames), then we should get this merged.

mike-neck:
It may become long(2 or 3 days) to rebase this branch and to make it work for NIO 2, because there are many changes in repository.

mike-neck:
@weissi 
I fixed compile error on NIO2. But this code doesn't work, because `NIOSSLClientHandler` closes a connection before `channelRead` of `HTTPResponseDecoder` is called.

weissi:
@mike-neck thanks very much for updating! I assume you've been testing on macOS? If yes, you're probably blocked on #72. At this moment, on macOS we can't find the trust roots so cert validation will fail...

weissi:
@mike-neck yes, I can confirm: I just tested on Linux and everything was fine after applying the diff below. You forgot to rename some `ctx: ChannelHandlerContext` to `context: ChannelHandlerContext` but that's all (aside from #72 not being merged yet)

```
diff --git a/Sources/NIOHTTP1Client/main.swift b/Sources/NIOHTTP1Client/main.swift
index ac39867..dd87ebb 100644
--- a/Sources/NIOHTTP1Client/main.swift
+++ b/Sources/NIOHTTP1Client/main.swift
@@ -30,7 +30,7 @@ private final class HTTPResponseHandler: ChannelInboundHandler {
 
     typealias InboundIn = HTTPClientResponsePart
 
-    func channelRead(ctx: ChannelHandlerContext, data: NIOAny) {
+    func channelRead(context: ChannelHandlerContext, data: NIOAny) {
         let httpResponsePart = unwrapInboundIn(data)
         switch httpResponsePart {
         case .head(let httpResponseHeader):
@@ -43,21 +43,21 @@ private final class HTTPResponseHandler: ChannelInboundHandler {
                 FileHandle.standardOutput.write(data)
             }
         case .end(_):
-            closeFuture = ctx.channel.close()
+            closeFuture = context.channel.close()
             promise.succeed(())
         }
     }
 
-    func channelInactive(ctx: ChannelHandlerContext) {
+    func channelInactive(context: ChannelHandlerContext) {
         if closeFuture == nil {
-            closeFuture = ctx.channel.close()
+            closeFuture = context.channel.close()
             promise.fail(ChannelError.inputClosed)
         }
     }
 
-    func errorCaught(ctx: ChannelHandlerContext, error: Error) {
+    func errorCaught(context: ChannelHandlerContext, error: Error) {
         print("Error: ", error)
-        closeFuture = ctx.channel.close()
+        closeFuture = context.channel.close()
         promise.succeed(())
     }
 }
```

mike-neck:
@weissi Yes, I had run this on macOS.

weissi:

weissi:

mike-neck:
@weissi Thanks I confirmed this works on ubuntu(Docker).

weissi:
@mike-neck thanks! it should also now work on macOS (and did on my machine)

PR #41, opened 2018-09-18
Make importing CNIOOpenSSL safer.

Lukasa:
Motivation:

There are a number of dangerous portions of CNIOOpenSSL that make it
tricky to use. This mostly doesn't manifest in debug mode, or in this
module, but when compiling in release mode or when others use this module
it can all go subtly wrong.

We should make this module easier to use correctly.

Modifications:

- Remove the constant defined in the header file, replace with #define.
- Move all static inline declarations to regular function declarations.

Result:

Fewer compile errors, easier to use this module.

### DIFF
--- a/Sources/CNIOOpenSSL/helpers.c
+++ b/Sources/CNIOOpenSSL/helpers.c
@@ -11,16 +11,7 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 //===----------------------------------------------------------------------===//
-// Note that we don't include our umbrella header because it declares a bunch of
-// static inline functions that will cause havoc if they end up in multiple
-// compilation units.
-//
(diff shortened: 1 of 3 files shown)


taskcruncher:
For what it's worth, this commit fixed a fatal error when attempting to create a simple Vapor cloud instance.

The following error no longer appears when calling 'vapor cloud deploy' (and deploy succeeds i/o fails):
/usr/bin/ld.gold: error: /vapor/xxx-production/code/.build/x86_64-unknown-linux/release/NIOOpenSSL.build/SSLContext.swift.o: multiple definition of 'CNIOOpenSSL_SSL_OP_NO_TLSv1_3'

weissi:
@taskcruncher thanks for reporting back, much appreciated

PR #52, opened 2018-11-06
Prevent crashes when closing twice.

Lukasa:
Motivation:

If a user called close on a channel without attaching a promise, and
subsequently called close again with a promise, before the TLS shutdown
completes, this would lead to a crash due to a force-unwrap. It has
been my experience that crashes are less than ideal, especially when
users have done nothing wrong.

Modifications:

Replaced the force-unwrap with code that correctly persists the promise.
Added a test.

Result:

Less likely to crash.

### DIFF
--- a/Sources/NIOOpenSSL/OpenSSLHandler.swift
+++ b/Sources/NIOOpenSSL/OpenSSLHandler.swift
@@ -151,8 +151,10 @@ public class OpenSSLHandler : ChannelInboundHandler, ChannelOutboundHandler {
         case .closing:
             // We're in the process of TLS shutdown, so let's let that happen. However,
             // we want to cascade the result of the first request into this new one.
-            if let promise = promise {
-                closePromise!.futureResult.cascade(promise: promise)
+            if let promise = promise, let closePromise = self.closePromise {
+                closePromise.futureResult.cascade(promise: promise)
+            } else if let promise = promise {
+                self.closePromise = promise
             }
         case .idle:
             state = .closed
--- a/Tests/NIOOpenSSLTests/OpenSSLIntegrationTest.swift
+++ b/Tests/NIOOpenSSLTests/OpenSSLIntegrationTest.swift
@@ -1245,4 +1245,42 @@ class OpenSSLIntegrationTest: XCTestCase {
 
         XCTAssertEqual(certificates.count, 1)
     }
+
+    func testRepeatedClosure() throws {
+        let serverChannel = EmbeddedChannel()
+        let clientChannel = EmbeddedChannel()
+
+        defer {
+            // We expect both cases to throw
+            XCTAssertThrowsError(try serverChannel.finish())
+            XCTAssertThrowsError(try clientChannel.finish())
+        }
+
+        let ctx = try assertNoThrowWithValue(configuredSSLContext())
+
+        XCTAssertNoThrow(try serverChannel.pipeline.add(handler: OpenSSLServerHandler(context: ctx)).wait())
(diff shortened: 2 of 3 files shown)


Tests/NIOOpenSSLTests/OpenSSLIntegrationTest.swift
```diff
@@ -1245,4 +1245,42 @@ class OpenSSLIntegrationTest: XCTestCase {
...
         XCTAssertEqual(certificates.count, 1)
     }
+
+    func testRepeatedClosure() throws {
+        let serverChannel = EmbeddedChannel()
+        let clientChannel = EmbeddedChannel()
+
+        defer {
+            // We expect both casees to throw
+            _ = try? serverChannel.finish()
```
weissi: `XCTAssertNoThrow(XCTAssertEqual(false, serverChannel.finish()))`?

weissi: oops, should read comments :)

weissi: then the other way around, `XCTAssertThrows`

Lukasa: Addressed.

Sources/NIOOpenSSL/OpenSSLHandler.swift
```diff
@@ -151,8 +151,10 @@ public class OpenSSLHandler : ChannelInboundHandler, ChannelOutboundHandler {
         case .closing:
             // We're in the process of TLS shutdown, so let's let that happen. However,
             // we want to cascade the result of the first request into this new one.
-            if let promise = promise {
-                closePromise!.futureResult.cascade(promise: promise)
+            if let promise = promise, let closePromise = self.closePromise {
```
normanmaurer: should we maybe refactor:

```swift
if let promise = promise {
    if let closePromise = self.closePromise {
        closePromise.futureResult.cascade(promise: promise)
    } else {
        self.closePromise = promise
    }
}
```

Lukasa: I thought about it, but decided against it. In particular, I don't believe the compiler will generate better code under this refactor: the compiler should be aware that optional unwrapping does not have side effects, so it can evaluate the unwrap only once. If the compiler won't generate more efficient code, I think we should go for clarity, and avoiding the extra indentation seems clearer to me.

Lukasa: That said, @weissi may disagree.

weissi: tbh I don't think better/worse code matters here. But yes, the compiler will know that unwrapping an optional doesn't have side-effects. I'd also go for clarity here.

normanmaurer: I actually think the suggestion is clearer ... but I am not strong here so works for me :)

weissi: maybe @Lukasa should toss a coin :P

Lukasa: My proposal is that I find the current construction clearer, and as I spend more time in this codebase than @normanmaurer that might be more important ;)

weissi: works for me

normanmaurer: @Lukasa haha... can't disagree with this ;) Like I said I am not strong here so ignore me buddy

PR #54, opened 2018-11-07
Support removing TLS handlers from live connections.

Lukasa:
Motivation:

In some unusual protocol scenarios it is desirable to remove a TLS
tunnel from an active connection. Currently NIOOpenSSL does not
support doing this: the only way to get it to issue a CLOSE_NOTIFY
is to call close() on the Channel. This is less than ideal.

This patch will continue to fail tests until apple/swift-nio#651 lands in a release. At that time, I'll update this repo to require the new SwiftNIO release, and everything should be hunky-dory.

Modifications:

- Added a OpenSSLHandler.stopTLS method
- Reworked the internal state machine to handle this use-case.

Result:

Users can safely unwrap TLS from a Channel.


Lukasa:
@weissi ready for review

Tests/NIOOpenSSLTests/UnwrappingTests.swift
```diff
@@ -0,0 +1,750 @@
...
+    func testSimpleUnwrapping() throws {
+        let serverChannel = EmbeddedChannel()
+        let clientChannel = EmbeddedChannel()
+
+        var clientClosed = false
+        var serverClosed = false
+        var unwrapped = false
+
+        defer {
+            // We expect the server case to throw
```
weissi: can we assert that then with `XCTAssertThrows`?

Tests/NIOOpenSSLTests/UnwrappingTests.swift
```diff
@@ -0,0 +1,750 @@
...
+            _ = try? serverChannel.finish()
```
weissi: assert more stuff?

Tests/NIOOpenSSLTests/UnwrappingTests.swift
```diff
@@ -0,0 +1,750 @@
...
+    func testUnwrappingFollowedByClosure() throws {
+        let serverChannel = EmbeddedChannel()
+        let clientChannel = EmbeddedChannel()
+
+        var clientClosed = false
+        var clientUnwrapped = false
+
+        defer {
+            // Both channels will already be closed
+            _ = try? serverChannel.finish()
```
weissi: in which case we can just assert `XCTAssertFalse(serverChannel.isActive)` no?

Tests/NIOOpenSSLTests/UnwrappingTests.swift
```diff
@@ -0,0 +1,750 @@
...
+    func testUnwrappingMeetsTCPFIN() throws {
+        let serverChannel = EmbeddedChannel()
+        let clientChannel = EmbeddedChannel()
+
+        var clientClosed = false
+        var clientUnwrapped = false
+
+        defer {
+            // The errors here are expected
+            _ = try? serverChannel.finish()
```
weissi: asserts for that?

Tests/NIOOpenSSLTests/UnwrappingTests.swift
```diff
@@ -0,0 +1,750 @@
...
+
+    func testDoubleUnwrapping() throws {
+        let serverChannel = EmbeddedChannel()
+        let clientChannel = EmbeddedChannel()
+
+        var promiseCalled = false
+
+        defer {
+            // We expect the server case to throw
+            _ = try? serverChannel.finish()
```
weissi: asserts for that too?

Issue #44, opened 2018-10-01
Deprecated usage of `changeCapacity`


armcknight:
I got a warning in a build log when building my project:

```
/Users/andrew/Projects/backend/vapor-postgres/postdb/.build/checkouts/swift-nio-ssl.git--8156502881551636107/Sources/NIOOpenSSL/OpenSSLHandler.swift:270:35: warning: 'changeCapacity(to:)' is deprecated: changeCapacity has been replaced by reserveCapacity
                    receiveBuffer.changeCapacity(to: receiveBuffer.capacity + 1024)
```

Which appears to be related to the change made as part of https://github.com/apple/swift-nio/issues/537. The line of code calling it is here: https://github.com/apple/swift-nio-ssl/blob/5c522115a3696701255b0a086508c9ea892584d4/Sources/NIOOpenSSL/OpenSSLHandler.swift#L270 but I haven't looked for others. Happy to put up a PR for it, and take a look around for other usages while doing so.

Lukasa:
While the call is deprecated, I’m not sure we gain much by removing it other than forcing users to roll forward their NIO version.

shpakovski:
Hello, could you please fix this already? This deprecation warning is a _single_ one in the huge Vapor project with 50 dependencies or so 😆

Lukasa:
I still don't think I've been convinced that "fixing" this is worthwhile. I appreciate that it looks ugly, but in a contest between "my build output looks ugly" and "my code doesn't build", the former is a less compelling complaint than the latter.

As a general rule, I'm happy to roll forward the NIO dependency if doing so fixes a bug. If it silences a deprecation warning (particularly a deprecation warning on a dependency whose release schedule we control, and which is appropriately semantically versioned.

I'm interested in how much people value removing this warning. Right now we call `changeCapacity` to attempt to avoid copying the inbound bytes unnecessarily, as we may be able to `realloc` to get the storage. Would people be willing to accept the performance cost of unconditionally copying the bytes to remove this warning?

shpakovski:
Thanks a lot for the reply Cory, I much appreciate your position. No problem, we will survive! Totally ready to wait until some future bug will result in removing this warning as a perfect side-effect 😄

PR #55, opened 2018-11-14
Deprecations and update to Swift 4.2

fassko:


Lukasa:
Thanks for this patch! Unfortunately, I don't think we'll be merging it.

The deprecation warnings have been discussed in #44, and I believe the current state of this discussion has not swayed me towards wanting to bump the dependency version.

As for Swift 4.2 support, we are intentionally continuing to support Swift 4.0 and later until the release of SwiftNIO 2.0, at which time we will move forward to Swift 5.

Thanks again for the work though!

PR #59, opened 2018-11-30
Use BoringSSL for TLS.

Lukasa:
This is the patch that transitions swift-nio-ssl to use BoringSSL, as discussed in [the Swift forums](https://forums.swift.org/t/rfc-moving-swiftnio-ssl-to-boringssl/18280).

Do not merge this yet! We are actively discussing this with the wider community, and we may yet not merge this in this form.


Sources/CNIOBoringSSL/include/boringssl_prefix_symbols.h
```diff
@@ -0,0 +1,3921 @@
...
+#define EC_POINT_dbl BORINGSSL_ADD_PREFIX(BORINGSSL_PREFIX, EC_POINT_dbl)
+#define EC_POINT_dup BORINGSSL_ADD_PREFIX(BORINGSSL_PREFIX, EC_POINT_dup)
+#define EC_POINT_free BORINGSSL_ADD_PREFIX(BORINGSSL_PREFIX, EC_POINT_free)
+#define EC_POINT_get_affine_coordinates_GFp BORINGSSL_ADD_PREFIX(BORINGSSL_PREFIX, EC_POINT_get_affine_coordinates_GFp)
+#define EC_POINT_invert BORINGSSL_ADD_PREFIX(BORINGSSL_PREFIX, EC_POINT_invert)
+#define EC_POINT_is_at_infinity BORINGSSL_ADD_PREFIX(BORINGSSL_PREFIX, EC_POINT_is_at_infinity)
+#define EC_POINT_is_on_curve BORINGSSL_ADD_PREFIX(BORINGSSL_PREFIX, EC_POINT_is_on_curve)
+#define EC_POINT_mul BORINGSSL_ADD_PREFIX(BORINGSSL_PREFIX, EC_POINT_mul)
+#define EC_POINT_new BORINGSSL_ADD_PREFIX(BORINGSSL_PREFIX, EC_POINT_new)
+#define
```
benlangmuir: For C++ symbols that are not `extern "C"` this is trying to `#define` the mangled name, which probably isn't what you want.  For symbols inside the `bssl` namespace you can probably rename the namespace itself, but I'm not sure about global non-extern-"C" symbols like this one.

Lukasa: These are C symbols. 😁

Lukasa: Ah, yes. Those are artefacts of the script, but it doesn’t actually matter: these functions are not part of the public API, so the fact that they define the mangled name doesn’t matter.

Lukasa:
This is officially no longer WIP. Tagging @weissi to review when he gets a change. Feel free to restrict your review to changes to .swift, .sh, and .py files only.

scripts/build-asm.py
```diff
@@ -0,0 +1,212 @@
+#!/usr/bin/env python
```
weissi: - needs license header
- `sanity.sh` scripts needs to have python file checking

scripts/vendor-boringssl.sh
```diff
@@ -0,0 +1,170 @@
...
+## Licensed under Apache License v2.0
+##
+## See LICENSE.txt for license information
+## See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+##
+## SPDX-License-Identifier: Apache-2.0
+##
+##===----------------------------------------------------------------------===##
+#
+# Copyright 2016, gRPC Authors All rights reserved.
```
weissi: - is this the right way of dealing with the license headers? In `swift-nio` we have exceptions to the copyright headers in `sanity.sh`
- needs `NOTICE.txt` entry

scripts/vendor-boringssl.sh
```diff
@@ -0,0 +1,170 @@
...
+#
+set -eou pipefail
+
+HERE=$(pwd)
+DSTROOT=Sources/CNIOBoringSSL
+TMPDIR=$(mktemp -d /tmp/.workingXXXXXX)
+SRCROOT="${TMPDIR}/src/boringssl.googlesource.com/boringssl"
+
+echo "REMOVING any previously-vendored BoringSSL code"
+rm -rf $DSTROOT/include
```
weissi: quotes missing, can we run this through `shellcheck`?  Especially the lines starting with `rm -rf`

Sources/NIOOpenSSL/SSLPointerTricks.swift
```diff
@@ -1,105 +0,0 @@
-//===----------------------------------------------------------------------===//
```
weissi: ♥️

Lukasa: RIP you awful mess.

Lukasa:
Feedback addressed.

NOTICE.txt
```diff
@@ -41,3 +41,13 @@ This product contains a derivation of the Tony Stone's 'process_test_files.rb'.
...
+
+---
+
+This product contains code derived from grpc-swift.
+
+  * LICENSE (Apache License 2.0):
+    * https://github.com/grpc/grpc-swift/blob/0.7.0/LICENSE
+  * HOMEPAGE:
+    * https://github.com/grpc/grpc-swift
+
```
weissi: we need to add BoringSSL here too

Lukasa: Oh lol, yes, that seems important.

Lukasa: Done.

PR #71, opened 2019-02-28
Locate default ca-certificates on Linux.

Lukasa:
Motivation:

When we moved away from the system copy of OpenSSL, we lost the
compiled-in locations of the default root CAs. We should go looking for
them so that certificate verification still works!

Modifications:

- Added a runtime one-off search to locate the roots.

Result:

Certificate verification works on Linux


Sources/NIOOpenSSL/SSLContext.swift
```diff
@@ -324,6 +305,52 @@ extension SSLContext {
         precondition(result == 1)
     }
     
+
+    private static func setAlpnProtocols(_ protocols: [[UInt8]], context: OpaquePointer) throws {
+        // This copy should be done infrequently, so we don't worry too much about it.
+        let protoBuf = protocols.reduce([UInt8](), +)
```
weissi: are you sure this is right? This will just concatenate all bytes without any separator? Ie. if the input is `["h2", "h2c"]`, the output will be `["h2h2c"]` (in UInt8s instead of chars obvs)

PR #88, opened 2019-03-21
Handle close reentrancy in NIOSSLHandler.

glbrntt:
Motivation:

- If the NIOSSLHandler had reads to fire during a clean shutdown, then
  they could cause an additional reentrant shutdown.
- This would fail a precondition and crash.

Modifications:

- If we're already in the shutdown state then don't perform shutdown again
- Added a test

Result:

More correct code.


weissi:

weissi:

PR #100, opened 2019-04-12
Add a timeout for TLS shutdown.

glbrntt:
Motivation:

The remote peer may not always respond to a close notify. This would
leave the SSL handler waiting to shutdown. The shutdown should timeout
in these cases to ensure shutdown completes.

Modifications:

- Added a shutdown timeout to the NIOSSLHandler, and a configurable
  timeout amount to TLSConfiguration.
- Added a test to validate timing out a shutdown.

Result:

TLS shutdown will complete even if the remote peer does not respond to a
close notify.

### DIFF
--- a/Sources/NIOSSL/SSLErrors.swift
+++ b/Sources/NIOSSL/SSLErrors.swift
@@ -82,6 +82,9 @@ extension NIOSSLError: Equatable {
     }
 }
 
+/// Closing the TLS channel cleanly timed out, so it was closed uncleanly.
+public struct NIOSSLCloseTimedOutError: Error {}
+
 /// An enum that wraps individual BoringSSL errors directly.
 public enum BoringSSLError: Error {
     case noError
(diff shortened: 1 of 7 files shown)


Sources/NIOSSL/SSLErrors.swift
```diff
@@ -58,6 +58,7 @@ public enum NIOSSLError: Error {
     case cannotFindPeerIP
     case readInInvalidTLSState
     case uncleanShutdown
+    case closeTimedOut
```
weissi: this would be SemVer major, you'll need to create a new error type for that :|

glbrntt: Agh -- removed!

Sources/NIOSSL/SSLErrors.swift
```diff
@@ -82,6 +82,8 @@ extension NIOSSLError: Equatable {
     }
 }
 
+public struct NIOSSLCloseTimedOutError: Error {}
```
Lukasa: Can we have a doc comment for this please?

weissi: @Lukasa I told George that he might get away without one given that none of the errors above have a comment ;). But maybe this one warrants a bit more explanation.

glbrntt: Done!

PR #102, opened 2019-04-17
Avoid using un-thread-safe static buffers when printing errors.

Lukasa:
Motivation:

BoringSSL's docs make it pretty clear that ERR_error_string is not safe to use in a
concurrent context. Sadly, NIO is pretty damn concurrent, so we should probably stop using it.

Modifications:

- Rewrite the error printing logic to use ERR_error_string_n which takes a buffer to write
    into.

Result:

Error strings will not be corrupted in multithreaded contexts.


normanmaurer:
@weissi what could go wrong 🤷‍♂️ 🔥

PR #109, opened 2019-06-18
Disable 32-bit assembly on Apple platforms.

Lukasa:
Motivation:

The 32-bit assembly in BoringSSL doesn't assemble correctly on 32-bit
apple platforms. It took us a while to notice because we previously
haven't had any supported 32-bit platforms. NIOSSL still doesn't really
support any 32-bit Apple platforms, as all production use on those
platforms should use NIOTS, but there's no excuse for the compilation
failures really.

Modifications:

Remove assembly acceleration on 32-bit Apple platforms.

Result:

Better compilation on 32-bit Apple platforms.


Package.swift
```diff
@@ -22,7 +22,7 @@ import PackageDescription
 // Sources/CNIOBoringSSL directory. The source repository is at
 // https://boringssl.googlesource.com/boringssl.
 //
-// BoringSSL Commit: ff62b38b4b5a0e7926034b5f93d0c276e55b571d
+// BoringSSL Commit: cef9d3f38d72f13412c79157c25753e22cb05f7e
```
Lukasa: This is an actual change: we updated BoringSSL as part of this.

Sources/CNIOBoringSSL/include/CNIOBoringSSL/base.h
```diff
@@ -52,6 +52,9 @@
 
 #ifndef OPENSSL_HEADER_BASE_H
 #define OPENSSL_HEADER_BASE_H
+#if defined(__APPLE__) && defined(__i386__)
+#define OPENSSL_NO_ASM
+#endif
```
Lukasa: This is a meaningful derived change: I amended the BoringSSL vending script to add this stanza.

scripts/build-asm.py
```diff
@@ -27,7 +27,6 @@
     ('linux', 'aarch64', 'linux64', [], 'S'),
     ('linux', 'x86', 'elf', ['-fPIC', '-DOPENSSL_IA32_SSE2'], 'S'),
     ('linux', 'x86_64', 'elf', [], 'S'),
-    ('mac', 'x86', 'macosx', ['-fPIC', '-DOPENSSL_IA32_SSE2'], 'S'),
```
Lukasa: This is a meaningful change.

scripts/vendor-boringssl.sh
```diff
@@ -196,6 +196,10 @@ do
 done
 namespace_inlines "$DSTROOT/include"
 
+# Removing ASM on 32 bit Apple platforms
+echo "REMOVING assembly on 32-bit Apple platforms"
+gsed -i "/#define OPENSSL_HEADER_BASE_H/a#if defined(__APPLE__) && defined(__i386__)\n#define OPENSSL_NO_ASM\n#endif" "$DSTROOT/include/openssl/base.h"
+
```
Lukasa: This is a meaningful change.

PR #111, opened 2019-06-26
Support renegotiation-based client auth.

### LINKED ISSUE #110: Add support for allowing renegotiation
While TLS renegotiation is basically bad, some users need it for post-handshake auth in HTTPS scenarios (see swift-server/swift-nio-http-client#27). We can allow it with BoringSSL by calling `SSL_set_renegotiate_mode` to `ssl_renegotiate_once`.

We should control this setting via the `TLSConfiguration` struct, and keep the current default behaviour of "forbidden".

Lukasa:
Motivation:

While TLS renegotiation is in general a bad idea, it is widely used for
delayed TLS-based client authentication. We are able to support this
use-case, and so we should enable it.

As with BoringSSL, we'd like to continue to default to not supporting
this mode of operation. However, users are free to enable support if
they so choose.

Modifications:

- Added flags for configuring renegotiation.
- Added integration test harness.
- Added renegotiation integration test with s_server.
- Added renegotiation and client cert support to sample client.

Result:

Users will be able to perform TLS renegotiation.
Resolves #110

### DIFF
--- a/Sources/NIOSSL/TLSConfiguration.swift
+++ b/Sources/NIOSSL/TLSConfiguration.swift
@@ -62,6 +62,23 @@ public enum CertificateVerification {
     case fullVerification
 }
 
+/// Support for TLS renegotiation.
+///
+/// In general, renegotiation should not be enabled except in circumstances where it is absolutely necessary.
+/// Renegotiation is only supported in TLS 1.2 and earlier, and generally does not work very well. NIOSSL will
+/// disallow most uses of renegotiation: the only supported use-case is to perform post-connection authentication
+/// *as a client*. There is no way to initiate a TLS renegotiation in NIOSSL.
+public enum NIORenegotiationSupport {
+    /// No support for TLS renegotiation. The default and recommended setting.
+    case none
+
+    /// Allow renegotiation exactly once. If you must use renegotiation, use this setting.
+    case once
+
+    /// Allow repeated renegotiation. To be avoided.
+    case always
+}
+
 /// A secure default configuration of cipher suites for TLS 1.2 and earlier.
 ///
 /// The goal of this cipher suite string is:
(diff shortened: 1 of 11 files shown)


Lukasa:
@weissi A note, we probably need to remember to actually run the integration tests.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -62,6 +62,23 @@ public enum CertificateVerification {
     case fullVerification
 }
 
+/// Support for TLS renegotiation.
+///
+/// In general, renegotiation should not be enabled except in circumstances where it is absolutely necessary.
+/// Renegotiation is only supported in TLS 1.2 and earlier, and generally does not work very well. NIOSSL will
+/// disallow most uses of renegotiation: the only supported use-case is to perform post-connection authentication
+/// *as a client*. There is no way to initiate a TLS renegotiation in NIOSSL.
+public enum NIORenegotiationSupport {
```
weissi: suggestions: Couldn't we nest this type in `TLSConfiguration` so ie would be `TLSConfiguration.RenegotiationSupport`? Then it doesn't need the `NIO` prefix either

Lukasa: The only reason I put the type here was to keep it in line with the other types used by `TLSConfiguration`, which are at the top level. If you think that consistency is less important than the convenience of nesting the type, I'm more than happy to move it: I don't really care either way.

weissi: @Lukasa I really don't mind either. Your call!

Lukasa: Let's leave it where it is: status quo trumps convenience when it comes to the number of types we're discussing. I'll file an issue to move them all into the TLSConfiguration struct for v3.

PR #117, opened 2019-07-16
workaround SR-11125: @_implementationOnly import CNIOBoringSSL

### LINKED ISSUE #116: build error on Ubuntu 18, Swift 5.1 : "reference to 'SHA256_CTX' is ambiguous"
I'm trying to build swift-nio-ssl on an Ubuntu 18 machine, using Swift 5.1, as part of a server based on Vapor 4 alpha (yes I know). Everything builds fine except for the following error :

`<unknown>:0: error: reference to 'SHA256_CTX' is ambiguous
/srv/datadisk01/www/VideoSaverServer/.build/checkouts/swift-nio-ssl/Sources/CNIOBoringSSL/include/CNIOBoringSSL/base.h:411:32: note: candidate found by name lookup is 'SHA256_CTX'
typedef struct sha256_state_st SHA256_CTX;
                               ^
//usr/include/openssl/sha.h:56:3: note: candidate found by name lookup is 'SHA256_CTX'
} SHA256_CTX;
  ^
Failed to reconstruct type for $sSo10SHA256_CTXaD
Original type:
(type_alias_type decl=CAPNSOpenSSL.(file).SHA256_CTX)`

Is that error expected somehow ? I have libssl 1.1 installed on this server, on which the vapor package depends.

weissi:
Motivation:

If we don't import CNIOBoringSSL @_implementationOnly, then some details
of libssl's C structs might leak into NIOSSL's ABI which might lead to
clashes further down the line.

Modification:
- make all CNIOBoringSSL(Shims) imports @_implementationOnly
- remove all concrete libssl types from instance variables

Result:
- fixes #116
- works around SR-11125
- thanks @jrose-apple for suggesting this

### DIFF
--- a/Sources/NIOSSL/SSLCertificate.swift
+++ b/Sources/NIOSSL/SSLCertificate.swift
@@ -12,8 +12,13 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if compiler(>=5.1) && compiler(<5.2)
+@_implementationOnly import CNIOBoringSSL
+@_implementationOnly import CNIOBoringSSLShims
+#else
 import CNIOBoringSSL
 import CNIOBoringSSLShims
+#endif
 import NIO
 
 /// A reference to a BoringSSL Certificate object (`X509 *`).
@@ -26,7 +31,11 @@ import NIO
 /// to obtain an in-memory representation of a TLS certificate from a buffer of
 /// bytes or from a file path.
 public class NIOSSLCertificate {
-    internal let ref: UnsafeMutablePointer<X509>
+    internal let _ref: UnsafeMutableRawPointer/*<X509>*/
+
+    internal var ref: UnsafeMutablePointer<X509> {
+        return self._ref.assumingMemoryBound(to: X509.self)
+    }
 
     internal enum AlternativeName {
         case dnsName([UInt8])
@@ -39,7 +48,7 @@ public class NIOSSLCertificate {
     }
 
     private init(withReference ref: UnsafeMutablePointer<X509>) {
-        self.ref = ref
+        self._ref = UnsafeMutableRawPointer(ref) // erasing the type for @_implementationOnly import CNIOBoringSSL
     }
 
     /// Create a NIOSSLCertificate from a file at a given path in either PEM or
@@ -102,7 +111,7 @@ public class NIOSSLCertificate {
     ///
     /// In general, however, this function should be avoided in favour of one of the convenience
     /// initializers, which ensure that the lifetime of the `X509` object is better-managed.
-    static public func fromUnsafePointer(takingOwnership pointer: UnsafeMutablePointer<X509>) -> NIOSSLCertificate {
+    static func fromUnsafePointer(takingOwnership pointer: UnsafeMutablePointer<X509>) -> NIOSSLCertificate {
         return NIOSSLCertificate(withReference: pointer)
     }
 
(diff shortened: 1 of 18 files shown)


Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -39,7 +48,7 @@ public class NIOSSLCertificate {
     }
 
     private init(withReference ref: UnsafeMutablePointer<X509>) {
-        self.ref = ref
+        self._ref = .init(ref)
```
Lukasa: While I appreciate the desire to make this concise, when it comes to working with the pointer types I really prefer being able to see what the actual type is at the point of initialisation, rather than finding the declaration. Can we use the longer form initializer here?

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -102,7 +111,7 @@ public class NIOSSLCertificate {
     ///
     /// In general, however, this function should be avoided in favour of one of the convenience
     /// initializers, which ensure that the lifetime of the `X509` object is better-managed.
-    static public func fromUnsafePointer(takingOwnership pointer: UnsafeMutablePointer<X509>) -> NIOSSLCertificate {
+    static func fromUnsafePointer(takingOwnership pointer: UnsafeMutablePointer<X509>) -> NIOSSLCertificate {
```
Lukasa: This is technically a breaking API change. Do we want to ship a 3.0 here? Or do we think this can be made an exception?

Sources/NIOSSL/SSLPrivateKey.swift
```diff
@@ -112,10 +116,14 @@ func globalBoringSSLPassphraseCallback(buf: UnsafeMutablePointer<CChar>?,
...
-    internal let ref: UnsafeMutablePointer<EVP_PKEY>
+    internal let _ref: UnsafeMutableRawPointer /*<EVP_PKEY>*/
+
+    internal var ref: UnsafeMutablePointer<EVP_PKEY> {
+        return self._ref.assumingMemoryBound(to: EVP_PKEY.self)
+    }
 
     private init(withReference ref: UnsafeMutablePointer<EVP_PKEY>) {
-        self.ref = ref
+        self._ref = .init(ref)
```
Lukasa: Same comment about the initialisation here.

Sources/NIOSSL/SSLPublicKey.swift
```diff
@@ -21,10 +25,14 @@ import CNIOBoringSSL
...
-    private let ref: UnsafeMutablePointer<EVP_PKEY>
+    private let _ref: UnsafeMutableRawPointer /*<EVP_PKEY>*/
+
+    private var ref: UnsafeMutablePointer<EVP_PKEY> {
+        return self._ref.assumingMemoryBound(to: EVP_PKEY.self)
+    }
 
     fileprivate init(withOwnedReference ref: UnsafeMutablePointer<EVP_PKEY>) {
-        self.ref = ref
+        self._ref = .init(ref)
```
Lukasa: Same comment about the initialisation here.

Sources/NIOSSL/SSLCertificate.swift
weissi: Yup, my vote goes for exception. There was never a legal way to obtain an `UnsafeMutablePointer<X509>` (with the _right_ `X509`...).

Lukasa: I think I agree.

PR #119, opened 2019-07-31
don't crash if user tried to use private key without password

weissi:
Motivation:

Previously, if the user tried to open a private key that had a password
but didn't specify one, we would just crash.

Modifications:

Throw an error instead of crashing if no password was supplied but one
was needed.

Result:

Fewer crashes.

### DIFF
--- a/Tests/NIOSSLTests/SSLPrivateKeyTests.swift
+++ b/Tests/NIOSSLTests/SSLPrivateKeyTests.swift
@@ -25,9 +25,11 @@ class SSLPrivateKeyTest: XCTestCase {
     static var dynamicallyGeneratedKey: NIOSSLPrivateKey! = nil
 
     override class func setUp() {
-        SSLPrivateKeyTest.pemKeyFilePath = try! dumpToFile(text: samplePemKey)
+        SSLPrivateKeyTest.pemKeyFilePath = try! dumpToFile(text: samplePemKey,
+                                                           fileExtension: ".pem")
         SSLPrivateKeyTest.derKeyFilePath = try! dumpToFile(data: sampleDerKey)
-        SSLPrivateKeyTest.passwordPemKeyFilePath = try! dumpToFile(text: samplePemRSAEncryptedKey)
+        SSLPrivateKeyTest.passwordPemKeyFilePath = try! dumpToFile(text: samplePemRSAEncryptedKey,
+                                                                   fileExtension: ".pem")
         SSLPrivateKeyTest.passwordPKCS8PemKeyFilePath = try! dumpToFile(text: samplePKCS8PemPrivateKey)
 
         let (_, key) = generateSelfSignedCert()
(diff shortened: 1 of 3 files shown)


Tests/NIOSSLTests/SSLPrivateKeyTests.swift
```diff
@@ -273,4 +273,29 @@ class SSLPrivateKeyTest: XCTestCase {
             XCTFail("Unexpected error: \(error)")
         }
     }
+
+    func testWrongPassword() {
+        XCTAssertThrowsError(try NIOSSLPrivateKey(buffer: [Int8](samplePemRSAEncryptedKey.utf8CString), format: .pem) {
+            closure in closure("incorrect password".utf8)
+        }) { error in
+            switch error as? NIOSSLError {
```
weissi: this is obviously less than ideal: I created a new error and depending on the situation you either get the existing one or the new one...

There's at least two ways out:
1. remove the new error and continue using the existing one (doesn't have the nice error description though which really helps if password is wrong/missing)
2. move everything to the new error (with the error stack) and deprecate the old error
3. leave it as is in the PR.

Sources/NIOSSL/SSLErrors.swift
```diff
@@ -106,6 +114,10 @@ public enum BoringSSLError: Error {
     case unknownError(NIOBoringSSLErrorStack)
     case invalidSNIName(NIOBoringSSLErrorStack)
     case failedToSetALPN(NIOBoringSSLErrorStack)
+
+    public static func failedToUsePrivateKey(_ errorStack: NIOBoringSSLErrorStack) -> FailedToUsePrivateKeyError {
```
weissi: wasn't quite sure if that should go into `BoringSSLError` or `NIOSSLError`. Went with `BoringSSLError` because there were other errors holding an error stack :P

Tests/NIOSSLTests/SSLPrivateKeyTests.swift
Lukasa: This might be a good time to move everything to the new error, tbh. We really need an extensive error solution, now is probably a good time to do it. What do you think?

weissi: @Lukasa agreed. I will do this on Monday. This PR isn't critical so it can sit until then.

weissi:
@Lukasa ok, updated

Tests/NIOSSLTests/SSLPrivateKeyTests.swift
```diff
@@ -273,4 +273,30 @@ class SSLPrivateKeyTest: XCTestCase {
...
+            case .failedToLoadPrivateKey:
```
Lukasa:
```suggestion
            case .some(.failedToLoadPrivateKey):
```

Tests/NIOSSLTests/SSLPrivateKeyTests.swift
```diff
@@ -273,4 +273,30 @@ class SSLPrivateKeyTest: XCTestCase {
...
+            }
+        }
+    }
+
+    func testMissingPassword() {
+        let configuration = TLSConfiguration.forServer(certificateChain: [],
+                                                       privateKey: .file(SSLPrivateKeyTest.pemKeyFilePath))
+        XCTAssertThrowsError(try NIOSSLContext(configuration: configuration)) { error in
+            switch error as? NIOSSLError {
+            case .failedToLoadPrivateKey:
```
Lukasa:
```suggestion
            case .some(.failedToLoadPrivateKey):
```

weissi:
should be addressed by https://github.com/apple/swift-nio/pull/1093

Issue #121, opened 2019-08-07
NIOSSLCertificate silently drops intermediate certificates when loading a PEM from memory


kevints:
I have a .pem file that contains the certificate chain my server will present (a leaf certificate and an intermediate CA certificate). However, when I load it from the filesystem myself the NIOSSLCertificate API doesn't parse the full chain. Worse, it silently drops the intermediary certificate.
```swift
// Load the certificate chain as a Data from somewhere
let pemEncodedCertificateChain = try FileManager.default.contents(atPath: "./chain.pem")
let certificate = try NIOSSLCertificate(buffer: .init(pemEncoded), format: .pem)
let source: NIOSSLCertificateSource = .certificate(certificate)
let config = TLSConfiguration.forServer(certificateChain: [source], ...)
// This appears to work, but now my server won't present its intermediate CA certificate
```

But when we allow NIOSSL to load the file itself it reads the whole chain out (which is what we want). The cardinality of certificateChain is wrong though, since chain.pem contains 2 certificates.
```swift
// Let NIO load the certificate chain from the filesystem itself
let config = TLSConfiguration.forServer(certificateChain: [.file("./chain.pem")], ...)
```

It seems like there should be a way to get the second behavior with a buffer I loaded myself (`chain.pem` might not be on the filesystem).

Lukasa:
Agreed, this API needs to be dramatically cleaned up. We can emulate the logic BoringSSL uses to load multiple certs ourselves, and then we should probably retire this API and force users to turn their files in to `NIOSSLCertificate` objects through our own helper methods.

Lukasa:
Fixed by #123.

PR #123, opened 2019-08-09
Add loading of multiple certificates from a single PEM file/buffer

glbrntt:
Motivation:

It is only possible to load multiple certificates from a PEM file,
sometimes it is useful to load multiple certificates from a buffer.

Modifications:

- Deprecate the `.file` source of loading a certificate, since it would
  only load the first certificate in a file.
- Provide methods for loading multiple certificates from a PEM file or
  buffer.

Result:

Multiple certificates can be loaded from a PEM file/buffer.


glbrntt:
See also: #121

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -189,6 +189,76 @@ extension NIOSSLCertificate {
...
+        }
+        defer {
+            CNIOBoringSSL_BIO_free(bio)
+        }
+
+        try path.utf8CString.withUnsafeBufferPointer {
+            guard let baseAddress = $0.baseAddress, CNIOBoringSSL_BIO_read_filename(bio, baseAddress) > 0 else {
+                throw NIOSSLError.failedToLoadCertificate
+            }
+        }
```
Lukasa: This can probably be simplified: you should be able to just do:

```swift
guard CNIOBoringSSL_BIO_read_filename(bio, &path) > 0 else {
    throw NIOSSLError.failedToLoadCertificate
}
```

Sources/NIOSSL/SSLErrors.swift
```diff
@@ -186,3 +186,24 @@ public enum NIOTLSUnwrappingError: Error {
...
+    return Int((packedError >> 24) & 0xff)
+}
+
+/// Get the reason code of a packed error.
+///
+/// - Parameter packedError: A packed error.
+/// - Returns: The reason code of the packed error.
+internal func BoringSSL_ERR_GET_REASON(_ packedError: UInt32) -> Int {
+    return Int(packedError & 0xfff)
+}
```
Lukasa: Rather than define these here, can you move them into `CNIOBoringSSLShims` and then defer to the macros? That will ensure that if BoringSSL changes the internal representation nothing breaks.

Sources/NIOSSL/SSLCertificate.swift
glbrntt: Neat, I didn't realise this was possible! (No `&` required though)

PR #127, opened 2019-08-12
Use UInt8 instead of Int8 for buffer types

kevints:
Motivation:

The underlying BoringSSL BIO APIs used by NIOSSLCertificate,
NIOSSLPrivateKey take a raw byte buffer and a count,
and do not need the input buffers to be NUL-terminated. However, the API
currently takes an `Array<Int8>` (which is a typealias of `Array<CChar>` on
platforms where NIO is primarily developed). This suggests an incorrect
completion in Xcode (`utf8CString`) which includes a trailing NUL byte.
In fact, all of the NIOSSL unit tests make this mistake. The more common
use case of creating a these types from a `Data`, `ByteBuffer`, or
`String.UTF8View` is made more awkward as these are easy to convert to
`[UInt8]` but harder to convert to `[CChar]` without extending them with
an unnessary NUL byte.

Modifications:

Add new `bytes:` overloads to the convenience initializers and deprecate
the `buffer:` convenience initalizers. Use these new initializers in the
unit tests.

Result:

A more ergnomic way to correctly create these types from `Data`,
`ByteBuffer`, and `String.UTF8View`.

### DIFF
--- a/Sources/NIOSSL/SSLCertificate.swift
+++ b/Sources/NIOSSL/SSLCertificate.swift
@@ -78,8 +78,17 @@ public class NIOSSLCertificate {
 
     /// Create a NIOSSLCertificate from a buffer of bytes in either PEM or
     /// DER format.
+    ///
+    /// - SeeAlso: `NIOSSLCertificate.init(bytes:format:)`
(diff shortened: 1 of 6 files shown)


Lukasa:

PR #133, opened 2019-09-24
swift 5.1 docker setup

tomerd:
motivation: support swift 5.1

changes:
* use formal swift docker images
* refactor docker-compose to seperate unit tests, integration tests and full test suites
* add docker compose file for swift 5.1, will add CI job once merged
* update swiftformat version to 0.40.12


docker/docker-compose.1804.51.yaml
```diff
@@ -0,0 +1,19 @@
+version: "3"
+
+services:
+
+  runtime-setup:
+    image: swift-nio-ssl:18.04-5.1
```
weissi: same comment here as in http2. We should have one of them on 16.04 and the other one on 18.04

PR #137, opened 2019-10-03
Use .note.GNU-stack section in ASM.

Lukasa:
Motivation:

Well-formed Linux assembly code declares a .note.GNU-stack section in
order to help out the linker. When absent, the linker makes some
suboptimal linking decisions that can cause pain in running programs.

Modifications:

- Amend the vendoring script to add the .note.GNU-stack section to all
    assembly files on Linux.
- Updated BoringSSL

Results:

Better linker behaviour.


Lukasa:

Issue #140, opened 2019-10-08
CNIOBoringSSL compiler error on aarch64 Linux with Swift 5.1 and Swift 5.0.1


wlisac:
Hi there,

I'm attempting to build SwiftNIO SSL on aarch64 (ARMv8) Linux, but I'm running into a compiler error.

```plain
[199/450] Compiling CNIOBoringSSL crypto/fipsmodule/sha256-armv8.linux.aarch64.S
/swift-nio-ssl/Sources/CNIOBoringSSL/crypto/fipsmodule/sha256-armv8.linux.aarch64.S:76:10: error: expected compatible register or logical immediate
 tst w16,#ARMV8_SHA256
         ^
The command '/bin/sh -c swift build --jobs 1' returned a non-zero code: 1
```

This looks like it might be related to https://github.com/apple/swift-nio-ssl/issues/124.

### Steps to Reproduce
One way to reproduce this error is to compile SwiftNIO SSL using Docker Desktop for macOS with an aarch64 base image. Docker Desktop for macOS will use qemu to emulate the aarch64 architecture.

I maintain Docker images for Swift on ARM at [wlisac/swift-on-balena](https://github.com/wlisac/swift-on-balena) that we can use for the aarch64 base image.

**Dockerfile**

```Dockerfile
ARG FROM_IMAGE

FROM $FROM_IMAGE

RUN git clone https://github.com/apple/swift-nio-ssl.git

WORKDIR swift-nio-ssl

RUN swift package resolve

RUN swift build --jobs 1
```

**Build using Swift 5.1**
```bash
$ docker build --build-arg FROM_IMAGE=wlisac/aarch64-swift:5.1-build .
```

**Build using Swift 5.0.1**
```
$ docker build --build-arg FROM_IMAGE=wlisac/aarch64-swift:5.0.1-build .
```

Alternatively, you can reproduce by building SwiftNIO SSL on aarch64 hardware directly.

### Full Build Logs
**Swift 5.1**
```
docker build --build-arg FROM_IMAGE=wlisac/aarch64-swift:5.1-build .
Sending build context to Docker daemon  10.75kB
Step 1/6 : ARG FROM_IMAGE
Step 2/6 : FROM $FROM_IMAGE
 ---> fc67c19563ee
Step 3/6 : RUN git clone https://github.com/apple/swift-nio-ssl.git
 ---> Using cache
 ---> f92f27d89c07
Step 4/6 : WORKDIR swift-nio-ssl
[...]

PR #141, opened 2019-10-09
Fix BoringSSL compilation on aarch64 ARM Linux

wlisac:
This fixes issue #140 

The `patch-2-arm-arch.patch` file was manually updated and the `vendor-boringssl.sh` script was used to apply the patch.

I specifically avoided updating the BoringSSL SHA checkout, but I'd be happy to bump that if that is the preferred approach (both approaches successfully compiled for me on aarch64).


Lukasa:

PR #144, opened 2019-10-23
Remove NIO 2.9.0 build warnings.

Lukasa:
Motivation:

We were one of the few first-party users of
writeWithUnsafeMutableWritableBytes, and so we hit some build warnings.
We should use the nice interface to request the number of bytes we
actually want.

Modifications:

- Remove build warnings.
- Correctly ask to read a whole SSL record instead of a fragment of one.

Result:

Better code

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -35,7 +35,7 @@ let package = Package(
 MANGLE_END */
     ],
     dependencies: [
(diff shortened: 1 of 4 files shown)


Lukasa:
Looks like a spurious failure: the tests all ran and succeeded.

Lukasa:

PR #152, opened 2019-11-25
Add regression test for executable stacks.

Lukasa:
Motivation:

We want the CI to make sure we don't accidentally regress the executable
stack property of the binary.

Modifications:

Run execstack against our build output in the integration tests.

Result:

We shouldn't regress the executable stacks.


weissi:

Lukasa:

PR #153, opened 2019-11-27
Add allocation counter tests.

Lukasa:
Motivation:

To do proper performance work it's necessary to have insight into what
programs are actually doing. One of the two major axes the SwiftNIO
project uses to measure is heap allocations. Currently swift-nio-ssl has
no introspection of heap allocations, so doing informed performance work
is very hard.

Modifications:

- Added the allocation counter test framework.
- Added two allocation counter tests.

Result:

Better idea of what's going on in heap allocations.

### DIFF
--- a/IntegrationTests/tests_02_allocation_counters/test_01_resources/run-nio-ssl-alloc-counter-tests.sh
+++ b/IntegrationTests/tests_02_allocation_counters/test_01_resources/run-nio-ssl-alloc-counter-tests.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+##===----------------------------------------------------------------------===##
+##
+## This source file is part of the SwiftNIO open source project
+##
+## Copyright (c) 2019 Apple Inc. and the SwiftNIO project authors
+## Licensed under Apache License v2.0
+##
+## See LICENSE.txt for license information
+## See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+##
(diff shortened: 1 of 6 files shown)


Lukasa:
Hey @tomerd, can we get the build timeout on the build boxes for swift-nio-ssl increased? We're regularly bumping into build timeouts, and the more testing infrastructure I add the more timeouts we'll see.

IntegrationTests/tests_02_allocation_counters/test_01_resources/run-nio-ssl-alloc-counter-tests.sh
```diff
@@ -0,0 +1,50 @@
...
+)
+
+"$nio_checkout/swift-nio/IntegrationTests/allocation-counter-tests-framework/run-allocation-counter.sh" \
+    -p "$here/../../.." \
+    -m NIO \
+    -m NIOSSL \
+    -s "$here/shared.swift" \
+    -t "$tmp_dir" \
+    -d <( echo '.package(url: "https://github.com/apple/swift-nio.git", from: "2.0.0"),' ) \
+    "$here"/test_*.swift
```
glbrntt: Marli updated the equivalent script in NIO to you can specify which tests to run instead of all of them, can we get that here too?

Lukasa:
Updated, thanks @glbrntt.

tomerd:

PR #154, opened 2019-11-27
Add benchmarks.

Lukasa:
Motivation:

When working with performance its useful to know what your approximate
throughput is. One of the ways we find this out in NIO is to add
benchmarks that let you know what the approximate runtime of a simple
sample program is. NIOSSL doesn't have any of these: it should.

Modifications:

Bring across the benchmark framework from other NIO projects.
Add two benchmarks mirroring the proposed allocation counter tests.

Result:

Insight into performance.


Lukasa:
@tomerd If we merge these, can we get infrastructure added to run them as we have for NIO and NIOHTTP2?

Package.swift
```diff
@@ -30,6 +30,7 @@ let package = Package(
         .library(name: "NIOSSL", targets: ["NIOSSL"]),
         .executable(name: "NIOTLSServer", targets: ["NIOTLSServer"]),
         .executable(name: "NIOSSLHTTP1Client", targets: ["NIOSSLHTTP1Client"]),
+        .executable(name: "NIOSSLPerformanceTester", targets: ["NIOSSLPerformanceTester"]),
```
glbrntt: Do we actually want to export this as a product? (https://github.com/apple/swift-nio/issues/1217)

Lukasa: Ah well spotted, probably not.

PR #155, opened 2019-11-28
Refactor doUnbufferWrites reduce memory pressure

Lukasa:
Motivation:

The current implementation of doUnbufferWrites has the unfortunate habit
of both causing a substantial amount of ARC traffic and also allocating
a promise for all writes, even when one isn't necessary. Worse, if there
are many promises in a single set of flushes then we will also have to
allocate a future callback to cascade each result. All of this leads to
excessive pressure on the memory systems in Swift.

Modifications:

Refactor the code to avoid excessive use of closures, reducing ARC.
Rewrite promise chaining to reduce allocations of closure contexts and
promises.

Result:

Better performance in doUnbufferWrites

While I didn't get around to merging #153 or #154 yet, this reduces the allocations in "many_writes" by 2 per iteration. Unfortunately, that's only about 1%, due to the number of allocations coming out of BoringSSL. This is also matched with a 1% perf improvement.

### DIFF
--- a/Sources/NIOSSL/NIOSSLHandler.swift
+++ b/Sources/NIOSSL/NIOSSLHandler.swift
@@ -581,42 +581,22 @@ extension NIOSSLHandler {
         var promises: [EventLoopPromise<Void>] = []
         var didWrite = false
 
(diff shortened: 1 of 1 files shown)


glbrntt:

PR #159, opened 2019-12-04
Update BoringSSL to 134fb89c4f9da7903d9aa81c6bd1d3c466782de1

Lukasa:


Lukasa:
@swift-nio-bot test perf please

Lukasa:
> It's probably possible to create a wrapper around BoringSSL (maybe this is in the pipeline), which is maybe a good idea...?

It's not. BoringSSL doesn't have a stable API, and to provide one either requires an abstraction at a very high level (a lot of work) or to expose too much of the internal details. I haven't yet come up with a way to handle this that doesn't lead immediately and inexorably to dependency hell.

Lukasa:
[This Swift Forums thread](https://forums.swift.org/t/rfc-moving-swiftnio-ssl-to-boringssl/18280) also contains a lengthy exposition of the issues involved.

PR #160, opened 2019-12-04
Add regression limits for allocation tests.

Lukasa:
Motivation:

We missed these out, which means we failed to detect the massive
allocation win in #159.

Modifications:

Correctly add allocation limits.

Result:

We'll know if we regress.


Lukasa:

PR #166, opened 2019-12-13
Remove unsafe code from hostname verification

Lukasa:
Motivation:

The hostname verification code operated on unsafe buffer pointers
without any clear justification for doing so. While it may have been
necessary to do that once, it certainly isn't necessary now.

Modifications:

Rewrite all the code to operate on Array.
Throw a more meaningful error.

Result:

Bounds checking, better safety.

### DIFF
--- a/Sources/NIOSSL/IdentityVerification.swift
+++ b/Sources/NIOSSL/IdentityVerification.swift
@@ -14,39 +14,18 @@
 
 import NIO
 
-private let ASCII_PERIOD: UInt8 = ".".utf8.first!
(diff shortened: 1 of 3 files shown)


Lukasa:
@swift-nio-bot test perf please

Lukasa:
No meaningful performance changes from adding the array. This is to be expected, it's not really hot code, but it does run on basically every client TLS handshake so it's welcome to have the safety back.

Lukasa:
I also probably bought back a bunch of perf by reducing ARC traffic and array constructions.

weissi:
looks great, thank you

PR #169, opened 2020-01-10
Add hashability to Certificate and PrivateKey

Lukasa:
Motivation:

When types are equatable its important to ask why they aren't hashable.
In the case of SSLCertificate and SSLPrivateKey there isn't any
particuarly good reason other than that their hashing implementations
are a bit fussy. We should support it, even if they're tricky.

Modifications:

Added Hashability to NIOSSLCertificate and NIOSSLPrivateKey.
Fixed NIOSSLPrivateKey equatable implementation.

Result:

Hashable keys and certs.

### DIFF
--- a/Sources/NIOSSL/SSLCertificate.swift
+++ b/Sources/NIOSSL/SSLCertificate.swift
@@ -204,28 +204,7 @@ extension NIOSSLCertificate {
     /// - returns: The DER-encoded bytes for this certificate.
     /// - throws: If an error occurred while serializing the certificate.
     public func toDERBytes() throws -> [UInt8] {
-        guard let bio = CNIOBoringSSL_BIO_new(CNIOBoringSSL_BIO_s_mem()) else {
-            throw NIOSSLError.unableToAllocateBoringSSLObject
-        }
-
-        defer {
-            CNIOBoringSSL_BIO_free(bio)
-        }
-
-        let rc = CNIOBoringSSL_i2d_X509_bio(bio, self.ref)
-        guard rc == 1 else {
-            let errorStack = BoringSSLError.buildErrorStack()
-            throw BoringSSLError.unknownError(errorStack)
-        }
-
-        var dataPtr: UnsafeMutablePointer<CChar>? = nil
-        let length = CNIOBoringSSL_BIO_get_mem_data(bio, &dataPtr)
-
-        guard let bytes = dataPtr.map({ UnsafeMutableRawBufferPointer(start: $0, count: length) }) else {
-            throw NIOSSLError.unableToAllocateBoringSSLObject
-        }
-
-        return Array(bytes)
+        return try self.withUnsafeDERCertificateBuffer { Array($0) }
     }
 
     /// Create an array of `NIOSSLCertificate`s from a buffer of bytes in PEM format.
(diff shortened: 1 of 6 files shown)


Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -284,6 +284,44 @@ extension NIOSSLCertificate: Equatable {
...
+        // A SHA256 digest is 256 bits long, or 32 bytes, or 4 UInt64s. We store this
+        // as UInt64s here, but they are always accessed as UInt8s.
+        private var storage: (UInt64, UInt64, UInt64, UInt64)
+
+        init() {
+            self.storage = (0, 0, 0, 0)
+        }
+
+        mutating func withUnsafeMutablePointer<T>(_ body: (UnsafeMutableBufferPointer<UInt8>) throws -> T) rethrows -> T {
+            return try Swift.withUnsafeMutablePointer(to: &self.storage) { (storagePtr: UnsafeMutablePointer<(UInt64, UInt64, UInt64, UInt64)>) -> T in
```
ianpartridge: Does Swift guarantee that tuples are laid out contiguously and won't potentially be reordered/split in memory then?

Lukasa: Yes, tuples are guaranteed to be laid out the same way as C arrays. This is why C arrays bridge into Swift as tuples. :smile:

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -284,6 +284,44 @@ extension NIOSSLCertificate: Equatable {
...
+                return try storagePtr.withMemoryRebound(to: UInt8.self, capacity: 1) { bytesPtr -> T in
```
weissi: `capacity: 1` isn't right here, should be `<(UInt64, ...)>.size`, right?

Lukasa: I honestly don't know. The [docs](https://developer.apple.com/documentation/swift/unsafepointer/2430863-withmemoryrebound) say:

> **`count`**
>   The number of instances of `Pointee` to bind to `type`.

Here, `type` is the name of the argument with the label `to`, which implies that I don't want to say how many of `to` I want, I want to say how many of the original elements I am binding, but I don't think it's very clear. @atrick, can you clarify?

atrick: The docs have this note:
```
  /// - Note: Only use this method to rebind the pointer's memory to a type
  ///   with the same size and stride as the currently bound `Pointee` type.
  ///   To bind a region of memory to a type that is a different size, convert
  ///   the pointer to a raw pointer and use the `bindMemory(to:capacity:)`
  ///   method.
```
See [SR-11082: Support withMemoryRebound(to:) with mismatched sizes](https://bugs.swift.org/browse/SR-11082).

Until then, the most "proper" thing to do would be call bind memory yourself, or create your own withMemoryRebound wrapper for it:

```
let toPtr = UnsafeMutableRawPointer(fromPtr).bindMemory(to: ToType.self, capacity: numElementsOfToType)

// do some stuff

UnsafeMutableRawPointer(toPtr).bindMemory(to: FromType.self, capacity: numElementsOfFromType)
```

Lukasa: Huh, weird. Thanks for weighing in, I’ll update the patch accordingly!

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -284,6 +284,44 @@ extension NIOSSLCertificate: Equatable {
...
+
+
+    public func hash(into hasher: inout Hasher) {
+        // Conveniently BoringSSL gives us a useful function for this: we can ask it to provide a digest of the X509 object.
+        // That's what we're going to do here. We use SHA256 here because it's secure and fast.
+        var digest = SHA256Digest()
+
+        digest.withUnsafeMutablePointer {
+            var length = CUnsignedInt(0)
+            let rc = CNIOBoringSSL_X509_digest(self.ref, CNIOBoringSSL_EVP_sha256(), $0.baseAddress, &length)
```
vlm: I think conversion to ASN.1 DER with a subsequent `hasher(bytes: )` is more efficient than this. The X509_digest does that conversion into DER (with a corresponding memory allocation) internally, then we're applying SHA256 on top, where for all practical purposes a simpler byte hashing would be significantly faster than properly digesting using a cryptographically sound hash.

Just a nit.

Lukasa: Yeah, I had wondered about making that change. It’s always more annoying to have to write more code though. :wink: I’ll sit with the idea this weekend and decide how much more time I want to spend on the process.

Lukasa:
Ok, I've updated this patch to simply hash the DER bytes, as this is going to be substantially less fragile going forward.

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -305,6 +284,36 @@ extension NIOSSLCertificate {
 
         return certificates
     }
+
+    /// Calls the given body function with a temporary buffer containing the DER-encoded bytes of this
+    /// certificate. This function does allocate for these bytes, but there is no way to avoid doing so with the
+    /// X509 API in BoringSSL.
+    ///
+    /// The pointer provided to the closure is not valid beyond the lifetime of this method call.
+    private func withUnsafeDERCertificateBuffer<T>(_ body: (UnsafeRawBufferPointer) throws -> T) throws -> T {
```
weissi: this should be `rethrows` I think and we should remove the `throw NIOSSLError.unableToAlloc` I think

Sources/NIOSSL/SSLPrivateKey.swift
```diff
@@ -289,8 +289,57 @@ public class NIOSSLPrivateKey {
...
 
+
+// MARK:- Utilities
+extension NIOSSLPrivateKey {
+    /// Calls the given body function with a temporary buffer containing the DER-encoded bytes of this
+    /// private key. This function does allocate for these bytes, but there is no way to avoid doing so with the
+    /// X509 API in BoringSSL.
+    ///
+    /// The pointer provided to the closure is not valid beyond the lifetime of this method call.
+    private func withUnsafeDERBuffer<T>(_ body: (UnsafeRawBufferPointer) throws -> T) throws -> T {
```
weissi: rethrows here too?

PR #171, opened 2020-01-14
Provide better certificate verification callback

Lukasa:
Motivation:

We've exposed the OpenSSL certificate verify callback for a while now.
Unfortunately, this callback is fairly hostile to callers, as it is
called repeatedly through the certificate verification process, instead
of presenting the unified certificate chain to the user.

BoringSSL has a better verification callback available that we use
internally, but that is not exposed to users. It would be nice to give
users the better API.

Modifications:

- Expose the BoringSSL verification callback to users.
- Re-plumb the Security.framework verification to use our public API.

Result:

Easier certificate chain verification overrides.


Sources/NIOSSL/NIOSSLClientHandler.swift
```diff
@@ -31,6 +31,11 @@ private extension String {
 /// This handler can be used in channels that are acting as the client
 /// in the TLS dialog. For server connections, use the `NIOSSLServerHandler`.
 public final class NIOSSLClientHandler: NIOSSLHandler {
+    public convenience init(context: NIOSSLContext, serverHostname: String?) throws {
+        try self.init(context: context, serverHostname: serverHostname, optionalCustomVerificationCallback: nil)
+    }
+
+    @available(*, deprecated, renamed: "init(context:serverHostname:customVerificationCallback:)")
```
glbrntt: This isn't strictly a rename because the callback type changed, right?

Sources/NIOSSL/SSLCallbacks.swift
```diff
@@ -58,9 +58,34 @@ public enum NIOSSLVerificationResult {
...
+/// Please be cautious with calling out from this method. This method is always invoked on the event loop,
+/// so you must not block or wait. However, you may perform asynchronous work by leaving the event loop context:
+/// when the verification is complete you must complete the provided `EventLoopPromise`.
+///
+/// This method must take care to ensure that it does not cause any `ChannelHandler` to recursively call back into
+/// the `NIOSSLHandler` that triggered it, as making re-entrant calls into BoringSSL is not supported by SwiftNIO and
+/// leads to undefined behaviour. It is acceptable to leave the event loop context and then call into the `NIOSSLHandler`,
+/// as this will not be re-entrant.
+///
+/// Note that setting this callback will override _all_ verification logic that
```
glbrntt: ... that BoringSSL provides?

Sources/NIOSSL/SSLConnection.swift
```diff
@@ -422,6 +453,17 @@ extension SSLConnection {
...
     }
+
+    /// The certificate chain presented by the peer.
+    func peerCertificateChain() throws -> [NIOSSLCertificate] {
+        return try self.withPeerCertificateChainBuffers { buffers in
+            guard let buffers = buffers else {
+                return []
+            }
+
+            return try buffers.map { try NIOSSLCertificate(bytes: $0, format: .der)}
```
glbrntt:
```suggestion
            return try buffers.map { try NIOSSLCertificate(bytes: $0, format: .der) }
```

Sources/NIOSSL/SSLContext.swift
```diff
@@ -411,7 +424,7 @@ extension NIOSSLContext {
             throw BoringSSLError.unknownError(errorStack)
         }
         #elseif os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
-        CNIOBoringSSL_SSL_CTX_set_custom_verify(context, SSL_VERIFY_PEER, securityFrameworkCustomVerify)
+        // We do this later.
```
glbrntt: nit: can you add that it's done in createConnection?

Sources/NIOSSL/NIOSSLClientHandler.swift
Lukasa: That's true, but it provides a useful fixit.

Lukasa:
Thanks @glbrntt, I think I addressed all but one of your comments.

Sources/NIOSSL/SSLCallbacks.swift
```diff
@@ -108,3 +133,116 @@ extension KeyLogCallbackManager {
...
+        // We need to attach our "do the thing" callback. This will always invoke the "ask me again" API, and it will do so in a separate
+        // event loop tick to avoid awkward re-entrancy with this method.
+        promise.futureResult.whenComplete { result in
+            // When we complete here we need to set our result state, and then ask to respin certificate verification.
+            // If we can't respin verification because we've dropped the parent handler, that's fine, no harm no foul.
+            // For that reason, we tolerate both the verify manager and the parent handler being nil.
+            eventLoop.execute {
+                // Note that we don't close over self here: that's to deal with the fact that this is a struct, and we don't want to
+                // escape the mutable ownership of self.
+                connection.customVerificationManager?.result = .complete(NIOSSLVerificationResult(result))
```
weissi: could you assert here that we're actually still in the state that we want to be in before overwriting the values here?

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -110,6 +110,35 @@ public class NIOSSLCertificate {
...
+            ref = CNIOBoringSSL_PEM_read_bio_X509(bio, nil, nil, nil)
+        case .der:
+            ref = CNIOBoringSSL_d2i_X509_bio(bio, nil)
+        }
+
+        if ref == nil {
+            throw NIOSSLError.failedToLoadCertificate
+        }
+
+        self.init(withReference: ref!)
```
weissi: the memory ownership here looks interesting. This can only be correct if `CNIOBoringSSL_PEM_read_bio_X509` and `CNIOBoringSSL_d2i_X509_bio` return a +1'd `ref` that is valid without the `bio`. If that's the case, I'd like to rename the `withReference` argument to `takingOwnership` to make this 100% clear.

Sources/NIOSSL/SSLConnection.swift
```diff
@@ -156,6 +160,32 @@ internal final class SSLConnection {
         }
     }
 
+    func setCustomVerificationCallback(_ callbackManager: CustomVerifyManager) {
+        // Store the verification callback. We need to do this to keep it alive throughout the connection.
+        // We'll drop this when we're told that it's no longer needed to ensure we break the reference cycles
+        // that this callback inevitably produces.
+        self.customVerificationManager = callbackManager
```
weissi: should we assert `self.customVerificationManager == nil` here?

Sources/NIOSSL/SSLContext.swift
```diff
@@ -411,7 +424,7 @@ extension NIOSSLContext {
             throw BoringSSLError.unknownError(errorStack)
         }
         #elseif os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
-        CNIOBoringSSL_SSL_CTX_set_custom_verify(context, SSL_VERIFY_PEER, securityFrameworkCustomVerify)
+        // We do this later, in createConnection.
```
weissi: this looks slightly odd to `#elseif` a certain set of platforms here just to put a comment in there. Shouldn't this just be `#else` or no `#else` at all and a comment "for non-Linux platforms, we do this later, in createConnection."?

Lukasa:
Thanks @weissi, I've addressed that feedback.

PR #172, opened 2020-01-14
Add API for dumping certificate bytes.

Lukasa:
Motivation:

In some cases it's useful to be able to ask questions about what the
certificate looks like. We don't have a full X509 API, so the best way
to handle this for now is to allow users to extract the bytes of the
certificate so they can use other libraries to do the job.

Modifications:

- Added a function for dumping certificate bytes.

Result:

Users can have the certificate bytes.


Lukasa:

PR #176, opened 2020-01-31
Avoid curried thunks (workaround SR-12115)

glbrntt:
Calling `function(otherFunction)` allocates, but
`function({ otherFunction($0) }` does not.

See: SR-12115 and SR-12116.

Modifications:

Avoid passing functions directly; pass them in new closures instead.

Result:

Fewer allocations.


Lukasa:

Issue #164, opened 2019-12-11
Try replacing explicit Equatable implementations with derived ones


Lukasa:
I think most of our previously-broken Equatable conformances (see #163) could probably be replaced with synthesised equatable conformances.

PR #186, opened 2020-02-19
Replace NIOSSLError.unableToAllocateBoringSSLObject with fatalError

### LINKED ISSUE #173: deprecate & remove throw NIOSSLError.unableToAllocateBoringSSLObject; replace by crash
instead of `throw NIOSSLError.unableToAllocateBoringSSLObject`, we should (like Swift does for all other failing memory allocations) just crash instead.

agnosticdev:
Motivation:

Provides a resolution to issue-173.
deprecate & remove throw NIOSSLError.unableToAllocateBoringSSLObject;
replace by crash #173

Modifications:

Removed all NIOSSLError.unableToAllocateBoringSSLObject and replaced
with fatalError to force a crash.

Result:

- In critical points where execution should not continue a fatalError will be thrown instead of an NIOSSLError.unableToAllocateBoringSSLObject.
- fixes #173

### DIFF
--- a/Sources/NIOSSL/SSLErrors.swift
+++ b/Sources/NIOSSL/SSLErrors.swift
@@ -52,7 +52,7 @@ public typealias NIOBoringSSLErrorStack = [BoringSSLInternalError]
 /// Errors that can be raised by NIO's BoringSSL wrapper.
 public enum NIOSSLError: Error {
     case writeDuringTLSShutdown
-    @available(*, deprecated, message: "unableToAllocateBoringSSLObject is no longer available.")
+    @available(*, deprecated, message: "unableToAllocateBoringSSLObject can no longer be thrown")
     case unableToAllocateBoringSSLObject
     case noSuchFilesystemObject
     case failedToLoadCertificate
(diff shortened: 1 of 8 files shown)


Sources/NIOSSL/SSLPKCS12Bundle.swift
```diff
@@ -64,7 +64,7 @@ public struct NIOSSLPKCS12Bundle {
             cert.map { CNIOBoringSSL_X509_free($0) }
             pkey.map { CNIOBoringSSL_EVP_PKEY_free($0) }
             caCerts.map { CNIOBoringSSL_sk_X509_pop_free($0, CNIOBoringSSL_X509_free) }
-            throw NIOSSLError.unableToAllocateBoringSSLObject
+            fatalError("Failed to obtain cert and pkey from a PKC12 file")
```
Lukasa: We can delete the cleanup code above: if we're going to `fatalError`, we gain very little from freeing this memory first.

weissi:
@agnosticdev this looks great, thank you! Mind deprecating the `unableToAllocateBoringSSLObject` enum case?

weissi:
@agnosticdev as @Lukasa just pointed out: deprecating the enum case will lead to warnings in the `==` function but if you don't mind, just delete the manual implementation, the Swift compiler should be able to synthesise that now

Lukasa:
See also #164.

agnosticdev:
Thank you for the feedback.  Latest changes just committed.

Sources/NIOSSL/SSLErrors.swift
```diff
@@ -52,6 +52,7 @@ public typealias NIOBoringSSLErrorStack = [BoringSSLInternalError]
 /// Errors that can be raised by NIO's BoringSSL wrapper.
 public enum NIOSSLError: Error {
     case writeDuringTLSShutdown
+    @available(*, deprecated, message: "unableToAllocateBoringSSLObject is no longer available.")
```
Lukasa: Can we reword this to say "can no longer be thrown"?

agnosticdev: Sounds good.  Would this be just "unableToAllocateBoringSSLObject can no longer be thrown," or just "can no longer be thrown?"

Lukasa: Let's use the full language.

Also see the discussion in the main thread re: the equatable implementation.

agnosticdev:
Changes made.  Let me know if this is what you had in mind for the removal of `==`.

Lukasa:

Lukasa:
@weissi We believe the API checker is wrong here, correct? While we removed the explicit function, that's an ABI breaker, not an API breaker.

agnosticdev:
Just seen that too.
```
02:35:30 /* Removed Decls */
02:35:30 Func NIOSSLError.==(_:_:) has been removed
```
Let me know if there's anything I can do on my end.

weissi:
@agnosticdev thanks very much for pointing this out. We believe it's a false positive: https://bugs.swift.org/browse/SR-12245

PR #191, opened 2020-02-24
Mitigate Folly/BoringSSL static linking issue.

### LINKED ISSUE #189: Crash when called from react native module
We use GRPC-Swift in our recent project, which imports swift-nio-ssl for channels. 

The implementation of the app called a grpc TLS connection from react-native's native module, the app crashed right away at sdallocx, right after OPENSSL_free(orig_ptr), at CNIOBoringSSL -> crypto -> mem.c -> OPENSSL_realloc. 

It looks like it was caused by boringSSL's context allocating failure. To verify that, we turn off the GRPC's TLS, the app stop crashing.

We also test the swift-grpc by calling directly from a native swift application, using TLS, which also works well without any crash.

We tried to find some similar issues at stripe-ios (https://github.com/stripe/stripe-ios/issues/1293), which eventually was found due to Some of BoringSSL symbols in their code weren't prefixed, causing conflicts with other copies of BoringSSL and OpenSSL in end user apps. Their solution was to add prefix to wide range of their BoringSSLl's symbols. 

So we are wondering that would it be the same reason causing our app's crashing? 

Cheers

Lukasa:
Motivation:

BoringSSL and Folly both attempt to play silly buggers with the sdallocx
symbol defined in jemalloc. Unfortunately, they don't play nicely
together, and the end result if you have both present is that BoringSSL
will reliably SIGBUS or SIGSEGV on initialization. Naturally, this isn't
good.

Modifications:

- Resolve the issue by patching out BoringSSL's allocator optimisation.
- Update BoringSSL to 6432bb46ab44731567ec923e6c8fc182f13d0070.

Result:

Flow and BoringSSL will play better together.

Fixes #189.


scripts/vendor-boringssl.sh
```diff
@@ -264,6 +264,7 @@ echo "RENAMING header files"
 echo "PATCHING BoringSSL"
 git apply "${HERE}/scripts/patch-1-inttypes.patch"
 git apply "${HERE}/scripts/patch-2-arm-arch.patch"
+git apply "${HERE}/scripts/patch-3-weak-linking.patch"
```
normanmaurer: @Lukasa is this something that should redone by BoringSSL itself /cc @davidben ?

Lukasa: Unclear: see the conversation in #189.

Sources/CNIOBoringSSL/crypto/err/err_data.c
```diff
@@ -190,44 +190,44 @@ const uint32_t kOpenSSLReasonValues[] = {
...
-    0x2c382fd6,
-    0x2c38afed,
-    0x2c392ffb,
-    0x2c39b00b,
-    0x2c3a301d,
-    0x2c3ab031,
-    0x2c3b3042,
-    0x2c3bb061,
+    0x2c372fb0,
+    0x2c37afc2,
```
weissi:
```suggestion
    0x2c37afb2,
```

Issue #190, opened 2020-02-23
SwiftNIOSSL CocoaPods Support


Jake-Prickett:
Over the last few days I've been working on adding support for CocoaPods to SwiftNIOSSL and ran into an issue I can't quite work through. I was hoping to get a push in the right direction. 🙂 

`SwiftNIOSSL` depends on `CNIOBoringSSL` & `CNIOBoringSSLShims`, both of which are C/C++ Dependencies.

The problem I'm facing is when I finally lint the podspec, it's not able to get beyond linking for x86_64 architecture in `CNIOBoringSSL`

```sh
[...]

PR #193, opened 2020-02-27
Add Build Podspec Script

Jake-Prickett:
Resolves: https://github.com/apple/swift-nio-ssl/issues/190

- Added standard `list_topsorted_dependencies.sh` script to grab dependencies
- Added `build_podspec.sh` script that will build `SwiftNIOSSL`, `CNIOBoringSSLShims`, & `CNIOBoringSSL`
- Validated that the CNIO Pods lint/compile successfully when integrated
- Will validate/push the final `SwiftNIOSSL.podspec` and `CNIOBoringSSLShims.podspec` Pod after next release

Created: https://cocoapods.org/pods/CNIOBoringSSL


scripts/build_podspec.sh
```diff
@@ -0,0 +1,135 @@
...
+    xcconfig="s.xcconfig = { 'HEADER_SEARCH_PATHS' => 'Sources/${target}/include', 'CLANG_CXX_LANGUAGE_STANDARD' => 'c++11', }"
+    public_header_files="s.public_header_files = 'Sources/${target}/include/*.h'"
+  fi
+
+  cat > "${tmpdir}/${target}.podspec" <<- EOF
+Pod::Spec.new do |s|
+  s.name = '$target'
+  s.version = '$version'
+  s.license = { :type => 'Apache 2.0', :file => 'LICENSE.txt' }
+  s.summary = 'Useful code around SwiftNIO.'
```
weissi: Actually, we should probably fix this summary here :)

scripts/build_podspec.sh
```diff
@@ -0,0 +1,135 @@
...
+newline=$'\n'
+
+here="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+tmpdir=$(mktemp -d /tmp/.build_podspecsXXXXXX)
+
+echo "Building podspec in $tmpdir"
+
+# JP - Think of a better way to handle this.. 
+#      Might be able to leverage the list_topsorted_dependencies.sh script fully
+targets=( "CNIOBoringSSL" "CNIOBoringSSLShims" "SwiftNIOSSL" )
```
weissi: yeah, the `list_topsorted_dependencies.sh` script should do exactly this for you. Let me know if anything doesn't work

scripts/build_podspec.sh
Jake-Prickett: Good catch!

scripts/build_podspec.sh
Jake-Prickett: Weird enough, running with the script actually results in more podspecs than just the libraries:
```bash
targets=( $("${here}/list_topsorted_dependencies.sh" -l -r | sed 's/^NIO/SwiftNIO/') )
```
Podspec files generated:
```
CNIOBoringSSL.podspec
CNIOBoringSSLShims.podspec
SwiftNIO.podspec
SwiftNIOConcurrencyHelpers.podspec
SwiftNIOFoundationCompat.podspec
SwiftNIOHTTP1.podspec
SwiftNIOSSL.podspec
SwiftNIOSSLPerformanceTester.podspec
SwiftNIOTLS.podspec
```

Even with the '-l' option passed 🤔

weissi: whoops, so it also goes into targets of dependencies... That's not good :)

Jake-Prickett: Yeah, now I remember why I put that comment! 😄 I can take on that as a follow-up item? The idea behind the `list_topsorted_dependencies.sh` script is pretty clever, and I'd like to leverage that here for sure.

weissi: added https://github.com/apple/swift-nio/pull/1420 to fix this

Jake-Prickett: Updated here as well 👍

scripts/build_podspec.sh
```diff
@@ -0,0 +1,133 @@
...
+    xcconfig="s.xcconfig = { 'HEADER_SEARCH_PATHS' => 'Sources/${target}/include', 'CLANG_CXX_LANGUAGE_STANDARD' => 'c++11', }"
+    public_header_files="s.public_header_files = 'Sources/${target}/include/*.h'"
+  fi
+
+  cat > "${tmpdir}/${target}.podspec" <<- EOF
+Pod::Spec.new do |s|
+  s.name = '$target'
+  s.version = '$version'
+  s.license = { :type => 'Apache 2.0', :file => 'LICENSE.txt' }
+  s.summary = 'Useful code around SwiftNIOSSL.'
```
weissi:
```suggestion
  s.summary = 'TLS Support for SwiftNIO, based on BoringSSL.'
```

scripts/build_podspec.sh
```diff
@@ -0,0 +1,133 @@
...
+
+  if [ "$target" == "CNIOBoringSSL" ] || [ "$target" == "CNIOBoringSSLShims" ]; then
+    libraries="s.libraries = 'c++'"
+    xcconfig="s.xcconfig = { 'HEADER_SEARCH_PATHS' => 'Sources/${target}/include', 'CLANG_CXX_LANGUAGE_STANDARD' => 'c++11', }"
+    public_header_files="s.public_header_files = 'Sources/${target}/include/*.h'"
+  fi
+
+  cat > "${tmpdir}/${target}.podspec" <<- EOF
+Pod::Spec.new do |s|
+  s.name = '$target'
```
weissi: doesn't this need a `Swift` in front? The pod names should be `SwiftNIOSSL` and `CNIOBoringSSL` to stay in line with SwiftNIO (ie. if the target starts with `NIO`, prepend a `Swift` to make it `SwiftNIO...`; if the target starts with a `C` or anything else, just leave it as is).

Jake-Prickett: Correct me if I'm wrong, I believe this one does need `Swift` in front of it, whereas below for the `module_name` we strip `Swift` from it.

For the SwiftNIOSSL specific podspec we would have:
```
s.name = 'SwiftNIOSSL'
s.module_name = 'NIOSSL'
```

Similarly, for SwiftNIO we have:
```
s.name = 'SwiftNIO'
s.module_name = 'NIO'
```

What do you think?

weissi: yes, that's correct

weissi: @Jake-Prickett is that already what's being generated? if yes, I'm happy :)

Jake-Prickett: Sure is! 👍🏼

weissi: thank you, let's get this merged then.

weissi:

weissi:

Issue #179, opened 2020-02-04
Pop unprocessed data on the read path before closing


Lukasa:
If the user initiates a close on a NIOSSL handler, it's possible that before the response CLOSE_NOTIFY is received we'll receive some application data (as it was already in flight). When that happens today, we'll be in the `.closing` state which will lead us to call `doShutdownStep`. That is fine, as it goes, but BoringSSL will error if there is unprocessed application data in that stage, which is sad and causes data loss.

A _better_ choice is to follow the same as the normal read path and call `doDecodeData`. That method has a special return value for having received EOF (`zeroReturnError`) and already handles it, correctly issuing a shutdown step. This should therefore totally handle the use-case, with almost no code changes required.

Naturally we'll want a test to be sure of this case, but that's easily done.

weissi:
the error a user would see is

```
NIOSSL.BoringSSLError.sslError([Error: 268435747 error:10000123:SSL routines:OPENSSL_internal:APPLICATION_DATA_ON_SHUTDOWN])
```

agnosticdev:
Is this closed now via https://github.com/apple/swift-nio-ssl/pull/198

Lukasa:
Sure is!

Issue #196, opened 2020-03-03
get rid of `do { ... } catch { ... }` for expected errors


weissi:
The

    do {
        try someOperation()
	XCTFail("should throw") // easy to forget
    } catch error as SomethingError {
        XCTAssertEqual(.something, error as? SomethingError)
    } catch {
        XCTFail("wrong error")
    }

pattern is not only very long, it's also very error prone. If you forget
any of the XCTFails, you might not tests what it looks like

    XCTAssertThrowsError(try someOperation) { error in
        XCTAssertEqual(.something, error as? SomethingError)
    }

is much safer and shorter.

We should replace all uses of this pattern in `Tests/**`.

See also https://github.com/apple/swift-nio/pull/1430 which does the same for `swift-nio`.

weissi:
see also:
- https://github.com/apple/swift-nio-extras/issues/83
- https://github.com/apple/swift-nio-ssl/issues/196
- https://github.com/apple/swift-nio-http2/issues/190

Issue #177, opened 2020-01-31
flaky test: NIOSSLIntegrationTest.testForcingVerificationFailureNewCallback


glbrntt:
```
16:37:14 Test Case 'NIOSSLIntegrationTest.testForcingVerificationFailureNewCallback' started at 2020-01-31 16:37:14.321
16:37:14 /code/Tests/NIOSSLTests/NIOSSLIntegrationTest.swift:1341: error: NIOSSLIntegrationTest.testForcingVerificationFailureNewCallback : XCTAssertEqual failed: ("0") is not equal to ("1") - 
```

agnosticdev:
I'm seeing consistent results on macOS and Linux for this test on master.  Is there anything more I need to reproduce this issue?

Lukasa:
It's highly likely this is a timing issue. Running the test in an infinite loop until it fails may be a useful first step.

agnosticdev:
That's what I was wondering.  I wrote a script to run this test over and over.  I try running it on a non-stop loop.  Thanks.

weissi:
failed again in 5.0  CI

```
10:41:57 /code/Tests/NIOSSLTests/NIOSSLIntegrationTest.swift:1408: error: NIOSSLIntegrationTest.testForcingVerificationFailureNewCallback : XCTAssertEqual failed: ("0") is not equal to ("1") - 
10:41:57 Fatal error: Unexpectedly found nil while unwrapping an Optional value
10:41:57 Current stack trace:
```

PR #198, opened 2020-03-07
Fix for Unprocessed Data before Closing

agnosticdev:
Motivation:

Fix for issue #179 that resolves error: APPLICATION_DATA_ON_SHUTDOWN
when data is present on read path while closing.

Modifications:

Minor modification to channelRead and a unit test.

Result:

Closing read paths with data will no longer error.
Data will be processed and closed gracefully.


962367274:
https://github.com/apple/swift-nio-ssl/issues/177#issue-558275626

agnosticdev:
@962367274 this failure continues to elude me on macOS and Linux.  (Even when running the tests in a continuous loop on both platforms) Can you provide more info on how you are reproducing this, and I will take a look?

Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
```diff
@@ -1633,11 +1662,9 @@ class NIOSSLIntegrationTest: XCTestCase {
         do {
             try clientChannel.writeInbound(buffer)
             XCTFail("Did not error")
-        } catch NIOSSLError.shutdownFailed {
-            // expected
         } catch {
-            XCTFail("Unexpected error: \(error)")
-        }
+            // expected
+        }
```
Lukasa: Why are we removing the detailed assertion on the error type?

Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
```diff
@@ -1038,6 +1038,35 @@ class NIOSSLIntegrationTest: XCTestCase {
...
+        let _ = serverChannel.close()
+        
+        // Create a new chunk of data after the close
+        var clientBuffer = clientChannel.allocator.buffer(capacity: 5)
+        clientBuffer.writeStaticString("hello")
+        _ = try clientChannel.writeAndFlush(clientBuffer).wait()
+
+        // Manually take the client outbound buffer and hand it off to the server to validate that an error is not thrown.
+        let clientOutboundBuffer = try clientChannel.readOutbound(as: ByteBuffer.self)!
+        XCTAssertNoThrow(try serverChannel.writeInbound(clientOutboundBuffer))
```
Lukasa: This could do with being a bit stricter about what we expect. If we replace this with `interactInMemory`, then what we expect to see is:

1. The server should read the data the client sent.
2. The server should be closed.
3. The client should be closed.
4. No errors in either pipeline.

Can we add validations for that?

Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
agnosticdev: The reason I removed this is because a new error is being thrown (WRONG_VERSION_NUMBER) after the clientChannel is being closed and I figured that was a result of the change I made and this is testing the old state.  I can add this back and rework if desired.

Lukasa: `WRONG_VERSION_NUMBER` is a surprising error: it suggests we broke something.

agnosticdev:
I can acknowledge the feedback, I will rework and submit an update soon.

Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
agnosticdev: Sounds like I need to add it back and do some investigation:
` error: -[NIOSSLTests.NIOSSLIntegrationTest testReceivingGibberishAfterAttemptingToClose] : failed - Unexpected error: sslError([Error: 268435703 error:100000f7:SSL routines:OPENSSL_internal:WRONG_VERSION_NUMBER])`

Lukasa: Oh, never mind: re-reading the test I think this error is acceptable. We can simply change the test to assert the new error.

Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
```diff
@@ -1633,10 +1663,8 @@ class NIOSSLIntegrationTest: XCTestCase {
         do {
             try clientChannel.writeInbound(buffer)
             XCTFail("Did not error")
-        } catch NIOSSLError.shutdownFailed {
-            // expected
         } catch {
-            XCTFail("Unexpected error: \(error)")
+            XCTAssertEqual(error.localizedDescription, "The operation couldn’t be completed. (NIOSSL.BoringSSLError error 0.)")
```
Lukasa: I think you'll get a better description if you use `.description` instead of `.localizedDescription`.

agnosticdev: I addressed this in the latest commit.  Let me know if this is what you had in mind?

Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
```diff
@@ -1633,10 +1663,10 @@ class NIOSSLIntegrationTest: XCTestCase {
         do {
             try clientChannel.writeInbound(buffer)
             XCTFail("Did not error")
-        } catch NIOSSLError.shutdownFailed {
-            // expected
-        } catch {
-            XCTFail("Unexpected error: \(error)")
+        } catch  {
+            let errorCode = CNIOBoringSSL_ERR_get_error()
```
Lukasa: Hmm, this is going a bit too far the other way. I was thinking more that we'd expect to just use `String(describing: error)`. That should give us a better error message (specifically, tell us which enum case we actually threw instead of lying to us about it being case 0).

agnosticdev: Thanks.  I dialed it in a bit to use `.sslError` and then validate the `.description` off of that error.

Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
```diff
@@ -1038,6 +1038,36 @@ class NIOSSLIntegrationTest: XCTestCase {
...
+
+        // Ok, we're connected. Now we want to close the server, and have that trigger a client CLOSE_NOTIFY.
+        // After the CLOSE_NOTIFY create another chunk of data.
+        let _ = serverChannel.close()
+        
+        // Create a new chunk of data after the close.
+        var clientBuffer = clientChannel.allocator.buffer(capacity: 5)
+        clientBuffer.writeStaticString("hello")
+        _ = try clientChannel.writeAndFlush(clientBuffer).wait()
+        let clientClosePromise = clientChannel.close()
```
Lukasa: Great, let's return to the feedback I gave here. I'd like us to confirm that the server channel actually did receive the client data, _and_ that the channel is now closed. Can we update the test to do that?

Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
```diff
@@ -1664,9 +1664,13 @@ class NIOSSLIntegrationTest: XCTestCase {
             try clientChannel.writeInbound(buffer)
```
weissi: minor: could we use  `XCTAssertThrowsError` here instead of the `do { ... } catch { ... }` pattern? I know that was pre-existing code but see #196

Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
agnosticdev: Feedback addressed and tested on macOS and Linux.

Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
agnosticdev: Feedback addressed and tested on macOS and Linux.

Lukasa:

Lukasa:

PR #206, opened 2020-03-25
Allow users to control the TLS unwrapping behaviour.

Lukasa:
Motivation:

In some highly unusual protocol situations it is possible that TLS
unwrapping may want to time out without taking out the TCP connection.
This is most common when interoperating with peers that don't send
CLOSE_NOTIFY messages. It's a useful tool to be able to request that
timeouts don't take out the TCP connection.

Modifications:

- Added NIOSSLHandler.UnwrappingBehaviour.
- Added support for passing this to stopTLS.

Result:

Users can control unwrapping behaviour.

### DIFF
--- a/Sources/NIOSSL/NIOSSLHandler.swift
+++ b/Sources/NIOSSL/NIOSSLHandler.swift
@@ -313,8 +313,24 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
     /// out. This task should be cancelled if the shutdown does not time out.
     private func scheduleTimedOutShutdown(context: ChannelHandlerContext) -> Scheduled<Void> {
         return context.eventLoop.scheduleTask(in: self.shutdownTimeout) {
-            self.state = .closed
-            self.channelClose(context: context, reason: NIOSSLCloseTimedOutError())
+            switch self.state {
+            case .idle, .handshaking, .active:
+                preconditionFailure("Cannot schedule timed out shutdown on non-shutting down handler")
+
+            case .closed, .unwrapped:
+                // This means we raced with the shutdown completing. We just let this one go: do nothing.
+                return
+
+            case .closing:
+                // We're closing, the only thing we do here is exit.
+                self.state = .closed
+                self.channelClose(context: context, reason: NIOSSLCloseTimedOutError())
+
+            case .unwrapping:
+                // The user only wants us to error and unwrap, not to close.
+                self.state = .unwrapped
+                self.channelUnwrap(context: context, failedWithError: NIOSSLCloseTimedOutError())
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -552,6 +613,54 @@ extension NIOSSLHandler {
...
+    /// this alert, or reply very slowly. If the configured `TLSConfiguration.shutdownTimeout` expires before the remote
+    /// peer sends its reply `CLOSE_NOTIFY`, the `NIOSSLHandler` will give up. The `UnwrapTimeoutBehaviour` controls what
+    /// exactly is done in that "give up" state.
+    public struct UnwrapTimeoutBehaviour {
+        /// If the shutdown timeout expires, an error will be reported on the user promise, the `NIOSSLHandler` will
+        /// `close` the channel, and will also remove itself from the `ChannelPipeline`
+        public static let errorAndClose = UnwrapTimeoutBehaviour(.errorAndClose)
+
+        /// If the shutdown timeout expires, an error will be reported on the user promise, and the `NIOSSLHandler` will
+        /// remove itself from the `ChannelPipeline`, but it will not remove itself from the channel.
```
glbrntt: I don't understand removing the handler from the pipeline but not the channel; how does that work?

Tests/NIOSSLTests/UnwrappingTests.swift
```diff
@@ -953,4 +957,177 @@ final class UnwrappingTests: XCTestCase {
...
+        for level in levels {
+            serverHandler.stopTLS(promise: promiseToAdd, onTimeout: level)
+            promiseToAdd = nil
+        }
+
+        XCTAssertFalse(clientClosed)
+        XCTAssertFalse(serverClosed)
+        XCTAssertFalse(unwrapped)
+
+        // Advance time by 5 seconds. This should fire the timeout. W
```
glbrntt:
```suggestion
        // Advance time by 5 seconds. This should fire the timeout.
```

Sources/NIOSSL/NIOSSLHandler.swift
Lukasa: It doesn't, that should say it will not _close_ the channel.

glbrntt: That makes much more sense; I should've made the connection

Lukasa:
Thanks @glbrntt, both addressed.

Lukasa:
Ok, this patch has been substantially rewritten. Instead of adding new API surface, we now just fire an error down the pipeline and don't close when the unwrap timeout fires. This gives users the same control with less API surface.

glbrntt:
The approval is subject to the CI passing, of course 😃

Lukasa:

PR #216, opened 2020-05-12
Update BoringSSL to 7c522995d1ea5386b3223a19b0f62e73c1f76b17

Lukasa:
Also remove a temporary BoringSSL patch and deal with no C files being
in fiat/


scripts/vendor-boringssl.sh
```diff
@@ -191,7 +191,7 @@ PATTERNS=(
 'crypto/*/*/*.S'
 'crypto/*/*/*/*.c'
 'third_party/fiat/*.h'
-'third_party/fiat/*.c'
+#'third_party/fiat/*.c'
```
glbrntt, line 194: If this remains commented out and files matching this pattern _do_ exist the next time this is run, will there be an obvious failure?

Lukasa: Yeah, the compile will fail due to missing symbols.

glbrntt: Cool, thanks

Lukasa:

Lukasa:

PR #220, opened 2020-06-04
Enable X509_V_FLAG_TRUSTED_FIRST.

Lukasa:
Motivation:

Unfortunately, while BoringSSL is mostly a fundamentally fantastic
library, it has inherited OpenSSL's X509 verifier. This verifier has a
number of issues, but its biggest problem is that it's simply not great
at verifying certificates!

Certificate chains are not so much "chains" as they are a single path
through a DAG. There are potentially many possible paths from a leaf
certificate to a set of roots. Some of these paths are not trusted, some
may be. BoringSSL's default verifier does not do a good job in some edge
cases because it prefers to use the intermediates provided by the client
when it could do a better job by using its own trust store.

As an example, consider the following three-certificate "chain":

CA A -- signs -- Intermediate CA B -- signs -- Leaf C

Imagine that CA A now becomes untrusted, either because it was revoked
or (more topically) because it expires. Imagine also that, in the
meantime, Intermediate CA B has become a fully-fledged CA, and has
re-issued its certificate as a self-signed root using the same key and
identifier.

The mistaken resolver may attempt to use as many certificates as
possible from the peer chain. If the peer chain is using the old,
now-replaced intermediate CA, this will cause BoringSSL to get confused:
it will build a chain that chains up to the intermediate CA, find the
old root in the trust store, and then complain that that root is expired
and fail the validation.

Instead, we want to ignore the intermediate CA B altogether! This is
because we now have a trusted root B', and if we use that then we'll
instantly get a valid chain and can move on.

This behaviour is controlled by a flag, `X509_V_FLAG_TRUSTED_FIRST`.
This flag is set by default in OpenSSL 1.1+. Sadly, it is not set by
default in BoringSSL. This patch resolves that issue.

Note that this issue only affects non-Apple platforms. On Apple
platforms we use the system resolver, which does not have this issue.

Modifications:

- Set `X509_V_FLAG_TRUSTED_FIRST`.
- Add a test to verify that flag is set.

Result:

We'll do a better (but still bad) job of building trust chains.


tiran:
For your information Python's ssl module has been setting the flag by default since 2015. I haven't got a complain yet.

Issue #147, opened 2019-11-04
NIOSSLServerHandler's init is throwing but that's not actionable


weissi:
NIOSSLServerHandler's init is throwing and that has two issues:

1. it's not really actionable so everybody writes `try! NIOSSLServerHandler(context: sslContext)`. If that's what we want, we can just do that in the init.
2. it's almost always used in the `clientChannelInitializer` where you have to return a future...

weissi:
@Lukasa wdyt?

wlisac:
I think I hit a similar challenge using the `NIOSSLClientHandler`.

This is how I solved it, but having an API that works nicer with the `channelInitializer` would be helpful 🙂

```swift
.channelInitializer { channel in
    channel.pipeline.addSSLHandlerIfNeeded(configuration: tlsConfiguration,
                                           serverHostname: host)
}
```

```swift
extension ChannelPipeline {
    fileprivate func addSSLHandlerIfNeeded(configuration: TLSConfiguration?, serverHostname: String) -> EventLoopFuture<Void> {
        guard let configuration = configuration else {
            return eventLoop.makeSucceededFuture(())
        }
        
        do {
            let sslContext = try NIOSSLContext(configuration: configuration)
            let handler = try NIOSSLClientHandler(context: sslContext, serverHostname: serverHostname)
            return addHandler(handler)
        } catch {
            return eventLoop.makeFailedFuture(error)
        }
    }
}
```

weissi:
Thanks @wlisac! I think there also other such APIs that we should improve on.

Lukasa:
The server handler can be made to stop throwing, as its only failure mode is an inability to allocate the BoringSSL object. The client handler is harder: it takes an SNI name and that can potentially be invalid. We can solve that by giving that idea a type that isn’t `String`, which would also allow us to introduce a non-throwing initializer.

Do we know what the API ramifications are of removing `throws` from a method signature? Is it backward compatible? I know it’ll introduce a warning, but that just requires us to have a semver minor.

weissi:
> The server handler can be made to stop throwing, as its only failure mode is an inability to allocate the BoringSSL object.

Awesome.

> The client handler is harder: it takes an SNI name and that can potentially be invalid. We can solve that by giving that idea a type that isn’t `String`, which would also allow us to introduce a non-throwing initializer.

[...]

PR #226, opened 2020-06-16
Add support for custom verify callback to servers.

Lukasa:
Motivation:

In #171 when we worked on providing access to the better verification
callback, we managed to entirely miss that we had not provided that
access to servers. This meant they were stuck only with the
substantially-less-useful old-school callback, instead of the much
better new-school one.

While we're here, as we had to add multiple new initializers to
NIOSSLServerHandler, I took the opportunity to also resolve the server
handler portion of #147. The issue itself is still open because the
client handlers still have throwing inits, but all "preferred"
initializers on NIOSSLServerHandler no longer throw.

Modifications:

- Deprecated NIOSSLServerHandler.init(context:verificationCallback:)
- Implemented two new initializers on NIOSSLServerHandler.
- Added tests to verify that the NIOSSLServerHandler verification
  callback is actually called.
- Removed all now-unnecessary try keywords.

Result:

Users will be able to provide custom verification callbacks that work
much better than they currently can when on the server, and the server
is now back into feature parity with the client.

### DIFF
--- a/Sources/NIOSSLPerformanceTester/BenchManyWrites.swift
+++ b/Sources/NIOSSLPerformanceTester/BenchManyWrites.swift
@@ -34,7 +34,7 @@ final class BenchManyWrites: Benchmark {
     }
 
     func setUp() throws {
-        let serverHandler = try NIOSSLServerHandler(context: self.serverContext)
+        let serverHandler = NIOSSLServerHandler(context: self.serverContext)
         let clientHandler = try NIOSSLClientHandler(context: self.clientContext, serverHostname: "localhost")
         try self.backToBack.client.pipeline.addHandler(clientHandler).wait()
         try self.backToBack.server.pipeline.addHandler(serverHandler).wait()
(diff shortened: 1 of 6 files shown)


Sources/NIOSSLPerformanceTester/BenchManyWrites.swift
```diff
@@ -34,7 +34,7 @@ final class BenchManyWrites: Benchmark {
     }
 
     func setUp() throws {
-        let serverHandler = try NIOSSLServerHandler(context: self.serverContext)
+        let serverHandler = NIOSSLServerHandler(context: self.serverContext)
```
weissi, line 37: 🥳

Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
```diff
@@ -1607,6 +1612,46 @@ class NIOSSLIntegrationTest: XCTestCase {
...
+
+        let handshakeResultPromise = group.next().makePromise(of: Void.self)
+        let handshakeWatcher = WaitForHandshakeHandler(handshakeResultPromise: handshakeResultPromise)
+        let serverChannel: Channel = try serverTLSChannel(context: context,
+                                                          preHandlers: [],
+                                                          postHandlers: [handshakeWatcher],
+                                                          group: group,
+                                                          customVerificationCallback: { _, promise in
+                                                            promise.succeed(.failed)
+              })
```
glbrntt, line 1635: nit: indentation

PR #232, opened 2020-07-05
Add option to specify signining and verification algorithms

fourplusone:
Motivation:

BoringSSLs built-in list of valid signing and verification algortihms is not
always the one you actually would like to use. For example adding or removing
algorithms from this list might increase compatibility as well as security

Modifications:

Add an option to set CNIOBoringSSL_SSL_CTX_set_verify_algorithm_prefs
and CNIOBoringSSL_SSL_CTX_set_signing_algorithm_prefs

Result:

The user can set signing and verification algortihms at their will

### DIFF
--- a/Sources/NIOSSL/SSLContext.swift
+++ b/Sources/NIOSSL/SSLContext.swift
@@ -163,7 +163,25 @@ public final class NIOSSLContext {
         try NIOSSLContext.configureCertificateValidation(context: context,
                                                       verification: configuration.certificateVerification,
                                                       trustRoots: configuration.trustRoots)
-
+        
+        // Configure verification algorithms
+        returnCode = configuration
+            .verifySignatureAlgorithms
+            .map{ $0.rawValue }
+            .withUnsafeBufferPointer { algo in
+            CNIOBoringSSL_SSL_CTX_set_verify_algorithm_prefs(context, algo.baseAddress, algo.count)
+        }
+        precondition(1 == returnCode)
+        
+        // Configure signing algorithms
+        returnCode = configuration
+            .signSignatureAlgorithms
+            .map{ $0.rawValue }
+            .withUnsafeBufferPointer { algo in
+            CNIOBoringSSL_SSL_CTX_set_signing_algorithm_prefs(context, algo.baseAddress, algo.count)
+        }
+        precondition(1 == returnCode)
+        
         // If we were given a certificate chain to use, load it and its associated private key. Before
         // we do, set up a passphrase callback if we need to.
         if let callbackManager = callbackManager {
--- a/Sources/NIOSSL/SSLContext.swift
+++ b/Sources/NIOSSL/SSLContext.swift
@@ -171,7 +171,10 @@ public final class NIOSSLContext {
                 .withUnsafeBufferPointer { algo in
                     CNIOBoringSSL_SSL_CTX_set_verify_algorithm_prefs(context, algo.baseAddress, algo.count)
             }
-            precondition(1 == returnCode)
+            if returnCode != 1 {
+                let errorStack = BoringSSLError.buildErrorStack()
+                throw BoringSSLError.unknownError(errorStack)
+            }
         }
         
         // Configure signing algorithms
@@ -181,7 +184,10 @@ public final class NIOSSLContext {
                 .withUnsafeBufferPointer { algo in
                     CNIOBoringSSL_SSL_CTX_set_signing_algorithm_prefs(context, algo.baseAddress, algo.count)
             }
-            precondition(1 == returnCode)
+            if returnCode != 1 {
+                let errorStack = BoringSSLError.buildErrorStack()
+                throw BoringSSLError.unknownError(errorStack)
+            }
         }
         
         // If we were given a certificate chain to use, load it and its associated private key. Before
--- a/Sources/NIOSSL/SSLContext.swift
+++ b/Sources/NIOSSL/SSLContext.swift
@@ -165,22 +165,24 @@ public final class NIOSSLContext {
                                                       trustRoots: configuration.trustRoots)
         
         // Configure verification algorithms
-        returnCode = configuration
-            .verifySignatureAlgorithms
-            .map{ $0.rawValue }
-            .withUnsafeBufferPointer { algo in
-            CNIOBoringSSL_SSL_CTX_set_verify_algorithm_prefs(context, algo.baseAddress, algo.count)
+        if let verifySignatureAlgorithms = configuration.verifySignatureAlgorithms {
+            returnCode = verifySignatureAlgorithms
+                .map { $0.rawValue }
+                .withUnsafeBufferPointer { algo in
+                    CNIOBoringSSL_SSL_CTX_set_verify_algorithm_prefs(context, algo.baseAddress, algo.count)
+            }
+            precondition(1 == returnCode)
         }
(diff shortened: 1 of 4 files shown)


Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -84,6 +84,55 @@ public enum NIORenegotiationSupport {
...
+    .ecdsaSecp256R1Sha256, .rsaPssRsaeSha256, .rsaPkcs1Sha256,
+    
+    // Larger hashes are acceptable.
+    .ecdsaSecp384R1Sha384, .rsaPssRsaeSha384, .rsaPkcs1Sha384,
+    
+    .rsaPssRsaeSha512, .rsaPkcs1Sha512,
+    
+    // For now, SHA-1 is still accepted but least preferable.
+    .rsaPkcs1Sha1
+]
```
Lukasa, line 116: I would really love to find a way to avoid hard-coding this, as swift-nio-ssl doesn’t have its own preferences about what to use here. Rather than use an explicit default of this kind, let’s default to `nil` instead.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -142,6 +191,11 @@ public struct TLSConfiguration {
     /// The pre-TLS1.3 cipher suites supported by this handler. This uses the OpenSSL cipher string format.
     /// TLS 1.3 cipher suites cannot be configured.
     public var cipherSuites: String
+    
+
```
Lukasa, line 195: Nit: remove one of these empty lines.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -142,6 +191,11 @@ public struct TLSConfiguration {
     /// The pre-TLS1.3 cipher suites supported by this handler. This uses the OpenSSL cipher string format.
     /// TLS 1.3 cipher suites cannot be configured.
     public var cipherSuites: String
+    
+    
+    public var verifySignatureAlgorithms : [SignatureAlgorithm]
+    
+    public var signSignatureAlgorithms : [SignatureAlgorithm]
```
Lukasa, line 198: Three notes here: firstly, SwiftNIO style is to have no leading space before the colon, i.e. `signSignatureAlgorithm: [SignatureAlgorithm]`.

Secondly, we should add some doc comments to these.

Third, let’s change `signSignatureAlgorithms` to `signingSignatureAlgorithms`.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -84,6 +84,55 @@ public enum NIORenegotiationSupport {
     case always
 }
 
+/// Signature algorithms. The values are defined as in TLS 1.3
+public enum SignatureAlgorithm : UInt16 {
```
Lukasa, line 88: Let’s not use an `enum` here: we can’t extend them in semver minors.

This should be a `RawRepresentable` `struct` instead.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -163,7 +163,25 @@ public final class NIOSSLContext {
         try NIOSSLContext.configureCertificateValidation(context: context,
                                                       verification: configuration.certificateVerification,
                                                       trustRoots: configuration.trustRoots)
-
+        
+        // Configure verification algorithms
+        returnCode = configuration
+            .verifySignatureAlgorithms
+            .map{ $0.rawValue }
```
Lukasa, line 170: Minor nit: NIO style is to have a space between `map` and the open brace.

fourplusone:
Thanks for the review. I pushed a new version implementing your suggestions.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -163,7 +163,27 @@ public final class NIOSSLContext {
...
-
+        
+        // Configure verification algorithms
+        if let verifySignatureAlgorithms = configuration.verifySignatureAlgorithms {
+            returnCode = verifySignatureAlgorithms
+                .map { $0.rawValue }
+                .withUnsafeBufferPointer { algo in
+                    CNIOBoringSSL_SSL_CTX_set_verify_algorithm_prefs(context, algo.baseAddress, algo.count)
+            }
+            precondition(1 == returnCode)
```
Lukasa, line 174: It would be better if these preconditions were replaced with errors, ideally ones that used the BoringSSL error stack. `BoringSSL.unknownError` would be fine for this, along with `BoringSSLError.buildErrorStack()` to get the error stack.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -84,6 +84,31 @@ public enum NIORenegotiationSupport {
     case always
 }
 
+/// Signature algorithms. The values are defined as in TLS 1.3
+public struct SignatureAlgorithm : RawRepresentable {
```
Lukasa, line 88: While we're here let's also add `Hashable` to this, it'll make our lives a little easier.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -278,6 +323,8 @@ public struct TLSConfiguration {
                                  keyLogCallback: NIOSSLKeyLogCallback? = nil,
                                  renegotiationSupport: NIORenegotiationSupport) -> TLSConfiguration {
         return TLSConfiguration(cipherSuites: cipherSuites,
+                                verifySignatureAlgorithms: verifySignatureAlgorithms,
+                                signingSignatureAlgorithms: signingSignatureAlgorithms,
```
Lukasa, line 327: So I know this is very annoying, but adding parameters to functions is a semver major change, even when those parameters have default values. We need to define new functions that take these extra arguments, sadly.

Tests/NIOSSLTests/TLSConfigurationTest.swift
```diff
@@ -90,6 +93,9 @@ class TLSConfigurationTest: XCTestCase {
         (cert, key) = generateSelfSignedCert()
         TLSConfigurationTest.cert2 = cert
         TLSConfigurationTest.key2 = key
+
+        TLSConfigurationTest.compatibleSignatureAlgorithm = .rsaPssRsaeSha256
+        TLSConfigurationTest.incompatibleSignatureAlgorithm = .ecdsaSecp384R1Sha384
```
Lukasa, line 98: These don't need to be at the test case scope, we can use them in the test directly.

Tests/NIOSSLTests/TLSConfigurationTest.swift
```diff
@@ -174,6 +180,40 @@ class TLSConfigurationTest: XCTestCase {
...
     }
+    
+    func assertHandshakeSucceeded(withClientConfig clientConfig: TLSConfiguration,
+                                  andServerConfig serverConfig: TLSConfiguration,
+                                  file: StaticString = #file,
+                                  line: UInt = #line) throws {
+        let clientContext = try assertNoThrowWithValue(NIOSSLContext(configuration: clientConfig))
+        let serverContext = try assertNoThrowWithValue(NIOSSLContext(configuration: serverConfig))
+
+        let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
```
Lukasa, line 191: No need for an `MTELG` if we're using `EmbeddedChannel`s.

Lukasa:

Lukasa:

PR #238, opened 2020-08-14
Stop using commonName to validate certificates.

Lukasa:
Motivation:

CommonName as part of certificate hostname verification has been
deprecated for more than two decades. Browsers have finally stopped
using the commonName as part of verification, and the WebPKI forbids
putting names into the commonName that are not also part of the
subjectAlternativeName extension.

Continuing to trust the commonName as part of the certificate hostname
will put us in a weaker position than most of the browser ecosystem.
That's not a good place to be, so we should flip our default to not
trusting the commonName any longer.

For pragmatic reasons, we do allow an override, but the default
configuration is off. Users using this override will likely find
themselves disappointed when we remove it at some point in the future.

Modifications:

- Added TLSConfiguration options to enable trusting commonName fields
  for hostname verification.
- Made investigating the commonName field optional during hostname
  verification.
- Tests to validate the new behaviour.

Result:

No trusting common names by default anymore.

### DIFF
--- a/Sources/NIOSSL/TLSConfiguration.swift
+++ b/Sources/NIOSSL/TLSConfiguration.swift
@@ -212,6 +212,18 @@ public struct TLSConfiguration {
     /// Whether renegotiation is supported.
     public var renegotiationSupport: NIORenegotiationSupport
 
+    /// Whether to trust the commonName field when validating certificate hostnames.
+    ///
+    /// The commonName field of certificates has been deprecated for use as a validation source
+    /// for nearly two decades. As browsers and the WebPKI now forbid the usage of common names
+    /// in certificates, swift-nio-ssl defaults to this strict level of enforcement as well.
+    ///
+    /// However, some older infrastructure does not issue certificates with valid subject alternative
+    /// name fields. For this reason, we currently allow an override for trusting the common name. This
+    /// is a temporary measure to allow users to work around these limitations, but we will remove this
+    /// functionality in future.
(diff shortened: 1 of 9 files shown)


glbrntt:

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -212,6 +212,18 @@ public struct TLSConfiguration {
...
+    ///
+    /// The commonName field of certificates has been deprecated for use as a validation source
+    /// for nearly two decades. As browsers and the WebPKI now forbid the usage of common names
+    /// in certificates, swift-nio-ssl defaults to this strict level of enforcement as well.
+    ///
+    /// However, some older infrastructure does not issue certificates with valid subject alternative
+    /// name fields. For this reason, we currently allow an override for trusting the common name. This
+    /// is a temporary measure to allow users to work around these limitations, but we will remove this
+    /// functionality in future.
+    public var trustCommonName: Bool
```
PeterAdams-A, line 225: Should we (assuming it's possible) mark the setter for this as deprecated right from the start?

PeterAdams-A (approved):
I think we should do this - anyone that's using functionality which has been deprecated for 20 years should probably invest the time to move on.

Davidde94 (approved):

Lukasa:
Ultimately, fear of breakage. It is probably worth polishing it back up and landing it anyway.

PR #246, opened 2020-10-12
Add `NIOSSLCertificate` debug description for useful debugging

Davidde94:
Printing a `NIOSSLCertificate` now gives the common and alternative names, useful for debugging which certificates are being handled.

### DIFF
--- a/Sources/NIOSSL/SSLCertificate.swift
+++ b/Sources/NIOSSL/SSLCertificate.swift
@@ -450,7 +450,7 @@ extension NIOSSLCertificate: CustomStringConvertible {
                 case .dnsName(let bytes):
                     desc += "\(String(decoding: bytes, as: UTF8.self)),"
                 case .ipAddress(let address):
-                    desc += "\(address.stringValue),"
+                    desc += "\(address.description),"
                 }
             }
         }
@@ -459,47 +459,43 @@ extension NIOSSLCertificate: CustomStringConvertible {
     
 }
 
-extension NIOSSLCertificate.IPAddress {
+extension NIOSSLCertificate.IPAddress: CustomStringConvertible {
     
-    static let ipv4AddressLength = 16
-    static let ipv6AddressLength = 46
+    private static let ipv4AddressLength = 16
+    private static let ipv6AddressLength = 46
     
     /// A string representaion of the IP address.
     /// E.g. IPv4: `192.168.0.1`
     /// E.g. IPv6: `2001:db8::1`
-    public var stringValue: String {
+    public var description: String {
         switch self{
         case .ipv4(let addr):
-            return ipv4ToString(addr)
+            return self.ipv4ToString(addr)
         case .ipv6(let addr):
-            return ipv6ToString(addr)
+            return self.ipv6ToString(addr)
         }
     }
     
     private func ipv4ToString(_ address: in_addr) -> String {
         var address = address
-        let pointer = UnsafeMutablePointer<CChar>.allocate(capacity: Self.ipv4AddressLength)
-        let result = inet_ntop(AF_INET, &address, pointer, socklen_t(Self.ipv4AddressLength))
-        precondition(result != nil, "The IP address was invalid. This should never happen as we're within the IP address struct.")
-        
-        defer {
-            pointer.deallocate()
+        var dest: [CChar] = []
+        dest.reserveCapacity(Self.ipv4AddressLength)
+        dest.withUnsafeMutableBufferPointer { pointer in
+            let result = inet_ntop(AF_INET, &address, pointer.baseAddress!, socklen_t(Self.ipv4AddressLength))
+            precondition(result != nil, "The IP address was invalid. This should never happen as we're within the IP address struct.")
         }
-        
-        return String(cString: pointer)
+        return String(cString: &dest)
     }
     
     private func ipv6ToString(_ address: in6_addr) -> String {
         var address = address
-        let pointer = UnsafeMutablePointer<CChar>.allocate(capacity: Self.ipv6AddressLength)
-        let result = inet_ntop(AF_INET6, &address, pointer, socklen_t(Self.ipv6AddressLength))
-        precondition(result != nil, "The IP address was invalid. This should never happen as we're within the IP address struct.")
-        
-        defer {
-            pointer.deallocate()
+        var dest: [CChar] = []
+        dest.reserveCapacity(Self.ipv6AddressLength)
+        dest.withUnsafeMutableBufferPointer { pointer in
+            let result = inet_ntop(AF_INET6, &address, pointer.baseAddress!, socklen_t(Self.ipv6AddressLength))
+            precondition(result != nil, "The IP address was invalid. This should never happen as we're within the IP address struct.")
         }
-        
-        return String(cString: pointer)
+        return String(cString: &dest)
     }
     
 }
--- a/Sources/NIOSSL/SSLCertificate.swift
+++ b/Sources/NIOSSL/SSLCertificate.swift
@@ -437,8 +437,11 @@ internal class SubjectAltNameSequence: Sequence, IteratorProtocol {
 
 extension NIOSSLCertificate: CustomStringConvertible {
     
+    static let ipv4AddressLength = 16
+    static let ipv6AddressLength = 46
+    
     public var description: String {
-        var desc = "<"
+        var desc = "<NIOSSLCertificate"
         if let commonNameBytes = self.commonName() {
             let commonName = String(decoding: commonNameBytes, as: UTF8.self)
             desc += ";common_name=" + commonName
@@ -468,18 +471,26 @@ extension NIOSSLCertificate: CustomStringConvertible {
     
     private func ipv4ToString(_ address: in_addr) -> String {
         var address = address
-        let pointer = UnsafeMutableBufferPointer<CChar>.allocate(ipv4Len)
-        inet_ntop(AF_INET, &address, pointer, socklen_t(INET_ADDRSTRLEN))
-        let string = String(cString: pointer)
-        return string
+        let pointer = UnsafeMutablePointer<CChar>.allocate(capacity: Self.ipv4AddressLength)
+        inet_ntop(AF_INET, &address, pointer, socklen_t(Self.ipv4AddressLength))
+        
+        defer {
+            pointer.deallocate()
+        }
+        
+        return String(cString: pointer)
     }
     
     private func ipv6ToString(_ address: in6_addr) -> String {
         var address = address
-        let pointer = malloc(Int(INET_ADDRSTRLEN)).assumingMemoryBound(to: CChar.self)
-        inet_ntop(AF_INET6, &address, pointer, socklen_t(INET6_ADDRSTRLEN))
-        let string = String(cString: pointer)
-        return string
+        let pointer = UnsafeMutablePointer<CChar>.allocate(capacity: Self.ipv6AddressLength)
+        inet_ntop(AF_INET6, &address, pointer, socklen_t(Self.ipv6AddressLength))
+        
+        defer {
+            pointer.deallocate()
+        }
+        
+        return String(cString: pointer)
     }
     
 }
--- a/Sources/NIOSSL/SSLCertificate.swift
+++ b/Sources/NIOSSL/SSLCertificate.swift
@@ -479,7 +479,8 @@ extension NIOSSLCertificate.IPAddress {
     private func ipv4ToString(_ address: in_addr) -> String {
         var address = address
         let pointer = UnsafeMutablePointer<CChar>.allocate(capacity: Self.ipv4AddressLength)
-        inet_ntop(AF_INET, &address, pointer, socklen_t(Self.ipv4AddressLength))
+        let result = inet_ntop(AF_INET, &address, pointer, socklen_t(Self.ipv4AddressLength))
+        precondition(result != nil, "The IP address was invalid. This should never happen as we're within the IP address struct.")
         
         defer {
             pointer.deallocate()
@@ -491,7 +492,8 @@ extension NIOSSLCertificate.IPAddress {
     private func ipv6ToString(_ address: in6_addr) -> String {
         var address = address
         let pointer = UnsafeMutablePointer<CChar>.allocate(capacity: Self.ipv6AddressLength)
-        inet_ntop(AF_INET6, &address, pointer, socklen_t(Self.ipv6AddressLength))
+        let result = inet_ntop(AF_INET6, &address, pointer, socklen_t(Self.ipv6AddressLength))
+        precondition(result != nil, "The IP address was invalid. This should never happen as we're within the IP address struct.")
         
         defer {
             pointer.deallocate()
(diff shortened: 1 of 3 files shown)


Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -434,3 +434,52 @@ internal class SubjectAltNameSequence: Sequence, IteratorProtocol {
         CNIOBoringSSL_GENERAL_NAMES_free(self.nameStack)
     }
 }
+
+extension NIOSSLCertificate: CustomStringConvertible {
+    
+    public var description: String {
+        var desc = "<"
```
Lukasa, line 441: Let's include the type here too: `NIOSSLCertificate`.

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -434,3 +434,52 @@ internal class SubjectAltNameSequence: Sequence, IteratorProtocol {
...
+        if let commonNameBytes = self.commonName() {
+            let commonName = String(decoding: commonNameBytes, as: Unicode.UTF8.self)
```
Lukasa, line 443: Nit:

```suggestion
            let commonName = String(decoding: commonNameBytes, as: UTF8.self)
```

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -434,3 +434,52 @@ internal class SubjectAltNameSequence: Sequence, IteratorProtocol {
...
+            desc += ";common_name=" + commonName
+        }
+        if let alternativeName = self.subjectAlternativeNames() {
+            desc += ";alternative_names="
+            for altName in alternativeName {
+                switch altName {
+                case .dnsName(let bytes):
+                    desc += "\(String(decoding: bytes, as: Unicode.UTF8.self)),"
```
Lukasa, line 451: Nit: 
```suggestion
                    desc += "\(String(decoding: bytes, as: UTF8.self)),"
```

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -434,3 +434,52 @@ internal class SubjectAltNameSequence: Sequence, IteratorProtocol {
...
+        case .ipv4(let addr):
+            return ipv4ToString(addr)
+        case .ipv6(let addr):
+            return ipv6ToString(addr)
+        }
+    }
+    
+    private func ipv4ToString(_ address: in_addr) -> String {
+        var address = address
+        let pointer = malloc(Int(INET_ADDRSTRLEN)).assumingMemoryBound(to: CChar.self)
```
Lukasa, line 471: We can't use `assumingMemoryBound` like this. If we were to do this it'd be `bindMemory(to:)` but we may as well allocate the memory pre-bound. Additionally, let's remove `INET_ADDRSTRLEN`: it's not really helping us at all here, and it becomes an awkward symbol we have to remove down the road. Let's use a `let ipv4Len = 16`.

```suggestion
        let pointer = UnsafeMutableBufferPointer<CChar>.allocate(ipv4Len)
```

Lukasa, line 471: Another nit, this memory is leaked. Immediately after this call we need:

```swift
defer {
    pointer.deallocate()
}

File: Sources/NIOSSL/SSLCertificate.swift [hunk 5]
```diff
@@ -434,3 +434,52 @@ internal class SubjectAltNameSequence: Sequence, IteratorProtocol {
...
+        inet_ntop(AF_INET, &address, pointer, socklen_t(INET_ADDRSTRLEN))
```
Lukasa, line 472: We need to check the return value of this call: if it's `nil`, the call failed and our printable string must be the empty string.

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -434,3 +434,52 @@ internal class SubjectAltNameSequence: Sequence, IteratorProtocol {
...
+        let string = String(cString: pointer)
+        return string
```
Lukasa, line 474: We can skip the intermediate `let string`, it doesn't help us, we can just return the `String` directly.

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -434,3 +434,52 @@ internal class SubjectAltNameSequence: Sequence, IteratorProtocol {
...
+    }
+    
+    private func ipv6ToString(_ address: in6_addr) -> String {
+        var address = address
+        let pointer = malloc(Int(INET_ADDRSTRLEN)).assumingMemoryBound(to: CChar.self)
```
Lukasa, line 479: This code has the same problems as the equivalent line above, plus one more: `INET_ADDRSTRLEN` is too small. Let's use `let ipv6Address = 46`.

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -434,3 +434,52 @@ internal class SubjectAltNameSequence: Sequence, IteratorProtocol {
...
+        inet_ntop(AF_INET6, &address, pointer, socklen_t(INET6_ADDRSTRLEN))
```
Lukasa, line 480: Same note here about return value.

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -434,3 +434,52 @@ internal class SubjectAltNameSequence: Sequence, IteratorProtocol {
...
+        let string = String(cString: pointer)
```
Lukasa, line 481: Same note here about returning directly.

Lukasa (changes requested):
Thanks! I've left some notes in the diff.

Sources/NIOSSL/SSLCertificate.swift
Davidde94: Oops, a copy-paste error, this was meant to be `INET6_ADDRSTRLEN`... I'll make your suggested change :)

Sources/NIOSSL/SSLCertificate.swift
Davidde94: Oops, yep, this was just a debugging thing for me

Sources/NIOSSL/SSLCertificate.swift
Davidde94: I thought I might as well leave that out, as surely we will always have a valid address at this point?

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -434,3 +434,72 @@ internal class SubjectAltNameSequence: Sequence, IteratorProtocol {
...
+                    desc += "\(address.stringValue),"
+                }
+            }
+        }
+        return desc + ">"
+    }
+    
+}
+
+extension NIOSSLCertificate.IPAddress {
```
Lukasa, line 462: Should this just be a conformance to `CustomStringConvertible`?

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -434,3 +434,72 @@ internal class SubjectAltNameSequence: Sequence, IteratorProtocol {
...
+    
+    static let ipv4AddressLength = 16
```
Lukasa, line 464: Let's up these to `private` as well.

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -434,3 +434,72 @@ internal class SubjectAltNameSequence: Sequence, IteratorProtocol {
...
+    
+    /// A string representaion of the IP address.
+    /// E.g. IPv4: `192.168.0.1`
+    /// E.g. IPv6: `2001:db8::1`
+    public var stringValue: String {
+        switch self{
+        case .ipv4(let addr):
+            return ipv4ToString(addr)
+        case .ipv6(let addr):
+            return ipv6ToString(addr)
```
Lukasa, line 475: Nit: explicit `self`.

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -434,3 +434,72 @@ internal class SubjectAltNameSequence: Sequence, IteratorProtocol {
...
+    
+    private func ipv4ToString(_ address: in_addr) -> String {
+        var address = address
+        let pointer = UnsafeMutablePointer<CChar>.allocate(capacity: Self.ipv4AddressLength)
+        let result = inet_ntop(AF_INET, &address, pointer, socklen_t(Self.ipv4AddressLength))
+        precondition(result != nil, "The IP address was invalid. This should never happen as we're within the IP address struct.")
+        
+        defer {
+            pointer.deallocate()
+        }
```
Lukasa, line 487: Let's move this `defer` literally right below the allocation. It helps us avoid the risk of adding leaks in future.

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -434,3 +434,72 @@ internal class SubjectAltNameSequence: Sequence, IteratorProtocol {
...
+        case .ipv4(let addr):
+            return ipv4ToString(addr)
+        case .ipv6(let addr):
+            return ipv6ToString(addr)
+        }
+    }
+    
+    private func ipv4ToString(_ address: in_addr) -> String {
+        var address = address
+        let pointer = UnsafeMutablePointer<CChar>.allocate(capacity: Self.ipv4AddressLength)
```
glbrntt, line 481: @Lukasa would it ever be preferred to create an array of `CChar` with the right capacity and use `withUnsafeMutableBufferPointer` so you avoid having to explicitly deallocate. I suppose one downside is that you could end up allocating more memory than you actually need.

Lukasa: In fact, doing that plus using an `&` is probably the best outcome, yeah.

Sources/NIOSSL/SSLCertificate.swift
Davidde94: Yeh, good spot, probably

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -434,3 +434,68 @@ internal class SubjectAltNameSequence: Sequence, IteratorProtocol {
...
+            return self.ipv4ToString(addr)
+        case .ipv6(let addr):
+            return self.ipv6ToString(addr)
+        }
+    }
+    
+    private func ipv4ToString(_ address: in_addr) -> String {
+        var address = address
+        var dest: [CChar] = []
+        dest.reserveCapacity(NIOSSLCertificate.IPAddress.ipv4AddressLength)
```
Lukasa, line 482: This won't work how you want: `withUnsafeMutableBufferPointer` won't include the trailing capacity. We need to zero-fill `dest`: `var dest: [CChar] = Array(repeating: 0, count: ipv4AddressLength)`.

Tests/NIOSSLTests/SSLCertificateTest.swift
```diff
@@ -380,4 +380,20 @@ class SSLCertificateTest: XCTestCase {
 
         XCTAssertEqual(certBytes, expectedCertBytes)
     }
+    
+    func testPrintingDebugDetails_commonName() throws {
```
glbrntt, line 384: super nit: none of the other tests use `_` in their names, let's be consistent

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -434,3 +434,66 @@ internal class SubjectAltNameSequence: Sequence, IteratorProtocol {
...
+    }
+    
+}
+
+extension NIOSSLCertificate.IPAddress: CustomStringConvertible {
+    
+    private static let ipv4AddressLength = 16
+    private static let ipv6AddressLength = 46
+    
+    /// A string representaion of the IP address.
```
Lukasa, line 467:
```suggestion
    /// A string representation of the IP address.
```

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -434,3 +434,66 @@ internal class SubjectAltNameSequence: Sequence, IteratorProtocol {
...
+    /// E.g. IPv4: `192.168.0.1`
+    /// E.g. IPv6: `2001:db8::1`
+    public var description: String {
+        switch self{
```
Lukasa, line 471:
```suggestion
        switch self {
```

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -434,3 +434,66 @@ internal class SubjectAltNameSequence: Sequence, IteratorProtocol {
...
+            let commonName = String(decoding: commonNameBytes, as: UTF8.self)
+            desc += ";common_name=" + commonName
+        }
+        if let alternativeName = self.subjectAlternativeNames() {
+            let altNames = alternativeName.map { name in
+                switch name {
+                case .dnsName(let bytes):
+                    return String(decoding: bytes, as: UTF8.self)
+                case .ipAddress(let address):
+                    return address.description
```
Lukasa, line 452:
```suggestion
                    return String(describing: address)
```

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -434,3 +434,66 @@ internal class SubjectAltNameSequence: Sequence, IteratorProtocol {
...
+        case .ipv6(let addr):
+            return self.ipv6ToString(addr)
+        }
+    }
+    
+    private func ipv4ToString(_ address: in_addr) -> String {
+        var address = address
+        var dest: [CChar] = Array(repeating: 0, count: NIOSSLCertificate.IPAddress.ipv4AddressLength)
+        dest.withUnsafeMutableBufferPointer { pointer in
+            let result = inet_ntop(AF_INET, &address, pointer.baseAddress!, socklen_t(NIOSSLCertificate.IPAddress.ipv4AddressLength))
```
Lukasa, line 483:
```suggestion
            let result = inet_ntop(AF_INET, &address, pointer.baseAddress!, socklen_t(pointer.count))
```

glbrntt (approved):
LGTM

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -434,3 +434,66 @@ internal class SubjectAltNameSequence: Sequence, IteratorProtocol {
...
+            precondition(result != nil, "The IP address was invalid. This should never happen as we're within the IP address struct.")
+        }
+        return String(cString: &dest)
+    }
+    
+    private func ipv6ToString(_ address: in6_addr) -> String {
+        var address = address
+        var dest: [CChar] = Array(repeating: 0, count: NIOSSLCertificate.IPAddress.ipv4AddressLength)
```
Lukasa, line 491:
```suggestion
        var dest: [CChar] = Array(repeating: 0, count: NIOSSLCertificate.IPAddress.ipv6AddressLength)
```

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -434,3 +434,66 @@ internal class SubjectAltNameSequence: Sequence, IteratorProtocol {
...
+        dest.withUnsafeMutableBufferPointer { pointer in
+            let result = inet_ntop(AF_INET6, &address, pointer.baseAddress!, socklen_t(NIOSSLCertificate.IPAddress.ipv6AddressLength))
```
Lukasa, line 493:
```suggestion
            let result = inet_ntop(AF_INET6, &address, pointer.baseAddress!, socklen_t(pointer.count))
```

PR #247, opened 2020-10-13
Add NIOSSLCertificate serial number var, and add to description

Davidde94:
It's now possible to get the serial number for any given NIOSSLCertificate using the new `serialNumber` variable. This has also been added to the beginning of the certificate description.

Implemented as a lazy var, also possible to compute in the initialiser, take your pick.

### DIFF
--- a/Sources/NIOSSL/SSLCertificate.swift
+++ b/Sources/NIOSSL/SSLCertificate.swift
@@ -47,15 +47,10 @@ public class NIOSSLCertificate {
         case ipv6(in6_addr)
     }
     
-    public lazy var serialNumber: String = {
-        let serialNumber = CNIOBoringSSL_X509_get_serialNumber(self.ref)
-        var bn = BIGNUM()
-        CNIOBoringSSL_ASN1_INTEGER_to_BN(serialNumber, &bn)
-        guard let result = CNIOBoringSSL_BN_bn2hex(&bn) else {
-            preconditionFailure("Failed to convert bignum to hex")
-        }
-        return String(cString: result)
-    }()
+    public var serialNumber: [UInt8] {
+        let serialNumber = CNIOBoringSSL_X509_get_serialNumber(self.ref)!
+        return Array(UnsafeBufferPointer(start: serialNumber.pointee.data, count: Int(serialNumber.pointee.length)))
+    }
 
     private init(withOwnedReference ref: UnsafeMutablePointer<X509>) {
         self._ref = UnsafeMutableRawPointer(ref) // erasing the type for @_implementationOnly import CNIOBoringSSL
@@ -448,7 +443,8 @@ internal class SubjectAltNameSequence: Sequence, IteratorProtocol {
 extension NIOSSLCertificate: CustomStringConvertible {
     
     public var description: String {
-        var desc = "<NIOSSLCertificate;serial_number=" + self.serialNumber
+        let serialNumber = self.serialNumber.map { String($0, radix: 16) }.reduce("", +)
+        var desc = "<NIOSSLCertificate;serial_number=\(serialNumber)"
         if let commonNameBytes = self.commonName() {
             let commonName = String(decoding: commonNameBytes, as: UTF8.self)
             desc += ";common_name=" + commonName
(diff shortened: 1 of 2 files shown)


Tests/NIOSSLTests/SSLCertificateTest.swift
```diff
@@ -380,4 +380,20 @@ class SSLCertificateTest: XCTestCase {
...
+    func testPrintingDebugDetailsNoAlternativeNames() throws {
+        let expectedDebugDescription = "<NIOSSLCertificate;serial_number=9fd7d05a34ca7984;common_name=robots.sanfransokyo.edu>"
+        let cert = try assertNoThrowWithValue(NIOSSLCertificate(bytes: .init(samplePemCert.utf8), format: .pem))
+        let debugString = String(describing: cert)
+
+        XCTAssertEqual(debugString, expectedDebugDescription)
+    }
+    
+    func testPrintingDebugDetailsWithAlternativeNames() throws {
+        let expectedDebugDescription = "<NIOSSLCertificateserial_number=46231a526848d57af4999e29f89988d178d94da2;common_name=localhost;alternative_names=localhost,example.com,192.168.0.1,2001:db8::1>"
```
glbrntt, line 393: `NIOSSLCertificateserial_number` 🤔

glbrntt (changes requested):

Tests/NIOSSLTests/SSLCertificateTest.swift
Davidde94: Oh oops, I hadn't pushed a commit that fixed this haha

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -46,6 +46,16 @@ public class NIOSSLCertificate {
         case ipv4(in_addr)
         case ipv6(in6_addr)
     }
+    
+    public lazy var serialNumber: String = {
```
Lukasa, line 50: `public lazy` is a nasty combination because it's not thread safe. In this case the computation is cheap, so we should just do this as a computed property.

Lukasa, line 50: Relatedly, I don't know if this property should be a `String`. A serial number is technically a number, but Swift doesn't have a currency bignum type and we don't expose one in NIOSSL. For that reason I think we should probably expose this as an array of UInt8? It's an awkward representation, but it does allow people to easily transform it into an appropriate type, or do what they need with the bytes. Our need to print it is a separate concern and we can handle that internally.

glbrntt (approved):
This looks fine to me but I'll defer to @Lukasa as I'm not an expert here.

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -46,6 +46,16 @@ public class NIOSSLCertificate {
...
+        let serialNumber = CNIOBoringSSL_X509_get_serialNumber(self.ref)
+        var bn = BIGNUM()
```
Lukasa, line 52: This BIGNUM based solution is a bit roundabout, and has cleanup concerns which we aren't addressing. I think we can do a bit better.

BoringSSL stores ASN.1 integers internally as "a big-endian representation of the data, ignoring the sign". The sign information is stored in the type detail. We also know, per X.509's requirements, that the serial number is not more than 20 bytes long, and must be positive.

With those constraints we can say that we'll have an easier time doing the hex-encoding ourselves and avoiding the dance into BoringSSL land. I think we want this:

```swift
        let serial = CNIOBoringSSL_X509_get_serialNumber(self.ref)!
        return Array(UnsafeBufferPointer(start: serial.pointee.data, count: Int(serial.pointee.length)))
```

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -438,7 +448,7 @@ internal class SubjectAltNameSequence: Sequence, IteratorProtocol {
 extension NIOSSLCertificate: CustomStringConvertible {
     
     public var description: String {
-        var desc = "<NIOSSLCertificate"
+        var desc = "<NIOSSLCertificate;serial_number=" + self.serialNumber
```
Lukasa, line 451: May as well string interpolate here? Just a bit nicer I think.

Lukasa (commented):
Generally looks good, great job navigating the BoringSSL APIs! I've left a few notes in the diff.

tomerd:

Lukasa (approved):
Nice, love it.

PR #250, opened 2020-10-14
Fix threading violation in KeyLogCallbackManager

Lukasa:
Motivation:

The KeyLogCallbackManager belongs on an SSLContext, and so may be
invoked from multiple threads. However, it incorrectly cached a
ByteBuffer on its storage, leading to thread unsafety. This tends to
manifest in assertion failures, but could lead to arbitrary data
corruption or double frees.

Modifications:

- Remove the cached bytebuffer.
- Add probabilistic test.

Result:

No more thread-safety violations.

### DIFF
--- a/Tests/NIOSSLTests/TLSConfigurationTest.swift
+++ b/Tests/NIOSSLTests/TLSConfigurationTest.swift
@@ -401,4 +401,44 @@ class TLSConfigurationTest: XCTestCase {
         XCTAssertEqual(config.applicationProtocols, ["h2", "http/1.1"])
         XCTAssertEqual(config.encodedApplicationProtocols, [[2, 104, 50], [8, 104, 116, 116, 112, 47, 49, 46, 49]])
     }
+
+    func testKeyLogManagerOverlappingAccess() throws {
+        // Tests that we can have overlapping calls to the log() function of the keylog manager.
+        // This test fails probabilistically! DO NOT IGNORE INTERMITTENT FAILURES OF THIS TEST.
+        let semaphore = DispatchSemaphore(value: 0)
+        let group = DispatchGroup()
+        let completionsQueue = DispatchQueue(label: "completionsQueue")
+        var completions: [Bool] = []
+
+        func keylogCallback(_ ign: ByteBuffer) {
+            completionsQueue.sync {
+                completions.append(true)
+                semaphore.wait()
+            }
+            group.leave()
+        }
+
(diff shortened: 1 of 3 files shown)


Tests/NIOSSLTests/TLSConfigurationTest.swift
```diff
@@ -401,4 +401,44 @@ class TLSConfigurationTest: XCTestCase {
...
+
+    func testKeyLogManagerOverlappingAccess() throws {
+        // Tests that we can have overlapping calls to the log() function of the keylog manager.
+        // This test fails probabilistically! DO NOT IGNORE INTERMITTENT FAILURES OF THIS TEST.
+        let semaphore = DispatchSemaphore(value: 0)
+        let group = DispatchGroup()
+        let completionsQueue = DispatchQueue(label: "completionsQueue")
+        var completions: [Bool] = []
+
+        func keylogCallback(_ ign: ByteBuffer) {
```
glbrntt, line 413:
```suggestion
        func keylogCallback(_ ignored: ByteBuffer) {
```

Davidde94 (approved):
LGTM

Tests/NIOSSLTests/TLSConfigurationTest.swift
```diff
@@ -401,4 +401,44 @@ class TLSConfigurationTest: XCTestCase {
         XCTAssertEqual(config.applicationProtocols, ["h2", "http/1.1"])
         XCTAssertEqual(config.encodedApplicationProtocols, [[2, 104, 50], [8, 104, 116, 116, 112, 47, 49, 46, 49]])
     }
+
+    func testKeyLogManagerOverlappingAccess() throws {
+        // Tests that we can have overlapping calls to the log() function of the keylog manager.
+        // This test fails probabilistically! DO NOT IGNORE INTERMITTENT FAILURES OF THIS TEST.
```
glbrntt, line 407: Did this fail frequently enough to warrant running it in a loop a handful of times? Bit of a bummer we need the 10ms wait though.

Lukasa: I don't think so, tbh. The test itself is constructed such that the act of fixing it made the failing version of the test impossible to write (`log` became non-`mutating`), so it's a bit odd anyway.

glbrntt: Oh of course, I didn't register the `mutating` being dropped.

PR #255, opened 2020-11-19
resolves #254

### LINKED ISSUE #254: Compilation fails for Swift version 5.3-dev (LLVM 69d8678431d3eee, Swift e1aef96b7fea59b)
I am using Tensorflow for Swift toolchain for my project. After upgrading to release V12 from https://github.com/tensorflow/swift/blob/master/Installation.md I started getting following error

```
139:7: error: missing '#include <stdlib.h>'; 'abort' must be declared before it is used
      abort();
      ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.0.sdk/usr/include/stdlib.h:131:7: note: declaration here is not visible
void     abort(void) __cold __dead2;
```

This can be solved by including stdlib.h to CNIOBoringSSL_span.h

SercanKaraoglu:
This resolves the following error that I get when I compile the project with Swift 5.3-dev

```
139:7: error: missing '#include <stdlib.h>'; 'abort' must be declared before it is used
      abort();
```

### DIFF
--- a/Sources/CNIOBoringSSL/include/CNIOBoringSSL_span.h
+++ b/Sources/CNIOBoringSSL/include/CNIOBoringSSL_span.h
@@ -24,6 +24,7 @@ extern "C++" {
 #include <algorithm>
 #include <cstdlib>
 #include <type_traits>
+#include <stdlib.h>
 
 BSSL_NAMESPACE_BEGIN
 


Lukasa (changes requested):
Thanks for filing this! However, this is a clang bug. Two lines above the include you added is an include of `<cstdlib>`. This is the exact same header, and should appropriately supply `abort`. It’s clang’s mistake to think it doesn’t. Please file this bug upstream with Tensorflow.

PR #266, opened 2021-01-25
Stop guarding implementationOnly

Lukasa:
Motivation:

When we added support for implementationOnly imports, we guarded them
behind a maximum compiler version to prevent the possibility of the
spelling breaking in a future compiler release. This was pessimistic and
largely unnecessary. If the spelling does change in a future compiler
release, we can always ship a patch update that adds the compiler guard
to fix older versions. The solution, as it stands, will require trivial
patches every six months to maintain, which is much messier than just
removing the upper bound.

Modifications:

- Remove the upper bound on the compiler version for
  `@_implementationOnly` import.

Result:

Fewer trivial patches.

### DIFF
--- a/Sources/NIOSSL/ByteBufferBIO.swift
+++ b/Sources/NIOSSL/ByteBufferBIO.swift
@@ -13,7 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 import NIO
-#if compiler(>=5.1) && compiler(<5.4)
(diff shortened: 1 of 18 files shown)


Lukasa:
Hmm, 5.2 tests failed for a reason unrelated to this change.

Davidde94 (approved):

Lukasa:

PR #269, opened 2021-02-11
Update vendor script to namespace C++ structures.

Lukasa:
Motivation:

C++ structures emit a number of symbols into their binaries that are
named based on the name of the structure. Most notably, these are the
three (or five, depending on mode) magic structure methods: copy
constructor, copy assignment operator, destructor.

By and large this isn't an issue for us as BoringSSL already namespaces
most of its structures. However, a few structures are the target of
opaque typedefs in the public header files. These need to use the names
that would appear in the internal source. If they didn't, the internal
source wouldn't compile! As a result they need to appear outside the
inner prefixing namespace.

We knew about this limitation at the time we did this work, but hoped
we'd get along just fine without changing anything. In fact we did, for
nearly three years. But the chickens have come home to roost in
https://github.com/grpc/grpc-swift/issues/1134.

So it's time to fix it.

Fortunately for us, we never write down the names of the inner types in
our source code. That means we don't actually care what that name _is_.
This gives us our escape hatch: we can namespace the names of the
underlying C++ types using the same preprocessor trick as the function
names. This will have the effect of also namespacing their magic
methods.

Modifications:

This patch attempts to make this change defensively, by assuming that
other structures may be added that meet this pattern. To do it, we
compile the binary _again_ and search for any C++ methods that haven't
already been namespaced and that don't belong to Swift. We then extract
the structure name from them and add a `#define` for it. This forcibly
renames the structure in all cases.

Result:

No more colliding symbols.

Resolves grpc/grpc-swift#1134


scripts/vendor-boringssl.sh
```diff
@@ -132,6 +132,41 @@ function mangle_symbols {
...
+    echo "MANGLING C++ structures"
+    (
+        # We need a .a: may as well get SwiftPM to give it to us.
+        # Temporarily enable the product we need.
+        $sed -i -e 's/MANGLE_START/MANGLE_START*\//' -e 's/MANGLE_END/\/*MANGLE_END/' "${HERE}/Package.swift"
+
+        # Build for macOS.
+        swift build --product CNIOBoringSSL
+
+        # Woah, this is a hell of a command! What does it do?
```
glbrntt, line 152: This whole comment is really helpful!

scripts/vendor-boringssl.sh
```diff
@@ -132,6 +132,41 @@ function mangle_symbols {
     namespace_inlines "$DSTROOT"
 }
 
+
+# BoringSSL includes a few non-namespaced C++ structures. These aren't namespaced because they're exposed
+# in C-land, which doesn't know about the namespacing. Sadly, these structures include constructors and destructors,
+# and if those aren't namespaced we're still able to conflict.
+#
+# This function is responsible for identifying them and manually cleaning them up. We run this only on
+# macOS because we don't believe that the cross-platform architectures will hit any other structures.
```
glbrntt, line 141: probably silly question, but why?

glbrntt (approved):
I _think_ this looks good

scripts/vendor-boringssl.sh
Lukasa: The only way this would happen is if there were structures present that met the following criteria:

1. Names exported into C interface files (i.e. public headers), so they can't be namespaced
2. But only on some platforms

This is fairly unlikely: BoringSSL doesn't have that much platform specific code, and even less platform-specific API.

weissi (approved):
beautiful :P

scripts/vendor-boringssl.sh
```diff
@@ -132,6 +132,41 @@ function mangle_symbols {
...
+        swift build --product CNIOBoringSSL
+
+        # Woah, this is a hell of a command! What does it do?
+        #
+        # The nm command grabs all global defined symbols. We then run the C++ demangler over them and look for methods with '::' in them:
+        # these are C++ methods. We then exclude any that contain CNIOBoringSSL (as those are already namespaced!) and any that contain swift
+        # (as those were put there by the Swift runtime, not us). This gives us a list of symbols. The following cut command
+        # grabs the type name from each of those (the bit preceding the '::'). Finally, we sort and uniqify that list. This gives us all the
+        # structures that need to be renamed.
+        structures=$(nm -gUj "$HERE/.build/x86_64-apple-macosx/debug/libCNIOBoringSSL.a" | c++filt | grep "::" | grep -v -e "CNIOBoringSSL" -e "swift" | cut -d : -f1 | sort | uniq)
```
t089, line 159: Out of curiosity, will this work on a M1 Mac? :)

PR #274, opened 2021-02-18
Allow warnings when building Podspecs

glbrntt:
Motivation:

When pushing Cocoapods a warning is emitted for SwiftNIOSSL
('SecTrustEvaluateAsync' was deprecated in watchOS 6.0). Pushing the
pods fails as a result.

Modifications:

Allow warnings when pushing pods.

Result:

Pods can be pushed with warnings.


Lukasa (approved):

Lukasa:
We should update to use the non-deprecated API where it's available as well.

PR #277, opened 2021-02-22
Update podspec script to remove 'S' files from watchOS target

### LINKED ISSUE #276: Remove 'S' files from watchOS target in CocoaPods
SwiftNIOSSL currently fails app store validation when rebuilding from Bitcode due to the presence of the assembly files. While it isn't currently possible to remove these files from SwiftPM-built targets, we can remove them from CocoaPods targets. As such, I believe the podspec should be updated to remove these files from watchOS and the latest version redeployed to the CocoaPods spec repo. This will enable a workaround for watchOS projects until SwiftPM adds the necessary support.

More context in #251

daltonclaybrook:
This PR resolves #276

### DIFF
--- a/scripts/build_podspec.sh
+++ b/scripts/build_podspec.sh
@@ -121,7 +121,10 @@ Pod::Spec.new do |s|
   s.tvos.deployment_target = '10.0'
   s.watchos.deployment_target = '6.0'
 
-  s.source_files = 'Sources/${target#Swift}/**/*.{swift,c,h,cc,S}'
+  s.source_files = 'Sources/${target#Swift}/**/*.{swift,c,h,cc}'
+  s.ios.source_files = 'Sources/${target#Swift}/**/*.S'
+  s.osx.source_files = 'Sources/${target#Swift}/**/*.S'
+  s.tvos.source_files = 'Sources/${target#Swift}/**/*.S'
   $public_header_files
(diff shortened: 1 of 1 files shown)


Lukasa:
@daltonclaybrook Should this patch also provide the #NO_ASM flag for watchOS, so we can get you all the way to “works fine” for your use-case?

daltonclaybrook:
@Lukasa Good thinking, just added that to the podspec. Let me know if this looks correct to you. In my testing, when I use `$(inherited)` with `GCC_PREPROCESSOR_DEFINITIONS`, the Pods project ends up with two instances of that variable. Not sure why this is happening, but I omitted it.
GCC_PREPROCESSOR_DEFINITIONS

Lukasa (approved):
This LGTM, I think. @glbrntt want to take a look as well?

Lukasa:

glbrntt (approved):
Looks good to me.

daltonclaybrook:
Can the podspec be republished for 2.10.3, or will it have to wait for another version?

Lukasa:
I don't believe we can publish a pod spec for the same version again. However, I'm open to cutting a new release explicitly to provide the new pod spec. @glbrntt, how do you feel here?

glbrntt:
New release sounds fine to me.

PR #280, opened 2021-03-13
Implement bestEffort functions for hashable and equatable.

madsodgaard:
This PR adds `bestEffortHash` and `bestEffortEquals` in order to provide best effort method for using `Hashable` and `Equatable` in other projects. This is useful/needed to solve a problem such as https://github.com/swift-server/async-http-client/issues/330


Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -393,3 +393,27 @@ public struct TLSConfiguration {
                                 renegotiationSupport: renegotiationSupport)
     }
 }
+
+// MARK: Hashable
+extension TLSConfiguration: Hashable {
+    public static func == (lhs: TLSConfiguration, rhs: TLSConfiguration) -> Bool {
+        lhs.hashValue == rhs.hashValue
```
weissi, line 400: That one’s not quite correct. Hash values of unequal things may be the same. You’ll have to compare all the individual fields.

weissi (changes requested):
Thank you. That’s a good start! We’ll need to fix the equals method and need some tests too.

Sources/NIOSSL/TLSConfiguration.swift
madsodgaard: Ah okay, that makes sense yea!

glbrntt:

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -393,3 +393,38 @@ public struct TLSConfiguration {
                                 renegotiationSupport: renegotiationSupport)
     }
 }
+
+// MARK: Hashable
+extension TLSConfiguration: Hashable {
+    public static func == (lhs: TLSConfiguration, rhs: TLSConfiguration) -> Bool {
+        lhs.minimumTLSVersion == rhs.minimumTLSVersion &&
```
glbrntt, line 400: You'll need an explicit `return` here since we still support Swift 5.0

Lukasa (commented):
Hmm, I'm uncertain about whether it's a good idea to define equatability and hashability for this object in a way that does not include the key log callback. This seems likely to cause surprising behaviours in cases where the type is used as a key.

I'm inclined to suggest we should solve the problem in async-http-client a different way, rather than using this incomplete and surprising definition for equatability and hashability.

madsodgaard:
What are the alternative ways of solving this problem? If `TLSConfiguration` is not `Hashable` we can't use it in our key for the connection pool, if I understand correctly? I'd be happy to try another way, just not sure what that way could be, other than not utilizing the pool.

Lukasa:
Use something that isn't `TLSConfiguration` but that either can be derived from it or that can derive it.

In this case it seems like if we wanted to do the least effort possible we could just put these definitions on a wrapper structure whose equality _is_ defined this way. However, a better move is probably to have AHC have its own notion of TLS configuration not directly tied to swift-nio-ssl's, not least given that it works with multiple TLS backends anyway.

weissi:
@Lukasa / @madsodgaard I agree that we should only hash in here if we can include all callbacks, we definitely can't just compare everything but the callbacks. If the callbacks are all C callbacks, then I'm happy comparing their pointer address. Cory wdyt?

@madsodgaard what Cory suggests is what I meant with if we can't reasonably hash it in here, then we can hack around it in AHC. The simplest way (in AHC) is `struct TLSConfigurationHashableHack: Hashable { var tlsConfigureation: TLSConfiguration }`. Then you can define hashable on `TLSConfigurationHashableHack`.

madsodgaard:
@weissi Oh okay. And then it would be fine that we don't include the callbacks in AHC's definition of what `Hashable` is for `TLSConfiguration(Wrapper)`?

weissi:
@madsodgaard tbh, I don't think we can just not hash the callbacks at all. There can conceivably two distinct `TLSConfiguration` that only differ in the callbacks. And the callbacks are very meaningful because that's where the user tells us whether to trust a cert or not. It would be _really bad_ if the user configured an "always accept" callback that would then get reused for something else.

Lukasa:
The callbacks are not C callbacks, they're Swift callbacks.

weissi:
> The callbacks are not C callbacks, they're Swift callbacks.

@Lukasa / @madsodgaard Okay, then we have basically one option: hash/compare the bytes of the closure
```swift
withUnsafeBytes(of: &self.theClosure) { closureBits in
    var hasher = Hasher()
    closureBits.forEach { hasher.combine($0) }
}
```

@Lukasa wdyt?

Lukasa:
I remain uncertain of why we think we have to do this, for two reasons.

Firstly, AHC has already made the mistake of trying to support two TLS implementations using the configuration for only one of them. This is a bad decision, and we shouldn't propagate it further. AHC should have a _general_ TLS configuration object that it can use to derive the specific ones, and that object absolutely can be hashable.

Secondly, even if we disregarded that, there is only one callback on `TLSConfiguration` and it's `keyLogCallback`, used to implement `SSLKEYLOGFILE` support. Do we really need to drop into unsafe code to add hashability to this type for that?

weissi:
> I remain uncertain of why we think we have to do this, for two reasons.

The upsides of doing it in NIOSSL are:
- NIOSSL knows exactly what's part of TLSConfiguration (in the latest version), it's harder to forget to add a newly added member to the equals/hashable implementations
- NIOSSL has to deal with unsafe code already so arguably it's the best place to do it.

One other option of doing it inside NIOSSL is to instead of implementing `Hashable`, exporting two functions:

```swift
extension TLSConfiguration {
    // will say no to two unequals, tries to say yes to two equals but can't guarantee no false negatives
    public func bestEffortIsEqualTo(_ other: TLSConfiguration) -> Bool {
       ...
    }

    // two objects that are equal wrt `bestEffortIsEqualTo` are guaranteed to have the same hash code, others do not.
    public func bestEffortHash() -> UInt64 {
        ...
    }
}
```

WDYT?


> Firstly, AHC has already made the mistake of trying to support two TLS implementations using the configuration for only one of them. This is a bad decision, and we shouldn't propagate it further. AHC should have a _general_ TLS configuration object that it can use to derive the specific ones, and that object absolutely can be hashable.
> 
> Secondly, even if we disregarded that, there is only one callback on `TLSConfiguration` and it's `keyLogCallback`, used to implement `SSLKEYLOGFILE` support. Do we really need to drop into unsafe code to add hashability to this type for that?

Fair enough, @madsodgaard in that case I'd suggest to move this code into an `internal` type into AHC which does

```swift
internal struct HashableTLSConfiguration {
   var tlsConfiguration: TLSConfiguration
}

extension HashableTLSConfiguration: Hashable { 
   ...
}
```

and then inside AHC, we'd just put a `HashableTLSConfiguration` into the connection pool instead of a `TLSConfiguration`.

If @Lukasa thinks the `bestEffortIsEqualTo/HashCode` here is a good idea, then we still need those types in AHC but they'd just need to call into the best effort functions.

weissi:
friendly ping @Lukasa on the `bestEffort*` in this repo

Lukasa:
I'm not sure that we gain much from the `bestEffort` implementations tbqh.

weissi:
> I'm not sure that we gain much from the `bestEffort` implementations tbqh.

But how are the dependencies supposed to hash/compare the right things? We'll add things to the TLS config in the future and then things will start to subtle break. So I think NIOSSL can definitely provide the best fidelity here (and IMHO should). But I agree with you that we shouldn't do it with `Hashable` because it's only a best effort.

Lukasa:
The question remains why dependencies are hashing the TLS configuration at all.

weissi:
> The question remains why dependencies are hashing the TLS configuration at all.

For connection pools, we need to be able to compare the TLS connection of the connection in the pool with the TLS connection that the user would like for the request. False negatives are fine, we can always create a new connection.

That means we need (best effort) comparison (so the user can create an internal type wrapping the TLS config that adds `Equatable`). And because it's kinda nice to store the pool in a dictionary so they'd need to make it hashable.

Lukasa:
Yes, I understand the proposed design. My question is whether that proposed design is _good_.

Specifically, should libraries encapsulating swift-nio-ssl pervasively expose `TLSConfiguration` as the mechanism by which they configure TLS. My general answer is no, they should not, because it couples them tightly to their TLS implementation. If users instead provide an encapsulating general-purpose TLS configuration that they use to generate a swift-nio-ssl `TLSConfiguration` object, only the general-purpose one needs to be `Hashable`. That seems to me to be a vastly preferable design, no?

weissi:
> Specifically, should libraries encapsulating swift-nio-ssl pervasively expose `TLSConfiguration` as the mechanism by which they configure TLS. My general answer is no, they should not, because it couples them tightly to their TLS implementation.

I see! I agree with you here.

> That seems to me to be a vastly preferable design, no?

Absolutely. I do however think that evangelism and not preventing the best effort eq/hash APIs. Fact is that there are libraries that use `NIOSSL.TLSConfiguration` today and I think we should allow them (through the best effort eq/hash APIs) to solve their problem today. I'm happy with adding a `- warning: If you need this, then you are probably doing ... but shouldn't`.

This is compounded by the fact that creating a TLS configuration object that can then yield a `NIOSSL.TLSConfiguration`/`Security.TLSOptions`/Windows equivalent isn't actually super easy and there's (AFAIK) no prior art in Swift. So not offering the proposed APIs leaves the authors of those libraries in a tough spot.

Short of now offering connection pools, they only really replicate the eq/hash functionality in their own packages which I think may be at a risk of security vulnerabilities because we may add properties that they miss. That could make their hand-rolled eq/hash functionality have _false positives_ (claiming two TLSConfigs are equal when they're not) and that's bad.

Lukasa:
I think if we think we should offer this, then we should go whole hog and use `withUnsafeBytes(of:)` the closure.

weissi:
> I think if we think we should offer this, then we should go whole hog and use `withUnsafeBytes(of:)` the closure.

Works for me. I'm mildly worried about ASan complaining that we read bytes that were never written (padding) but I'm totally happy deferring that to when/if we actually hit this issue.

madsodgaard:
@Lukasa @weissi I implemented the bestEffort* functions from what I understood from your conclusion. I added some documentation/warnings, but let me know what you think is appropriate to put there!

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -393,3 +393,58 @@ public struct TLSConfiguration {
...
+
+// MARK: BestEffortHashable
+extension TLSConfiguration {
+    /// Returns a best effort result of whether two `TLSConfiguration` objects are equal.
+    ///
+    /// The "best effort" stems from the fact that we are checking the pointer to the `keyLogCallback` closure.
+    ///
+    /// - warning: You should probably not use this function. This function can return false-negatives, but not false-positives.
+    public func bestEffortEquals(_ comparing: TLSConfiguration) -> Bool {
+        let isKeyLoggerCallbacksEqual = withUnsafeBytes(of: self.keyLogCallback) { callbackPointer1 in
```
weissi, line 405: given [SR-14520](https://bugs.swift.org/browse/SR-14520) we should probably make this

```swift
let isBitEqual = withUnsafeBytes(of: self) { selfBits in
    withUnsafeBytes(of: comparing) { otherBits in
        return selfBits.elementsEqual(otherBits)
    }
}
```

with a comment liking the bug.

weissi: @Lukasa wdyt?

Tests/NIOSSLTests/TLSConfigurationTest.swift
```diff
@@ -441,4 +441,21 @@ class TLSConfigurationTest: XCTestCase {
...
+        theSameConfig.bestEffortHash(into: &hasher2)
+        XCTAssertEqual(hasher.finalize(), hasher2.finalize())
+        XCTAssertTrue(config.bestEffortEquals(theSameConfig))
+    }
+    
+    func testDifferentHashValues() {
+        let config = TLSConfiguration.forServer(certificateChain: [], privateKey: .file("fake.file"), applicationProtocols: ["http/1.1"])
+        let differentConfig = TLSConfiguration.forServer(certificateChain: [], privateKey: .file("fake2.file"), applicationProtocols: ["http/1.1"])
+        XCTAssertFalse(config.bestEffortEquals(differentConfig))
+    }
```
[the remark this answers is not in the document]
madsodgaard: sure thing!

Sources/NIOSSL/TLSConfiguration.swift
Lukasa: I still think we should stick with the closure-based version, but yes agree we should link the bug.

weissi: @Lukasa the problem is that if we do hit the bug, then they will never be equal. The allocations are for reabstraction thunks which aren't equal because they capture (the heap allocs) :P

weissi (approved):
Thank you! That looks good to me!

weissi:

Lukasa (changes requested):
Can you run the `generate_linux_tests.rb` script? That's part of the issues we're seeing in CI: we're missing your new tests on Linux.

madsodgaard:
@Lukasa Done!

weissi:
@Lukasa this is now all green, the failure is the new 5.5 job which doesn't have docker images yet ;)

Lukasa (approved):

PR #281, opened 2021-03-15
Call SecTrustEvaluateAsync from correct queue.

Lukasa:
Motivation:

SecTrustEvaluateAsync documents that it must be called from the same
queue it will call you back on. We failed to do this. Let's fix that up.
While we're here, we should use the non-deprecated API where it's
practical to do so.

Modifications:

- Async out to the queue before calling SecTrustEvaluateAsync.
- Call SecTrustEvaluateAsyncWithError where it's available.

Result:

Following the rules.

### DIFF
--- a/Sources/NIOSSL/SecurityFrameworkCertificateVerification.swift
+++ b/Sources/NIOSSL/SecurityFrameworkCertificateVerification.swift
@@ -55,22 +55,40 @@ extension SSLConnection {
             // We create a DispatchQueue here to be called back on, as this validation may perform network activity.
             let callbackQueue = DispatchQueue(label: "io.swiftnio.ssl.validationCallbackQueue")
 
-            result = SecTrustEvaluateAsync(actualTrust, callbackQueue) { (_, result) in
-                switch result {
-                case .proceed, .unspecified:
-                    // These two cases mean we have successfully validated the certificate. We're done!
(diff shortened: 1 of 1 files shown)


glbrntt (approved):
LGTM

Sources/NIOSSL/SecurityFrameworkCertificateVerification.swift
```diff
@@ -55,24 +55,42 @@ extension SSLConnection {
...
-                    promise.succeed(.certificateVerified)
-                default:
-                    // Oops, we failed.
-                    promise.succeed(.failed)
+            // SecTrustEvaluateAsync and its cousin withError require that they are called from the same queue given to
+            // them as a parameter. Thus, we async away now.
+            callbackQueue.async {
+                let result: OSStatus
+
+                if #available(iOS 13, macOS 10.15, macCatalyst 13, tvOS 13, watchOS 6, *) {
```
glbrntt, line 63: CI isn't happy about this:

```
SecurityFrameworkCertificateVerification.swift:63:66: error: unrecognized platform name 'macCatalyst'
```

glbrntt (approved):

Lukasa:

Lukasa:
Merging over the nightly failure.

PR #282, opened 2021-03-17
Provide notValidBefore and notValidAfter.

Lukasa:
Motivation:

In some cases it's useful to know when a certificate's validity dates
are. This is not always terribly straightforward to calculate, so we can
encapsulate that math within the library.

Modifications:

- Add NIOSSLCertificate.notValidBefore
- Add NIOSSLCertificate.notValidAfter

Result:

Users will be able to ask what certificate validity dates are.

### DIFF
--- a/Sources/NIOSSL/SSLCertificate.swift
+++ b/Sources/NIOSSL/SSLCertificate.swift
@@ -295,6 +295,24 @@ extension NIOSSLCertificate {
         return try readCertificatesFromBIO(bio)
     }
 
+    /// Returns the timestamp before which this certificate is not valid.
+    ///
+    /// The value is in seconds since the UNIX epoch.
+    public var notValidBefore: time_t {
+        // This ref is owned by self.
+        let notBefore = CNIOBoringSSL_X509_get0_notBefore(self.ref)!
+        return notBefore.timeSinceEpoch
+    }
+
+    /// Returns the timestamp after which this certificate is not valid.
+    ///
+    /// The value is in seconds since the UNIX epoch.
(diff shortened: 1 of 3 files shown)


Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -295,6 +295,24 @@ extension NIOSSLCertificate {
         return try readCertificatesFromBIO(bio)
     }
 
+    /// Returns the timestamp before which this certificate is not valid.
+    ///
+    /// The value is in seconds since the UNIX epoch.
+    public var notValidBefore: Int {
```
weissi, line 301: Should this be `NIODeadline`?

weissi (commented):
Just a question about the time type

Sources/NIOSSL/SSLCertificate.swift
Lukasa: I don't think so: `NIODeadline` refers to an instant in time but it is not associated with any other calendaring system. The backing storage is `uptimeNanoseconds`, and is a `UInt64`. This means any valid cert may have an unrepresentable `notBefore` time, as it will likely be before the system start time.

weissi: @Lukasa the `uptimeNanoseconds` is a good point. Do we need another time type in NIO? Because I don't think a naked `Int` is the right type here...

Lukasa: Swapped to time_t.

weissi (approved):
Looks good!

PR #284, opened 2021-03-19
Add support for additional TLS trust roots

simonjbeaumont:
Until now `TLSConfiguration` has support for specifying trust roots
using _one_ of the following:

* Path to file or directory with certificates, or
* A list of `NIOSSLCertificate` the caller has in their hand, or
* The system default.

It's potentially useful to specify a combination of these, e.g. the
system default _and_ some additional provided certificates.

This patch adds a new field, `TLSConfiguration.additionalTrustRoots`
which will be used in combination with the existing
`TLSConfiguration.trustRoots`.

`TLSConfigurationTest` needed the following changes:
* Stop using `connectInMemory` for testing successful handshakes on
  Darwin because it is not thread-safe when used with a verify callback.
* Add Extended Key Usage extenstions to the keys used in tests to
  satisfy the stricter requirements of the Security.framework verifier.

Signed-off-by: Si Beaumont <beaumont@apple.com>


Lukasa:

Lukasa:

Sources/NIOSSL/SSLContext.swift
```diff
@@ -159,10 +159,26 @@ public final class NIOSSLContext {
         returnCode = CNIOBoringSSL_SSL_CTX_set_cipher_list(context, configuration.cipherSuites)
         precondition(1 == returnCode)
 
+        // On non-Linux platforms, when using the platform default trust roots, we make use of a
+        // custom verify callback. If we have also been presented with additional trust roots of
+        // type `.file`, we take the opportunity now to load them in memory to avoid doing so
+        // repeatedly on the request path.
+        //
+        // However, to avoid closely coupling this code with other parts (e.g. the platform-specific
+        // concerns, and the defauling of `trustRoots` to `.default` when `nil`), we unilaterally
```
Lukasa, line 168: Nit: defaulting

```suggestion
        // concerns, and the defaulting of `trustRoots` to `.default` when `nil`), we unilaterally
```

Sources/NIOSSL/SSLContext.swift
```diff
@@ -159,10 +159,26 @@ public final class NIOSSLContext {
...
+        // convert any `additionalTrustRoots` of type `.file` to `.certificates`.
+        var configuration = configuration
+        configuration.additionalTrustRoots = try configuration.additionalTrustRoots.map { trustRoots in
+            guard case .file(let path) = trustRoots else { return trustRoots }
```
Lukasa, line 172: This might be cleaner as a `switch` rather than a `guard case`.

Sources/NIOSSL/SecurityFrameworkCertificateVerification.swift
```diff
@@ -47,11 +47,48 @@ extension SSLConnection {
...
+                case .none, .some(.default):
+                    guard SecTrustSetAnchorCertificatesOnly(actualTrust, false) == errSecSuccess else {
+                        throw NIOSSLError.failedToLoadCertificate
+                    }
+                default:
+                    break
+                }
+                // We should add our additional anchors to any custom ones that already exist.
+                var existingCustomAnchorCertificates: CFArray? = nil
+                guard SecTrustCopyCustomAnchorCertificates(actualTrust, &existingCustomAnchorCertificates) == errSecSuccess else {
```
Lukasa, line 81: This should not be necessary: `SecTrustSetAnchorCertificatesOnly(false)` should be sufficient to trust the built in anchors without us re-adding them.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -42,9 +42,43 @@ public enum NIOSSLPrivateKeySource {
...
+    ///
+    ///     -----BEGIN CERTIFICATE-----
+    ///     ... (CA certificate in base64 encoding) ...
+    ///     -----END CERTIFICATE-----
+    ///
+    /// sequences. Before, between, and after the certificates, text is allowed which can be used e.g.
+    /// for descriptions of the certificates.
+    ///
+    /// If a path to a directory is provided, the files each contain one CA certificate in PEM format.
     case file(String)
```
Lukasa, line 57: When used in the base trust root system this actually doesn't have to be a PEM file, it can be a DER file as well: we pass it directly to the load functions.

Lukasa: Similarly when pointing at a directory those certs do not have to be in PEM format.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -390,6 +469,41 @@ public struct TLSConfiguration {
...
+                                 minimumTLSVersion: TLSVersion = .tlsv1,
+                                 maximumTLSVersion: TLSVersion? = nil,
+                                 certificateVerification: CertificateVerification = .fullVerification,
+                                 trustRoots: NIOSSLTrustRoots = .default,
+                                 certificateChain: [NIOSSLCertificateSource] = [],
+                                 privateKey: NIOSSLPrivateKeySource? = nil,
+                                 applicationProtocols: [String] = [],
+                                 shutdownTimeout: TimeAmount = .seconds(5),
+                                 keyLogCallback: NIOSSLKeyLogCallback? = nil,
+                                 renegotiationSupport: NIORenegotiationSupport,
```
Lukasa, line 492: This one can default `renegotiationSupport` (as we now have it in enough places) to `.none`.

Lukasa (changes requested):
Thanks for this! Generally the patch looks really good. I've got a few notes inline that I'd like to see addressed.

Sources/NIOSSL/TLSConfiguration.swift
simonjbeaumont: Oh, interesting. FWICT `NIOSSLContext.configureCertificateValidation` calls  `NIOSSLContext.loadVerifyLocations` when passed `.file(path)` which just calls through to `SSL_CTX_load_verify_locations` with the path provided.

The documentation I provided on this enum case was based on the man page which states:
> If CAfile is not NULL, it points to a file of CA certificates in PEM format.

and later on

> If CApath is not NULL, it points to a directory containing CA certificates in PEM format.

Link: https://www.openssl.org/docs/man1.0.2/man3/SSL_CTX_load_verify_locations.html

Sources/NIOSSL/SecurityFrameworkCertificateVerification.swift
simonjbeaumont: I had some reason for this, let me re-read the docs and see if my reasoning was flawed and comment back.

Sources/NIOSSL/SSLContext.swift
simonjbeaumont: OK, will do!

Sources/NIOSSL/TLSConfiguration.swift
Lukasa: Yup, you're right. Let's leave the comment as is then. Good catch!

Sources/NIOSSL/SecurityFrameworkCertificateVerification.swift
simonjbeaumont: OK, here is the documentation for `SecTrustSetAnchorCertificates` that led me to include this step...

> If you want to create a set of anchor certificates by modifying the default set, call the `SecTrustCopyAnchorCertificates(_:)` function to obtain the current set of anchor certificates, modify that set as you wish, and create a new array of certificates. Then call `SecTrustSetAnchorCertificates(_:_:)` with the modified array.
>
> Note that when you call the `SecTrustSetAnchorCertificates(_:_:)` function, you are effectively telling the evaluation function to use the anchor certificates in the specified array to evaluate trust regardless of any user-specified trust settings for those certificates. Furthermore, any intermediate certificates based on those anchor certificates are also trusted without consulting user trust settings.

The way I interpreted this was that, without this copy and mutation, we would potentially be removing some of the "built-in" certs from the set of anchors—Do I misunderstand this?

(I also realise that I am calling `SecTrustCopyCustomAnchorCertificates(_:)` (cf. `SecTrustCopyAnchorCertificates(_:)`) which may not be intentional.

I think I wanted to make sure that our use of additional anchors was truly additional w.r.t. what the platform was configured with at the time of the call. I'm not sure what "built-in" means in the docs so tried to make sure I propagated everything.

In light of the above docs, do you think I need to make some tweaks to this read-and-update above, or still think I should remove it altogether?

Lukasa: > The way I interpreted this was that, without this copy and mutation, we would potentially be removing some of the "built-in" certs from the set of anchors—Do I misunderstand this?

Yes, you do, but it's an easy mistake to make. I believe the above advice is supposed to cover only _modifications_: that is, when you want to perform general CRUD operations to the anchors. We don't want to do that, we want only to supplement the default anchors. That's handled by `SecTrustSetAnchorCertificatesOnly`, whose documentation says:

> **anchorCertificatesOnly**
> If true, disables trusting any anchors other than the ones passed in with the SecTrustSetAnchorCertificates(_:_:) function.  If false, the built-in anchor certificates are also trusted. If SecTrustSetAnchorCertificates(_:_:) is called and SecTrustSetAnchorCertificatesOnly(_:_:) is not called, only the anchors explicitly passed in are trusted.

Setting this to `false` in combination with a call to `SecTrustSetAnchorCertificates` should achieve our intended effect.

simonjbeaumont: OK I understand that `SecTrustSetAnchorCertificatesOnly(false)` and `SecTrustSetAnchorCertificates(...)` will mean that we get the "built-in anchor certs" and the ones we pass in in the `SecTrustSetAnchorCertificates(...)` call.

But what about the "custom anchors"? I was worried that these might inadvertently be discarded by these two calls from the verification, which I now remember is why I used `SecTrustCopy**Custom**AnchorCertificates`.

Shall we assume that we do not want these (if any have been configured)?

The docs for `SecTrustCopyCustomAnchorCertificates` state:

> You can use the `SecTrustSetAnchorCertificates(_:_:)` function to set custom anchor certificates.

which led me to think that certs passed into `SecTrustSetAnchorCertificates(_:_:)` are not considered "built-in" w.r.t. the  `SecTrustSetAnchorCertificatesOnly(false)` call?

Lukasa: Certs passed into `SecTrustSetAnchorCertificates` are not built-in, they _are_ the custom anchors. So we don't need to preserve any: there is nowhere else they can have come from but us.

simonjbeaumont: Right. I think you have unlocked something in my understanding here... These custom certs are only custom to the SecTrust that we have created ourselves? They are not representing things that a user may have gone and done in Keychain?

If the above statement of my understanding is now correct then I realise now that I need to be removing this copy.

Sources/NIOSSL/TLSConfiguration.swift
simonjbeaumont: Addressed in https://github.com/apple/swift-nio-ssl/pull/284/commits/8e47d73fdc3270539a2dffdbcc09ef61120a840e.

Sources/NIOSSL/SSLContext.swift
simonjbeaumont: Addressed in https://github.com/apple/swift-nio-ssl/pull/284/commits/a9c3ba7a8d20f4b91be33bd53b8f3a816d751c62.

Sources/NIOSSL/SSLContext.swift
simonjbeaumont: Addressed in https://github.com/apple/swift-nio-ssl/pull/284/commits/0bae5eae1f6588c0d8d9a0ea6269454ea4194581.

Sources/NIOSSL/SecurityFrameworkCertificateVerification.swift
simonjbeaumont: Addressed in https://github.com/apple/swift-nio-ssl/pull/284/commits/ed01d7ad612f3401265d99672e9aa2efcc522f79.

Will leave this comment for you to resolve, @Lukasa, so you can validate my understanding in my previous comment.

Lukasa: Yup, that understanding is correct.

Sources/NIOSSL/SecurityFrameworkCertificateVerification.swift
```diff
@@ -47,11 +47,40 @@ extension SSLConnection {
...
+            result = SecTrustCreateWithCertificates(peerCertificates as CFArray, policy, &trust)
             guard result == errSecSuccess, let actualTrust = trust else {
                 throw NIOSSLError.unableToValidateCertificate
             }
 
+            // If there are additional trust roots then we need to add them to the SecTrust as anchors.
+            let additionalAnchorCertificates: [SecCertificate] = try self.parentContext.configuration.additionalTrustRoots.flatMap { trustRoots -> [NIOSSLCertificate] in
+                switch trustRoots {
+                case .file(let path):
+                    return try NIOSSLCertificate.fromPEMFile(path)
```
Lukasa, line 59: This branch is untakeable, as we did the conversion earlier. We can `preconditionFailure` here.

Sources/NIOSSL/SecurityFrameworkCertificateVerification.swift
```diff
@@ -47,11 +47,40 @@ extension SSLConnection {
...
+            }
+            if !additionalAnchorCertificates.isEmpty {
+                // To use additional anchors _and_ the built-in ones we must reenable the built-in ones expicitly.
+                switch self.parentContext.configuration.trustRoots {
+                case .none, .some(.default):
+                    guard SecTrustSetAnchorCertificatesOnly(actualTrust, false) == errSecSuccess else {
+                        throw NIOSSLError.failedToLoadCertificate
+                    }
+                default:
+                    break
```
Lukasa, line 77: I don't think `break` is right: I think we can only get here in the above branch case. We can again just `preconditionFailure` here.

Lukasa (changes requested):

Sources/NIOSSL/SecurityFrameworkCertificateVerification.swift
simonjbeaumont: OK, I designed it to not be coupled with the other code that did the conversion and saw the conversion in `SSLContext` to be an optimisation.

But I can fail hard here if we want to make it a requirement for the user of this callback to have done any conversion?

Sources/NIOSSL/SecurityFrameworkCertificateVerification.swift
simonjbeaumont: Again this was done for loose coupling reasons. E.g. if we wanted to make use of Security.framework for the case where we have trust roots and additional trust roots but none of them are the platform default (e.g. some files and some `NIOSSLCertificate` values).

I know that currently in this case the code currently goes through BoringSSL for that configuration but do we want to bake that into this layer too?

Sources/NIOSSL/SecurityFrameworkCertificateVerification.swift
Lukasa: I think that would be good, it clarifies intended flow.

Sources/NIOSSL/SecurityFrameworkCertificateVerification.swift
Lukasa: I think we do. Specifically, I think the case where we change the behaviour of the verification code will make us want to think harder about what we're doing, and I'd rather that instead of silently appearing to work it loudly failed. Preconditions can always be removed.

simonjbeaumont: Addressed in https://github.com/apple/swift-nio-ssl/pull/284/commits/9d613d303a6623c384855ea67b52af5a318aeea6.

Sources/NIOSSL/SecurityFrameworkCertificateVerification.swift
simonjbeaumont: Addressed in https://github.com/apple/swift-nio-ssl/pull/284/commits/9d613d303a6623c384855ea67b52af5a318aeea6.

simonjbeaumont:
@Lukasa Do we think this is looking in good shape now?

Lukasa (approved):
LGTM, thanks for the patch @simonjbeaumont.

Lukasa:
Merging over broken nightly.

Issue #253, opened 2020-11-10
Don't report uncleanShutdown when we get EOF during a handshake


Lukasa:
If we're in the middle of a TLS handshake, and we get channelInactive, we currently report this as `uncleanShutdown`. That's not really a good error here: we need to say something more authoritative, like handshake failed.

agnosticdev:
Should we close this?

Lukasa:
Yup!

weissi:
Just as an FYI to others: as of yesterday, this is also released

PR #285, opened 2021-04-07
Bug fix to replace uncleanShutdown with handshakeFailed during TLS failure.

agnosticdev:
Bug fix to replace uncleanShutdown with handshakeFailed during TLS failure.

Motivation:

To address https://github.com/apple/swift-nio-ssl/issues/253

Modifications:

When channelInactive is called during a handshake failure it currently defaults to uncleanShutdown.
This bugfix and test is to report handshakeFailed instead of uncleanShutdown.

Added the handshaking case to the channelInactive function to report handshakeFailed.
Added a new test case, testChannelInactiveDuringHandshake, in UnwrappingTests to test this fix.

Result:

Fix of #253

### DIFF
--- a/Sources/NIOSSL/NIOSSLHandler.swift
+++ b/Sources/NIOSSL/NIOSSLHandler.swift
@@ -98,25 +98,35 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
         // keeping track of the state we're in properly before we do anything else.
         let oldState = state
         state = .closed
+        var channelError: NIOSSLError
 
         switch oldState {
         case .closed, .idle:
             // Nothing to do, but discard any buffered writes we still have.
             discardBufferedWrites(reason: ChannelError.ioOnClosedChannel)
+            // Return early
+            context.fireChannelInactive()
+            return
+        case .handshaking:
+            // In this case the channel is going through the doHandshake steps and
+            // a channelInactive is fired taking down the connection.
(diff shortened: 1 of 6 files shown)


Lukasa:

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -98,25 +98,35 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
         // keeping track of the state we're in properly before we do anything else.
         let oldState = state
         state = .closed
+        var channelError: NIOSSLError
```
Lukasa, line 101: Nit: this can be `let`

Lukasa (commented):
One minor nit here and I think we'll be good to go.

Sources/NIOSSL/NIOSSLHandler.swift
agnosticdev: Good call.  Addressed in fa996f7

Lukasa (approved):
Cool, assuming the tests pass this is ready to go. Thanks for the patch!

Lukasa:
@agnosticdev Mind rebasing on the current `main`? That should fix up the nightly test failures.

agnosticdev:
@Lukasa no problem.  All rebased.

Lukasa:
Hmm, something went weird here and you picked up a bunch of the other commits from main. Mind cleaning the branch up?

agnosticdev:
Closing to manually rebase.

PR #286, opened 2021-04-07
Manage early-exit of ByteBufferBIO in tests.

Lukasa:
Motivation:

Nightly Swift's improvement to the copy propagation pass has caused
ByteBufferBIOTests to start misbehaving, as they now deallocate
some ByteBufferBIOs very early in the test case. This is because the
backing BIO pointer does not have a strong reference to the
ByteBufferBIO.

Modifications:

- Give the backing BIO a strong ref that can be broken by the
    ByteBufferBIO calling close().
- Update the tests to call close()

Result:

No leaking BIOs.

### DIFF
--- a/Sources/NIOSSL/ByteBufferBIO.swift
+++ b/Sources/NIOSSL/ByteBufferBIO.swift
@@ -217,19 +217,34 @@ final class ByteBufferBIO {
             preconditionFailure("Unable to initialize custom BIO")
         }
 
-        // We now need to complete the BIO initialization. The BIO does not have an owned pointer
-        // to us, as that would create an annoying-to-break reference cycle.
+        // We now need to complete the BIO initialization. The BIO takes an owned reference to self here,
+        // which is broken on close().
         self.bioPtr = bio
-        CNIOBoringSSL_BIO_set_data(self.bioPtr, Unmanaged.passUnretained(self).toOpaque())
+        CNIOBoringSSL_BIO_set_data(self.bioPtr, Unmanaged.passRetained(self).toOpaque())
         CNIOBoringSSL_BIO_set_init(self.bioPtr, 1)
         CNIOBoringSSL_BIO_set_shutdown(self.bioPtr, 1)
     }
 
     deinit {
-        // On deinit we need to drop our reference to the BIO, and also ensure that it doesn't hold any
-        // pointers to this object anymore.
-        CNIOBoringSSL_BIO_set_data(self.bioPtr, nil)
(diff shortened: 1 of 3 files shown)


Sources/NIOSSL/ByteBufferBIO.swift
```diff
@@ -217,21 +217,33 @@ final class ByteBufferBIO {
...
+    ///
+    /// This method is idempotent: it is safe to call more than once.
+    internal func close() {
+        guard let selfRef = CNIOBoringSSL_BIO_get_data(self.bioPtr) else {
+            // Shutdown is safe to call more than once.
+            return
+        }
+
+        // We consume the original retain of self, and then nil out the ref in the BIO so that this can't happen again.
+        _ = Unmanaged<ByteBufferBIO>.fromOpaque(selfRef).takeRetainedValue()
```
weissi, line 243: I thought there's a `.release()`?

Lukasa: There is.

Sources/NIOSSL/ByteBufferBIO.swift
```diff
@@ -217,21 +217,33 @@ final class ByteBufferBIO {
...
         CNIOBoringSSL_BIO_set_init(self.bioPtr, 1)
         CNIOBoringSSL_BIO_set_shutdown(self.bioPtr, 1)
     }
 
     deinit {
-        // On deinit we need to drop our reference to the BIO, and also ensure that it doesn't hold any
-        // pointers to this object anymore.
-        CNIOBoringSSL_BIO_set_data(self.bioPtr, nil)
+        // On deinit we need to drop our reference to the BIO.
         CNIOBoringSSL_BIO_free(self.bioPtr)
```
weissi, line 230: could we assert we've been closed?

weissi: prolly you can `CNIOBoringSSL_BIO_get_data(self.bioPtr) == nil` ?

weissi (commented):
just an assertion suggestion

Sources/NIOSSL/ByteBufferBIO.swift
Lukasa: Yup, added an assertion.

weissi (approved):
lgtm!

glbrntt (approved):

Lukasa:
Build timeout.

Lukasa:

PR #287, opened 2021-04-07
Bug fix to replace uncleanShutdown with handshakeFailed during TLS failure.

agnosticdev:
(**Manually rebased**) Bug fix to replace uncleanShutdown with handshakeFailed during TLS failure.

Motivation:

To address #253

Modifications:

When channelInactive is called during a handshake failure it currently defaults to uncleanShutdown.
This bugfix and test is to report handshakeFailed instead of uncleanShutdown.

Added the handshaking case to the channelInactive function to report handshakeFailed.
Added a new test case, testChannelInactiveDuringHandshake, in UnwrappingTests to test this fix.

Result:

Fix of #253


Lukasa (approved):
LGTM, nice.

Issue #207, opened 2020-04-06
Add an API for enumerating the specific chosen ciphers instead of the weird string.


Lukasa:
While we allow setting and unsetting the ciphers using the weird string language, it would be good if we could also get BoringSSL to parse these for us and expose a different object instead. I think to do it we may need to allocate a temporary `SSL_CTX` structure, which is kinda gross, but it's not the end of the world for now.

Lukasa:
I think this works best as a computed property on `TLSConfiguration`, as well as a new `NIOTLSCipher` type. That type is a `struct` that conforms to `RawRepresentable` whose raw type is `UInt16`, as well as a bunch of static lets.

agnosticdev:
Thoughts on a new `NIOTLSCipher` struct:
```swift
public struct NIOTLSCipher: OptionSet {
    public let rawValue: UInt16
    
    public static let ECDH_AESGC =     NIOTLSCipher(rawValue: 1 << 0)
    public static let DH_AESGCM =      NIOTLSCipher(rawValue: 1 << 1)
    public static let DH_CHACHA20 =    NIOTLSCipher(rawValue: 1 << 2)
    public static let ECDH_AES256 =    NIOTLSCipher(rawValue: 1 << 3)
    public static let DH_AES256 =      NIOTLSCipher(rawValue: 1 << 4)
    public static let ECDH_AES128 =    NIOTLSCipher(rawValue: 1 << 5)
    public static let DH_AES =         NIOTLSCipher(rawValue: 1 << 6)
    public static let RSA_AES =        NIOTLSCipher(rawValue: 1 << 7)
    public static let defaultCiphers: NIOTLSCipher = [
        .ECDH_AESGC, .DH_AESGCM, DH_CHACHA20,
        .ECDH_AES256, .DH_AES256, .ECDH_AES128,
        .DH_AES, .RSA_AES
    ]
    public init(rawValue: UInt16) { self.rawValue = rawValue }
}
```
That changes the cipherSuites type from string to `NIOTLSCipher` on `TLSConfiguration`.
```swift
public struct TLSConfiguration {

	...
	
	public var cipherSuites: NIOTLSCipher
	...
	private init(cipherSuites: NIOTLSCipher,
                 minimumTLSVersion: TLSVersion,
                 maximumTLSVersion: TLSVersion?,
                 certificateVerification: CertificateVerification,
                 trustRoots: NIOSSLTrustRoots,
[...]

PR #290, opened 2021-04-28
Added API for setting cipher suites in TLSConfiguration

agnosticdev:
Motivation:

Addressing #207 for adding an API for setting a cipher suite while also preserving the string based declaration.
The NIOTLSCipher struct provides the common IANA cipher names instead while preserving the string based route.

Modifications:

Added NIOTLSCipher struct and added a constructor for client and server to set a new array of cipher in TLSConfiguration.
Added test cases in TLSConfigurationTest.

Result:

Additional API for #207.

### DIFF
--- a/Sources/NIOSSL/TLSConfiguration.swift
+++ b/Sources/NIOSSL/TLSConfiguration.swift
@@ -81,6 +81,39 @@ public enum NIOSSLAdditionalTrustRoots {
     case certificates([NIOSSLCertificate])
 }
 
+/// Available ciphers to use for TLS instead of a string based representation.
+public struct NIOTLSCipher: RawRepresentable, Hashable {
+    public init(rawValue: UInt16) {
+        self.rawValue = rawValue
+    }
+    
+    public init(_ rawValue: RawValue) {
+        self.rawValue = rawValue
+    }
+    
+    public var rawValue: UInt16
+    public typealias RawValue = UInt16
+    
+    public static let TLS_RSA_WITH_AES_128_CBC_SHA                    = NIOTLSCipher(rawValue: 0x2F)
+    public static let TLS_RSA_WITH_AES_256_CBC_SHA                    = NIOTLSCipher(rawValue: 0x35)
+    public static let TLS_RSA_WITH_AES_128_GCM_SHA256                 = NIOTLSCipher(rawValue: 0x9C)
+    public static let TLS_RSA_WITH_AES_256_GCM_SHA384                 = NIOTLSCipher(rawValue: 0x9D)
+    public static let TLS_AES_128_GCM_SHA256                          = NIOTLSCipher(rawValue: 0x1301)
+    public static let TLS_AES_256_GCM_SHA384                          = NIOTLSCipher(rawValue: 0x1302)
+    public static let TLS_CHACHA20_POLY1305_SHA256                    = NIOTLSCipher(rawValue: 0x1303)
+    public static let TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA            = NIOTLSCipher(rawValue: 0xC009)
+    public static let TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA            = NIOTLSCipher(rawValue: 0xC00A)
+    public static let TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA              = NIOTLSCipher(rawValue: 0xC013)
+    public static let TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA              = NIOTLSCipher(rawValue: 0xC014)
+    public static let TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256         = NIOTLSCipher(rawValue: 0xC02B)
+    public static let TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384         = NIOTLSCipher(rawValue: 0xC02C)
+    public static let TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256           = NIOTLSCipher(rawValue: 0xC02F)
+    public static let TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384           = NIOTLSCipher(rawValue: 0xC030)
+    public static let TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256     = NIOTLSCipher(rawValue: 0xCCA8)
+    public static let TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256   = NIOTLSCipher(rawValue: 0xCCA9)
+    
+}
+
 /// Formats NIOSSL supports for serializing keys and certificates.
 public enum NIOSSLSerializationFormats {
     case pem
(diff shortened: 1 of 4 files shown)


Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -200,7 +233,27 @@ public struct TLSConfiguration {
...
     /// The pre-TLS1.3 cipher suites supported by this handler. This uses the OpenSSL cipher string format.
     /// TLS 1.3 cipher suites cannot be configured.
-    public var cipherSuites: String
+    internal var cipherSuiteFormattedString: String = defaultCipherSuites
+    
+    /// Public property used to set the internal cipherSuiteFormattedString from NIOTLSCipher.
+    public var cipherSuite: [NIOTLSCipher] {
+        get {
+            // Default to a ECDHE_RSA based cipher.
+            return [.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256]
```
Lukasa, line 242: This should presumably return the parsed version of `cipherSuiteFormattedString`?

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -200,7 +233,27 @@ public struct TLSConfiguration {
 
     /// The pre-TLS1.3 cipher suites supported by this handler. This uses the OpenSSL cipher string format.
     /// TLS 1.3 cipher suites cannot be configured.
-    public var cipherSuites: String
+    internal var cipherSuiteFormattedString: String = defaultCipherSuites
```
Lukasa, line 236: We cannot hide `public var cipherSuites: String`: it's part of our stable API.

Lukasa (changes requested):
Thanks, this looks like a good start! I've got a few more notes on the change.

Sources/NIOSSL/TLSConfiguration.swift
agnosticdev: Great point.  I believe this was caught in the automated check below as well.  I will address.

Sources/NIOSSL/TLSConfiguration.swift
agnosticdev: I removed the default `TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256` suite here and built an array of suites based on  what is contained in the `defaultCipherSuites` string (which is ultimately set to the cipherSuites property).  I felt like this put `cipherSuite` inline with `defaultCipherSuites`.  However, to do that I had to create a temporary SSL_CTX, set the suites, and then call into the `STACK(SSL_CIPHER)`'s  to extract the `uint16_t` value to match it against the same in `NIOTLSCipher` enum.  Would like to get your thoughts on this route.

Sources/NIOSSL/TLSConfiguration.swift
agnosticdev: This has been addressed in https://github.com/apple/swift-nio-ssl/pull/290/commits/a803d27261ad7f05a4377ece263eaff064107997

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -200,7 +233,45 @@ public struct TLSConfiguration {
...
     /// The pre-TLS1.3 cipher suites supported by this handler. This uses the OpenSSL cipher string format.
     /// TLS 1.3 cipher suites cannot be configured.
-    public var cipherSuites: String
+    public var cipherSuites: String = defaultCipherSuites
+    
+    /// Public property used to set the internal cipherSuiteFormattedString from NIOTLSCipher.
+    public var cipherSuite: [NIOTLSCipher] {
+        get {
+            // Set a temporary context with the default ciphers from cipherSuites
+            guard let tempContext = CNIOBoringSSL_SSL_CTX_new(CNIOBoringSSL_TLS_method()) else {
```
Lukasa, line 242: Rather than use the unsafe API directly, can I propose constructing an `SSLContext` object that we can use? There are a number of ways the unsafe code can be misused (here we're leaking memory, for example), so I think it would be substantially better to extend the safe API on `SSLContext` to address our use-case.

In this case we can do this by simply passing `self` to the `SSLContext` (that is, fully initialising it) and then we can just add a getter that handles `SSL_CTX_get_ciphers` to get the stack.

We'll also want a safe abstraction to handle working with the stack. I've written most of one already in `PeerCertificateChainBuffers`: you could rewrite the majority of that data structure to provide a safe API for iterating `STACK_OF(SSL_CIPHER)`.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -200,7 +233,45 @@ public struct TLSConfiguration {
...
+            return defaultCiphers
+        }
+        set {
+            // Iterate through an array of provided ciphers and build out a dynamic cipherSuites
+            // string based on the passed in suites.
+            var ciphers = ""
+            for cipher in newValue {
+                let cipherValue: UInt16 = cipher.rawValue
+                let boringSSLCipher = CNIOBoringSSL_SSL_get_cipher_by_value(cipherValue)
+                let standardName = String(cString: CNIOBoringSSL_SSL_CIPHER_standard_name(boringSSLCipher))
```
Lukasa, line 268: This is structurally a bit of a nit, but can we make `SSL_get_cipher_by_value` and `SSL_CIPHER_standard_name` part of a method on our `NIOTLSCipher`? Something like:

```
var standardName: String {
}
```

That will clean up this usage-site substantially.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -200,7 +233,45 @@ public struct TLSConfiguration {
...
+                ciphers += "\(standardName):"
```
Lukasa, line 269: Instead of repeatedly appending to this string, we should use `map` to create an array of `Strings` and then use `joined` to join them together.

Lukasa (changes requested):
Great, nice change to the design. I want us to get a handle on the unsafe code to try to constrain its spread, so I've suggested some changes to try to do that. Let me know if you need further clarification on any of the suggestions.

Sources/NIOSSL/TLSConfiguration.swift
agnosticdev: > Rather than use the unsafe API directly, can I propose constructing an SSLContext object that we can use?

Yeah, I like that idea, especially to avoid the leak here.

Regarding:
> In this case we can do this by simply passing self to the SSLContext (that is, fully initialising it) and then we can just 
> add a getter that handles SSL_CTX_get_ciphers to get the stack.

How about something like the following:
```swift
public struct TLSConfiguration {
    ...

    public var cipherSuite: [NIOTLSCipher] {
        get {
            var defaultCiphers: [NIOTLSCipher] = []
            do {
                let sslConext = try NIOSSLContext(configuration: self)
                let cipherIDs = sslConext.getCiphersIDsFromContext()
                for id in cipherIDs {
                    defaultCiphers.append(NIOTLSCipher(id))
                }
            } catch {
                // Catch and return no ciphers
            }
            return defaultCiphers
        }
        set {
            ...
        }
    }
}


public final class NIOSSLContext {

   ...

   public func getCiphersIDsFromContext() -> [UInt16] {
        var cipherIDs: [UInt16] = []

        let bsslCipherStack = CNIOBoringSSL_SSL_CTX_get_ciphers(self.sslContext)
        /*
            Build abstraction for interating `STACK_OF(SSL_CIPHER)`
        */
        return cipherIDs
   }

}
```
Was this what you had in mind? (Minus the safe API for iterating the STACK_OF(SSL_CIPHER))

Lukasa: I think we don't even need the ceremony of `getCipherIDsFromContext`: it can just be `var ciphers: [NIOTLSCipher]`.

Sources/NIOSSL/TLSConfiguration.swift
agnosticdev: This is addressed in https://github.com/apple/swift-nio-ssl/pull/290/commits/1811292f3d5e17db5b3af27cc90f2009870e91d0

Sources/NIOSSL/TLSConfiguration.swift
agnosticdev: This is addressed in https://github.com/apple/swift-nio-ssl/pull/290/commits/1811292f3d5e17db5b3af27cc90f2009870e91d0

Sources/NIOSSL/TLSConfiguration.swift
agnosticdev: I refactored this code to adopt the safer approach here and also implemented the `NIOTLSCipherBuffers` struct in an attempt to safely navigate the `STACK_OF(SSL_CIPHER)`'s based on what you had for the `PeerCertificateChainBuffers`.  Would like to get your feedback here on the latest update.

agnosticdev:
Looks like I need to rebase here as there is a conflict and APIs that I am no longer aligning with.  I will rebase re-open this PR for additional feedback.

Lukasa:
You can rebase and then force push to this branch to avoid needing to reopen the PR, if that's helpful

PR #293, opened 2021-05-05
Rebase: Added API for setting cipher suites in TLSConfiguration

agnosticdev:
Motivation:

Issue-207 for adding an API for setting a cipher suite while also preserving the string based declaration.
Additions to TLSConfiguration and SSLContext to allow safe access to ciphers on a context.

Modifications:

Additions to TLSConfiguration and SSLContext to allow safe access to ciphers on a context.
Added test cases in TLSConfigurationTest.

Result:

Additional API for Issue-207.
Simplified and safer access to STACK_OF(SSL_CIPHER) on a context.

Rebased from (https://github.com/apple/swift-nio-ssl/pull/290)

### DIFF
--- a/Sources/NIOSSL/SSLContext.swift
+++ b/Sources/NIOSSL/SSLContext.swift
@@ -118,6 +118,13 @@ public final class NIOSSLContext {
     private let callbackManager: CallbackManagerProtocol?
     private var keyLogManager: KeyLogCallbackManager?
     internal let configuration: TLSConfiguration
+    
+    /// Access cipher suites applied to the context
+    public var cipherSuites: [NIOTLSCipher] {
+        get {
+            return self.stackOfCipherSuites()
+        }
+    }
 
     /// Initialize a context that will create multiple connections, all with the same
     /// configuration.
@@ -499,6 +506,89 @@ extension NIOSSLContext {
             // either.
             parentSwiftContext.keyLogManager!.log(linePointer)
         }
+    }
+}
+
+// For accessing STACK_OF(SSL_CIPHER) from a SSLContext
+extension NIOSSLContext {
+    /// A collection of buffers representing a STACK_OF(SSL_CIPHER)
+    struct NIOTLSCipherBuffers {
+        private let basePointer: OpaquePointer
+
+        fileprivate init(basePointer: OpaquePointer) {
+            self.basePointer = basePointer
+        }
+    }
+
+    /// Invokes a block with a collection of pointers to STACK_OF(SSL_CIPHER).
+    ///
+    /// The pointers are only guaranteed to be valid for the duration of this call.  This method aligns with the RandomAccessCollection protocol
+    /// to access UInt16 pointers at a specific index.  This pointer is used to safely access id values of the cipher to create a new NIOTLSCipher.
+    func withStackOfCipherSuiteBuffers<Result>(_ body: (NIOTLSCipherBuffers?) throws -> Result) rethrows -> Result {
+        guard let stackPointer = CNIOBoringSSL_SSL_CTX_get_ciphers(self.sslContext) else {
+            return try body(nil)
+        }
+        return try body(NIOTLSCipherBuffers(basePointer: stackPointer))
+    }
+
+    /// Cipher suites derived from the default ciphers on the SSLContext matched against the available ciphers in NIOTLSCipher.
+    func stackOfCipherSuites() -> [NIOTLSCipher] {
+        return self.withStackOfCipherSuiteBuffers { buffers in
+            guard let buffers = buffers else {
+                return []
+            }
+            return buffers.map { NIOTLSCipher($0) }
+        }
+    }
+}
+
+extension NIOSSLContext.NIOTLSCipherBuffers: RandomAccessCollection {
+    
+    struct Index: Hashable, Comparable, Strideable {
+        typealias Stride = Int
+
+        fileprivate var index: Int
+
+        fileprivate init(_ index: Int) {
+            self.index = index
+        }
+
+        static func < (lhs: Index, rhs: Index) -> Bool {
+            return lhs.index < rhs.index
+        }
+
+        func advanced(by n: NIOSSLContext.NIOTLSCipherBuffers.Index.Stride) -> NIOSSLContext.NIOTLSCipherBuffers.Index {
+            var result = self
+            result.index += n
+            return result
+        }
+
+        func distance(to other: NIOSSLContext.NIOTLSCipherBuffers.Index) -> NIOSSLContext.NIOTLSCipherBuffers.Index.Stride {
+            return other.index - self.index
+        }
+    }
+
+    typealias Element = UInt16
+
+    var startIndex: Index {
+        return Index(0)
+    }
+
+    var endIndex: Index {
+        return Index(self.count)
+    }
+
+    var count: Int {
+        return CNIOBoringSSL_sk_SSL_CIPHER_num(self.basePointer)
+    }
+    
+    subscript(position: Index) -> UInt16 {
+        precondition(position < self.endIndex)
+        guard let ptr = CNIOBoringSSL_sk_SSL_CIPHER_value(self.basePointer, position.index) else {
+            preconditionFailure("Unable to locate backing pointer.")
+        }
+        let cipherID = CNIOBoringSSL_SSL_CIPHER_get_protocol_id(ptr) as UInt16
+        return cipherID
     }
 }
 
--- a/Sources/NIOSSL/TLSConfiguration.swift
+++ b/Sources/NIOSSL/TLSConfiguration.swift
@@ -81,6 +81,45 @@ public enum NIOSSLAdditionalTrustRoots {
     case certificates([NIOSSLCertificate])
 }
 
+/// Available ciphers to use for TLS instead of a string based representation.
+public struct NIOTLSCipher: RawRepresentable, Hashable {
+    public init(rawValue: UInt16) {
+        self.rawValue = rawValue
+    }
+    
+    public init(_ rawValue: RawValue) {
+        self.rawValue = rawValue
+    }
+    
+    public var rawValue: UInt16
+    public typealias RawValue = UInt16
+    
+    public static let TLS_RSA_WITH_AES_128_CBC_SHA                    = NIOTLSCipher(rawValue: 0x2F)
+    public static let TLS_RSA_WITH_AES_256_CBC_SHA                    = NIOTLSCipher(rawValue: 0x35)
+    public static let TLS_RSA_WITH_AES_128_GCM_SHA256                 = NIOTLSCipher(rawValue: 0x9C)
+    public static let TLS_RSA_WITH_AES_256_GCM_SHA384                 = NIOTLSCipher(rawValue: 0x9D)
+    public static let TLS_AES_128_GCM_SHA256                          = NIOTLSCipher(rawValue: 0x1301)
+    public static let TLS_AES_256_GCM_SHA384                          = NIOTLSCipher(rawValue: 0x1302)
+    public static let TLS_CHACHA20_POLY1305_SHA256                    = NIOTLSCipher(rawValue: 0x1303)
+    public static let TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA            = NIOTLSCipher(rawValue: 0xC009)
+    public static let TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA            = NIOTLSCipher(rawValue: 0xC00A)
+    public static let TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA              = NIOTLSCipher(rawValue: 0xC013)
+    public static let TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA              = NIOTLSCipher(rawValue: 0xC014)
+    public static let TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256         = NIOTLSCipher(rawValue: 0xC02B)
+    public static let TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384         = NIOTLSCipher(rawValue: 0xC02C)
+    public static let TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256           = NIOTLSCipher(rawValue: 0xC02F)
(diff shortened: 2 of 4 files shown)


Sources/NIOSSL/SSLContext.swift
```diff
@@ -502,6 +509,89 @@ extension NIOSSLContext {
...
+        fileprivate init(basePointer: OpaquePointer) {
+            self.basePointer = basePointer
+        }
+    }
+
+    /// Invokes a block with a collection of pointers to STACK_OF(SSL_CIPHER).
+    ///
+    /// The pointers are only guaranteed to be valid for the duration of this call.  This method aligns with the RandomAccessCollection protocol
+    /// to access UInt16 pointers at a specific index.  This pointer is used to safely access id values of the cipher to create a new NIOTLSCipher.
+    func withStackOfCipherSuiteBuffers<Result>(_ body: (NIOTLSCipherBuffers?) throws -> Result) rethrows -> Result {
```
Lukasa, line 527: This may as well be `fileprivate` as we don't need anyone else to use it, I don't think.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -502,6 +509,89 @@ extension NIOSSLContext {
...
+        guard let stackPointer = CNIOBoringSSL_SSL_CTX_get_ciphers(self.sslContext) else {
+            return try body(nil)
+        }
+        return try body(NIOTLSCipherBuffers(basePointer: stackPointer))
+    }
+
+    /// Cipher suites derived from the default ciphers on the SSLContext matched against the available ciphers in NIOTLSCipher.
```
Lukasa, line 534: This comment isn't strictly accurate: we will represent them all, we just won't necessarily have nice names for some of them.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -502,6 +509,89 @@ extension NIOSSLContext {
...
+            result.index += n
+            return result
+        }
+
+        func distance(to other: NIOSSLContext.NIOTLSCipherBuffers.Index) -> NIOSSLContext.NIOTLSCipherBuffers.Index.Stride {
+            return other.index - self.index
+        }
+    }
+
+    typealias Element = UInt16
```
Lukasa, line 571: I think we may as well make the `Element` type `NIOTLSCipher` and do the instantiation in here.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -502,6 +509,89 @@ extension NIOSSLContext {
...
+    var endIndex: Index {
+        return Index(self.count)
+    }
+
+    var count: Int {
+        return CNIOBoringSSL_sk_SSL_CIPHER_num(self.basePointer)
+    }
+    
+    subscript(position: Index) -> UInt16 {
+        precondition(position < self.endIndex)
```
Lukasa, line 586: Can we also `precondition(position >= self.startIndex)`?

Sources/NIOSSL/SSLContext.swift
```diff
@@ -502,6 +509,89 @@ extension NIOSSLContext {
...
+        guard let ptr = CNIOBoringSSL_sk_SSL_CIPHER_value(self.basePointer, position.index) else {
+            preconditionFailure("Unable to locate backing pointer.")
+        }
+        let cipherID = CNIOBoringSSL_SSL_CIPHER_get_protocol_id(ptr) as UInt16
```
Lukasa, line 590: I think this `as` does nothing and so it can be removed.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -81,6 +81,45 @@ public enum NIOSSLAdditionalTrustRoots {
...
+    public static let TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA              = NIOTLSCipher(rawValue: 0xC014)
+    public static let TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256         = NIOTLSCipher(rawValue: 0xC02B)
+    public static let TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384         = NIOTLSCipher(rawValue: 0xC02C)
+    public static let TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256           = NIOTLSCipher(rawValue: 0xC02F)
+    public static let TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384           = NIOTLSCipher(rawValue: 0xC030)
+    public static let TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256     = NIOTLSCipher(rawValue: 0xCCA8)
+    public static let TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256   = NIOTLSCipher(rawValue: 0xCCA9)
+    
+    var standardName: String {
+        get {
```
Lukasa, line 116: We can remove the superfluous `get` here: if the property is get-only you can just write the definition directly without needing the `get` block.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -118,6 +118,13 @@ public final class NIOSSLContext {
     private let callbackManager: CallbackManagerProtocol?
     private var keyLogManager: KeyLogCallbackManager?
     internal let configuration: TLSConfiguration
+    
+    /// Access cipher suites applied to the context
+    public var cipherSuites: [NIOTLSCipher] {
```
Lukasa, line 123: This API does not need to be public, and it's a bit duplicative of `stackOfCipherSuites`. I think we should just make `stackOfCipherSuites` a `var` and `internal` and call it `cipherSuites`.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -200,7 +239,21 @@ public struct TLSConfiguration {
 
     /// The pre-TLS1.3 cipher suites supported by this handler. This uses the OpenSSL cipher string format.
     /// TLS 1.3 cipher suites cannot be configured.
-    public var cipherSuites: String
+    public var cipherSuites: String = defaultCipherSuites
+    
+    /// Public property used to set the internal cipherSuiteFormattedString from NIOTLSCipher.
+    public var cipherSuite: [NIOTLSCipher] {
+        get {
+            guard let sslConext = try? NIOSSLContext(configuration: self) else {
```
Lukasa, line 247: Nit: spelling

```suggestion
            guard let sslContext = try? NIOSSLContext(configuration: self) else {
```

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -200,7 +239,21 @@ public struct TLSConfiguration {
...
+                return []
+            }
+            return sslConext.cipherSuites
+        }
+        set {
+            let assignedCiphers = newValue.map { $0.standardName }
+            cipherSuites = assignedCiphers.joined(separator: ":")
```
Lukasa, line 254: Nit: we use explicit `self` in SwiftNIO where we can.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -282,6 +335,70 @@ public struct TLSConfiguration {
         self.applicationProtocols = applicationProtocols
         self.keyLogCallback = keyLogCallback
     }
+    
+    private init(cipherSuites: [NIOTLSCipher],
```
Lukasa, line 339: I think it might be nicer, instead of duplicating the memberwise init, to instead implement this on top of the existing one and just overriding the cipher suites afterwards.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -200,7 +239,21 @@ public struct TLSConfiguration {
 
     /// The pre-TLS1.3 cipher suites supported by this handler. This uses the OpenSSL cipher string format.
     /// TLS 1.3 cipher suites cannot be configured.
-    public var cipherSuites: String
+    public var cipherSuites: String = defaultCipherSuites
+    
+    /// Public property used to set the internal cipherSuiteFormattedString from NIOTLSCipher.
+    public var cipherSuite: [NIOTLSCipher] {
```
Lukasa, line 245: I think this needs a better name. Let's call it `cipherSuiteValues`?

Lukasa (changes requested):
Cool, looks a lot better! I've left some more notes in the diff.

Sources/NIOSSL/TLSConfiguration.swift
agnosticdev: Do you mean to add another parameter to the existing private init?  Something like?
```swift
private init(cipherSuites: String,
		   cipherSuiteValues: [NIOTLSCipher],
		    ...
			 )
```

Sources/NIOSSL/TLSConfiguration.swift
agnosticdev: Resolved in https://github.com/apple/swift-nio-ssl/pull/293/commits/dc9625669fc700a8f1cb8a064ad4c59a84fc6563

Sources/NIOSSL/TLSConfiguration.swift
agnosticdev: Resolved in https://github.com/apple/swift-nio-ssl/pull/293/commits/dc9625669fc700a8f1cb8a064ad4c59a84fc6563

Sources/NIOSSL/TLSConfiguration.swift
agnosticdev: Thank you!  Resolve in https://github.com/apple/swift-nio-ssl/pull/293/commits/dc9625669fc700a8f1cb8a064ad4c59a84fc6563

Sources/NIOSSL/SSLContext.swift
agnosticdev: Good call.  Resolved in https://github.com/apple/swift-nio-ssl/pull/293/commits/dc9625669fc700a8f1cb8a064ad4c59a84fc6563

Sources/NIOSSL/TLSConfiguration.swift
agnosticdev: Resolved in https://github.com/apple/swift-nio-ssl/pull/293/commits/dc9625669fc700a8f1cb8a064ad4c59a84fc6563

Sources/NIOSSL/SSLContext.swift
agnosticdev: Resolved in https://github.com/apple/swift-nio-ssl/pull/293/commits/dc9625669fc700a8f1cb8a064ad4c59a84fc6563

Sources/NIOSSL/SSLContext.swift
agnosticdev: Resolved in https://github.com/apple/swift-nio-ssl/pull/293/commits/dc9625669fc700a8f1cb8a064ad4c59a84fc6563

Sources/NIOSSL/SSLContext.swift
agnosticdev: Resolved in https://github.com/apple/swift-nio-ssl/pull/293/commits/dc9625669fc700a8f1cb8a064ad4c59a84fc6563

Sources/NIOSSL/SSLContext.swift
agnosticdev: Resolved in https://github.com/apple/swift-nio-ssl/pull/293/commits/dc9625669fc700a8f1cb8a064ad4c59a84fc6563

Sources/NIOSSL/SSLContext.swift
agnosticdev: Resolved in https://github.com/apple/swift-nio-ssl/pull/293/commits/dc9625669fc700a8f1cb8a064ad4c59a84fc6563

agnosticdev:
I just pushed a change to remove duplicating the memberwise init.  Let me know your thoughts.

agnosticdev:
Looks like `testNewCallbackCanDelayHandshake` timed out?

Sources/NIOSSL/TLSConfiguration.swift
Lukasa: Yeah, and we can just pass the default string.

weissi:

Sources/NIOSSL/SSLContext.swift
```diff
@@ -502,6 +502,90 @@ extension NIOSSLContext {
...
+        return try body(NIOTLSCipherBuffers(basePointer: stackPointer))
+    }
+
+    /// Access cipher suites applied to the context
+    internal var cipherSuites: [NIOTLSCipher] {
+        return self.withStackOfCipherSuiteBuffers { buffers in
+            guard let buffers = buffers else {
+                return []
+            }
+            return buffers.map { $0 }
```
Lukasa, line 533: Nit: I think it's better to just use the `Array` constructor.

```suggestion
            return Array(buffers)
```

Tests/NIOSSLTests/TLSConfigurationTest.swift
```diff
@@ -434,6 +437,7 @@ class TLSConfigurationTest: XCTestCase {
         let serverConfig = TLSConfiguration.forServer(
             certificateChain: [.certificate(TLSConfigurationTest.cert1)],
             privateKey: .privateKey(TLSConfigurationTest.key1),
+            cipherSuites: defaultCipherSuites,
```
Lukasa, line 440: I think the above changes in this file are unnecessary.

Lukasa (changes requested):
Great, this is looking really good! I've left two notes in the diff. One request: can we add some tests to the TLSConfiguration tests that validate that changes to one of the cipher suite operations is reflected in the others? I.e. a default value can get both lists, and then that changing one changes the other and vice versa?

Lukasa:
I don't think something needs to be done to make them align, really.

BoringSSL inherits the idea of cipher suite groups from OpenSSL. It really only manifests in the magical string configuration language, and most of the time we don't care about it. It's totally fine for you to set a short string that manifests in lots of cipher suites, not all of which we name.

weissi:
@Lukasa I think we've not got the tests.

weissi:
@agnosticdev feel free to re-request review (click the ♻ button next to the reviewer name) when you addressed the feedback. That way we'll get notified. I just re-requests Cory's review here because I think you addressed the feedback.

Lukasa (approved):
Nice one, LGTM. Thanks @agnosticdev!

PR #294, opened 2021-05-10
Explicit `time_t` import

helje5:
W/o this `time_t` is only available as an implicit import via `CNIOBoringSSL`, which is `@_implementionOnly`.

The issue popped up in `swift-lambda`: https://github.com/SwiftXcode/swift-lambda/issues/6

I suspect it is due to slightly different import ordering (which makes the compiler pickup the alias). It probably is not specific to SPMDestinations, the import order is just undefined?


Lukasa (approved):
Good catch, this seems like a good fix.

Lukasa:

PR #299, opened 2021-05-13
Rationalise TLSConfiguration construction.

Lukasa:
Motivation:

Currently whenever we add new config to TLSConfiguration, we add a new
static factory function that needs to provide the field. This is because
we currently allow configuration primarily by way of initializers. This
is increasingly not scaling, leading to a proliferation of almost
identical static factory functions.

We can replace this proliferation by having only "default" constructions
that default basically everything, and then having users use
getters/setters to initialize things. This also works well when we have
multiple ways of interacting with config, e.g. with the various cipher
suite operations.

Modifications:

- Deprecate all existing factory functions.
- Replace with two: forClient() and
  forServer(certificateChain:privateKey:). These configure only the
  mandatory things for each mode.
- Replace all uses of the deprecated initializers with the new ones,
  configuring fields manually as needed.

Result:

More consistent configuration objects.

### DIFF
--- a/Sources/NIOSSL/TLSConfiguration.swift
+++ b/Sources/NIOSSL/TLSConfiguration.swift
@@ -227,7 +227,7 @@ internal func decodeALPNIdentifier(identifier: [UInt8]) -> String {
 /// Manages configuration of TLS for SwiftNIO programs.
 public struct TLSConfiguration {
     /// A default TLS configuration for client use.
-    public static let clientDefault = TLSConfiguration.forClient()
+    public static let clientDefault = TLSConfiguration.makeClientConfiguration()
 
     /// The minimum TLS version to allow in negotiation. Defaults to tlsv1.
     public var minimumTLSVersion: TLSVersion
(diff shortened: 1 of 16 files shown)


Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -337,11 +337,120 @@ public struct TLSConfiguration {
...
+        hasher.combine(renegotiationSupport)
+    }
+
+    /// Creates a TLS configuration for use with client-side contexts.
+    ///
+    /// This provides sensible defaults, and can be used without customisation. For server-side
+    /// contexts, you should use `forServer` instead.
+    ///
+    /// For customising fields, modify the returned TLSConfiguration object.
+    public static func forClient() -> TLSConfiguration {
```
weissi, line 402: I think I'm okay with `.forClient/Server` but we should consciously decide that we're not sticking to the ["Begin names of factory methods with “make”, e.g. x.makeIterator()."](https://swift.org/documentation/api-design-guidelines/#begin-factory-name-with-make) Swift naming guideline here.

Other options could be:
  - `(TLSConfiguration).makeClientTLSConfiguration()`
  - `(TLSConfiguration).makeClientConfiguration()`

Sadly, we can't really go for  `init`s here because the client constructor doesn't have arguments so there's no where to stick the word "client".

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -337,11 +337,120 @@ public struct TLSConfiguration {
...
+}
+
+// MARK: Deprecated constructors.
+
+extension TLSConfiguration {
     /// Create a TLS configuration for use with server-side contexts. This allows setting the `NIOTLSCipher` property specifically.
     ///
     /// This provides sensible defaults while requiring that you provide any data that is necessary
     /// for server-side function. For client use, try `forClient` instead.
+    @available(*, deprecated, renamed: "TLSConfiguration.forServer(certificateChain:privateKey:)")
```
glbrntt, line 453: Is this right (and elsewhere)? I _think_ we want just `"forServer(certificateChain:privateKey:)"` here otherwise Xcode shoves in the `TLSConfiguration` as well when you apply the fix its.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -465,11 +577,12 @@ public struct TLSConfiguration {
...
                                 additionalTrustRoots: additionalTrustRoots)
     }
-    
+
     /// Creates a TLS configuration for use with client-side contexts. This allows setting the `NIOTLSCipher` property specifically.
     ///
     /// This provides sensible defaults, and can be used without customisation. For server-side
     /// contexts, you should use `forServer` instead.
+    @available(*, deprecated, renamed: "TLSConfiguration.forClient()")
     public static func forClient(cipherSuites: [NIOTLSCipher],
```
agnosticdev, line 586: Does deprecating these constructors move away from setting fields like `[NIOTLSCipher]` on `forClient` and `forServer` and move these customizations towards something like?
```swift
var client = TLSConfiguration.forClient()
client.cipherSuiteValues = [.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384]
```

Lukasa: Yes, that's the goal. This avoids the painful process you had to go through of repeatedly adding new initialisers.

agnosticdev: Okay, thank you.

glbrntt (approved):
LGTM assuming the CI turns green

Tests/NIOSSLTests/TLSConfigurationTest.swift
```diff
@@ -259,197 +259,226 @@ class TLSConfigurationTest: XCTestCase {
...
-        let serverConfig = TLSConfiguration.forServer(certificateChain: [.certificate(TLSConfigurationTest.cert1)],
-                                                      privateKey: .privateKey(TLSConfigurationTest.key1),
-                                                      cipherSuites: [.TLS_RSA_WITH_AES_256_CBC_SHA],
-                                                      maximumTLSVersion: .tlsv12)
+        var clientConfig = TLSConfiguration.makeClientConfiguration()
+        clientConfig.cipherSuiteValues = [.TLS_RSA_WITH_AES_128_CBC_SHA]
+        clientConfig.maximumTLSVersion = .tlsv12
+        clientConfig.trustRoots = .certificates([TLSConfigurationTest.cert1])
+
+
```
fabianfett, line 279:
```suggestion

```

fabianfett (approved):
I think this is a very sensible change!

weissi (approved):
Awesome, this looks great to me!

Davidde94 (approved):

Lukasa:

Lukasa:

Lukasa:

Lukasa:

PR #300, opened 2021-06-14
Enable tests for Android

buttaface:
Motivation:

Run the tests on Android.

Modifications:

- Enable the tests on Android.
- Replace ~all~a few uses of `/tmp` with `FileManager.default.temporaryDirectory` in the Swift tests.

Result:

All the tests pass on Android AArch64.

I added swift-nio-ssh to my daily Android CI for Swift 5.4 yesterday, buttaface/swift-android-sdk@a023ecb. With this pull, I can add this repo too without patching it locally.


Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
```diff
@@ -1004,7 +1006,7 @@ class NIOSSLIntegrationTest: XCTestCase {
         let serverCtx = try configuredSSLContext()
         var clientConfig = TLSConfiguration.makeClientConfiguration()
         clientConfig.certificateVerification = .noHostnameVerification
-        clientConfig.trustRoots = .file("/tmp")
+        clientConfig.trustRoots = .file(tmpDir)
```
Lukasa, line 1009: I think this is the only one of the changes in this file we actually need.

Tests/NIOSSLTests/UnwrappingTests.swift
```diff
@@ -16,8 +16,10 @@ import XCTest
 import NIO
 @testable import NIOSSL
 
+private let tmpDir = FileManager.default.temporaryDirectory.path
+
 func connectInMemory(client: EmbeddedChannel, server: EmbeddedChannel) throws {
-    let addr = try assertNoThrowWithValue(SocketAddress(unixDomainSocketPath: "/tmp/whatever2"))
+    let addr = try assertNoThrowWithValue(SocketAddress(unixDomainSocketPath: "\(tmpDir)/whatever2"))
```
Lukasa, line 22: We shouldn't need the changes in this file.

Lukasa (changes requested):
The changes for the tmpdir got a bit excessive: the ones used to `EmbeddedChannel.connect` are fake and never touch the FS, so it doesn't matter what they are. If we can remove those changes we should be good.

Lukasa:

Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
buttaface: Yep, changed this pull to only change this line in this file.

Tests/NIOSSLTests/UnwrappingTests.swift
buttaface: Yeah, I saw some tests that call `/tmp` failing on Android, so I search-and-replaced all of them. These two work fine the way they were, so I reverted in the latest push.

Lukasa (approved):
Nice, LGTM.

Lukasa:

PR #306, opened 2021-07-22
Android: FILE is an opaque struct since API 24

buttaface:
Motivation:

Get this Swift file building for Android.

Modifications:

Use an OpaquePointer for `FILE*` on Android.

Result:

This file builds and all the tests pass on Android.

This change is needed because of aosp-mirror/platform_bionic@3037ea4, as noted on the similar pull apple/swift-tools-support-core#243.

An earlier unconditionally-compiled version of this patch got this package cross-compiled for Android, buttaface/swift-android-sdk#11, and all the tests passing in the Android emulator, so that I can add it to my daily Android CI next. I also built it on linux to make sure it didn't break anything there.

### DIFF
--- a/Sources/NIOSSL/PosixPort.swift
+++ b/Sources/NIOSSL/PosixPort.swift
@@ -22,10 +22,16 @@
 // can lift anything missing from there and move it over without change.
 import NIO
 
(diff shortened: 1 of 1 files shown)


Lukasa (approved):
Thanks for this fix!

Lukasa:

Issue #231, opened 2020-07-03
Send client certificate CA names


fourplusone:
In order to perform certificate based authentication, clients expect to receive a list of accepted root CAs. However, the example code below fails to send this information (an implementation for `selfSignedCertificate()` can be found at https://gist.github.com/fourplusone/3561b888d26123814da69a213b246a4a)

```swift
import NIO
import NIOSSL

let (serverCert, serverKey) = try! selfSignedCertificate()
let (clientCert, _) = try! selfSignedCertificate()

let configuration = TLSConfiguration.forServer(certificateChain: [.certificate(serverCert)],
                                        privateKey: .privateKey(serverKey),
                                        trustRoots: .certificates([clientCert]))

let sslContext = try NIOSSLContext(configuration: configuration)
let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)

let channel = try! ServerBootstrap(group: group).childChannelInitializer { channel in
    let handler = NIOSSLServerHandler(context: sslContext)
    return channel.pipeline.addHandler(handler)
    }.bind(host: "127.0.0.1", port: 8443).wait()
try! channel.closeFuture.wait()
```

This behaviour can be verified by using `openssl s_client -connect localhost:8443`:

```
...
No client certificate CA names sent
Server Temp Key: ECDH, X25519, 253 bits
...
```

Lukasa:
Yup, we don't have support for this functionality today. To add it we'd probably want to add a new field to `TLSConfiguration` to control the behaviour, `sendCANameList` and default it to `true`. Then we can arrange to do the dance to call `SSL_CTX_set_client_CA_list`.

agnosticdev:
Support for CA names has now been added.

PR #307, opened 2021-08-12
issue-231: Adding CA Names to the TLS handshake.

agnosticdev:
Motivation:

This is an enhancement to add support for send CA names during the handshake for #231.

Modifications:

Changes to:
TLSConfiguration.swift: Added the `sendCANameList` flag to indicate loading the CA names on `.fullVerification`.
SSLContext.swift: Added the CA Names derived from the `trustRoots`.
TLSConfigurationTests.swift: Added the test case for `testFullVerificationWithCANames`.
NIOSSLTestHelpers.swift: Added a new set of PEMs for TLS and client authentication testing.

Result:

Addition for #231.

I have a few concerns about this PR and would like to get some feedback.

1. I added 5 new testing PEMs in `NIOSSLTestHelpers.swift` to represent a root from a CustomCA that issued an identity for TLS and for client authentication.  Testing client authentication would not work with the CertificateVerification case of .fullVerification without a client identity issued from the CustomCA. I also did not see a client identity with the Extended Key Usage for Client Authentication in `NIOSSLTestHelpers.swift` so this was another reason why I added these PEMs. 

2. The addition of `NIOX509NameListBuffers`.  This is essentially adding overhead to `SSLContext.swift` just so this functionality can be tested.  Normally I would try and come at this from the `NIOSSLContext` instance that is created in the test case but that does not allow access to the `OpaquePointer` for sslContext (and rightfully so).  So this type is shimmed in there for testing access to the context.

3. The addition of he PEMs in item 1 have to account for `CERTIFICATE_VERIFY_FAILED` case when tested on remote systems because the root certificate is not installed in the trust store.  I suspect that this is the way it has to be in order to test this, but wanted to call it out as an addition to item 1.


Sources/NIOSSL/SSLContext.swift
```diff
@@ -434,15 +434,32 @@ extension NIOSSLContext {
...
+            }
 
             func configureTrustRoots(trustRoots: NIOSSLTrustRoots) throws {
                 switch trustRoots {
                 case .default:
                     try NIOSSLContext.platformDefaultConfiguration(context: context)
                 case .file(let path):
                     try NIOSSLContext.loadVerifyLocations(path, context: context)
+                    // Add the CA name from the trust root
+                    if sendCANames {
```
Lukasa, line 452: This code path is not quite right: if you crack open `loadVerifyLocations` you can see that we may have many certificates within the specific location that we might need to handle.

In this instance we probably need to use `loadVerifyLocations` to help us out. If the path is a file, we can use `SSL_load_client_CA_file` to give us a stack of X509 names and use those. If the path is a directory, we have to work a little bit harder, and enumerate the files in the directory and load them.

We'll need test cases for all of these cases as well.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -434,15 +434,32 @@ extension NIOSSLContext {
...
+                        try loadCACertificate(certificate: NIOSSLCertificate(file: path, format: .pem))
+                    }
                 case .certificates(let certs):
-                    try certs.forEach { try NIOSSLContext.addRootCertificate($0, context: context) }
+                    try certs.forEach {
```
Lukasa, line 456: I'd rather use `for cert in certs` than use `forEach`.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -520,6 +537,78 @@ extension NIOSSLContext {
     }
 }
 
+extension NIOSSLContext {
+    /// A collection of buffers representing a STACK_OF(X509_NAME)
+    struct NIOX509NameListBuffers {
```
Lukasa, line 542: I think if we're only going to use `count` then for now let's save ourself some complexity and just have `getX509NameListCount` call `CNIOBoringSSL_sk_X509_NAME_num` directly. Otherwise, we should arrange a more complex implementation that can actually be iterated, but for now I think that's excessive.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -302,6 +302,10 @@ public struct TLSConfiguration {
 
     /// Whether renegotiation is supported.
     public var renegotiationSupport: NIORenegotiationSupport
+    
+    /// Send the CA names derived from the `trustRoots`  for client authentication.
+    /// This instructs the client which identities can be used by evaluating what CA the identity certificate was issued from.
+    public var sendCANameList: Bool = true
```
Lukasa, line 308: Why does this default to `true` here but `false` elsewhere? I think we can just remove the default here.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -317,7 +321,8 @@ public struct TLSConfiguration {
                  shutdownTimeout: TimeAmount,
                  keyLogCallback: NIOSSLKeyLogCallback?,
                  renegotiationSupport: NIORenegotiationSupport,
-                 additionalTrustRoots: [NIOSSLAdditionalTrustRoots]) {
+                 additionalTrustRoots: [NIOSSLAdditionalTrustRoots],
+                 sendCANameList: Bool = true) {
```
Lukasa, line 325: I don't think we should default this here either.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -439,7 +446,8 @@ extension TLSConfiguration {
                                 shutdownTimeout: .seconds(5),
                                 keyLogCallback: nil,
                                 renegotiationSupport: .none,
-                                additionalTrustRoots: [])
+                                additionalTrustRoots: [],
+                                sendCANameList: true)
```
Lukasa, line 450: I'm inclined to say this shouldn't default to `true`: it's an unexpected change in behaviour. The correct default is `false`.

Lukasa (changes requested):

Sources/NIOSSL/TLSConfiguration.swift
agnosticdev: If we remove the default here on the private constructor I would need to add a default value to all of the deprecated constructors, correct?  If I changed the default value to `false` here then the both the deprecated and new constructors could take advantage of overriding this property in the same way.  For example:
```swift
var client = TLSConfiguration.forClient(...)
client.sendCANameList = true
        
var client = TLSConfiguration.makeClientConfiguration()
client.sendCANameList = true
```
Thoughts?

Sources/NIOSSL/SSLContext.swift
agnosticdev: Great idea!

Sources/NIOSSL/SSLContext.swift
agnosticdev: Good call out.  I'll get a fix together for this with some tests.

Sources/NIOSSL/TLSConfiguration.swift
Lukasa: Oh, good note. In that case, change the default value to `false` here.

agnosticdev: Resolved in the latest commit.

Sources/NIOSSL/SSLContext.swift
agnosticdev: Resolved in the latest commit.

Sources/NIOSSL/SSLContext.swift
agnosticdev: Added a new code path here in the latest commit.

agnosticdev:
@Lukasa  Notes on the latest commit:
1. I'm not too wild about having separate code paths for loading the CA names depending on whether the `trustRoot` is a file path or a directory, but `CNIOBoringSSL_SSL_load_client_CA_file` seem like it will work well in the case that the PEM has many certificates in it.  Thoughts?
2. Out in `loadVerifyLocations` I created a routine to scan a directory.  Do you want me to leave that there or move it somewhere else?

agnosticdev:
Looks like there was some failures on the latest commit.  One of them was a gem failure for Jazzy on Swift 5.3, let me know if I need to refactor or resubmit in any way.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -451,8 +457,15 @@ extension NIOSSLContext {
             break
         }
     }
+    
+    private static func addCACertificateNameToList(context: OpaquePointer, certificate: NIOSSLCertificate) throws {
+        // Adds the CA name extracted from cert to the list of CAs sent to the client when requesting a client certificate.
+        guard 1 == CNIOBoringSSL_SSL_CTX_add_client_CA(context, certificate.ref) else {
```
Lukasa, line 463: We have to be a bit careful with lifetimes here: we need to use `withExtendedLifetime(certificate)` around this entire block to make sure we don't accidentally release it.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -472,6 +485,32 @@ extension NIOSSLContext {
...
+                while let dirent: UnsafeMutablePointer<dirent> = readdir(dir) {
+                    let fileName = withUnsafePointer(to: &dirent.pointee.d_name) { (ptr) -> String in
+                        return ptr.withMemoryRebound(to: CChar.self, capacity: Int(ptr.pointee.0) + 1) { String(cString: $0) }
+                    }
+                    if fileName.suffix(4) == ".pem" {
+                        pemFilePaths.append(path + fileName)
+                    }
+                }
+                closedir(dir)
+            }
```
Lukasa, line 505: We need to factor this code out into its own method for more easy auditing.

Lukasa: Actually, maybe we should just refactor the directory iteration code: the sequence of calls to `opendir`, `readdir`, and `closedir`. This seems like a great candidate to be a custom `Sequence` whose element type is `String`. Then we can simply express `pemFilePaths` as `pemFilePaths = DirectoryContents(path).filter { $0.suffix(4) == ".pem" }`.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -472,6 +485,32 @@ extension NIOSSLContext {
...
+            // Load the PEM files one by one and perform the following functions:
+            // 1. addRootCertificate to the context.
+            // 2. addCACertificateNameToList to send the CA names during the handshake.
+            for path in pemFilePaths {
+                let cert = try NIOSSLCertificate(file: path, format: .pem)
+                try NIOSSLContext.addRootCertificate(cert, context: context)
```
Lukasa, line 511: I don't think this call is needed: we've already added the certificate in the load verify locations call.

Tests/NIOSSLTests/TLSConfigurationTest.swift
```diff
@@ -254,6 +254,67 @@ class TLSConfigurationTest: XCTestCase {
...
+                XCTFail("Unexpected error: \(err.localizedDescription)")
+            }
+            _ = clientChannel.close()
+        }
+        
+        clientChannel.closeFuture.whenComplete { _ in
+            XCTAssertEqual(eventHandler.errors.count, 1)
+            
+            guard let error = eventHandler.errors.first else {
+                XCTFail("Could not unwrapp expected .uncleanShutdown error")
```
Lukasa, line 277:
```suggestion
                XCTFail("Could not unwrap expected .uncleanShutdown error")
```

Sources/NIOSSL/SSLContext.swift
Lukasa: Note that there's also no good reason to constrain ourselves only to `PEM`. Instead, we should probably just try both `pem` and `der` for a given cert, probably hinted by (but not limited by) the filename.

Lukasa (changes requested):

Tests/NIOSSLTests/TLSConfigurationTest.swift
agnosticdev: Fixed in latest commit.

Sources/NIOSSL/SSLContext.swift
agnosticdev: It does look like `SSL_CTX_load_verify_locations` is supposed to load the certificate.  Now, with and without the usage of `NIOSSLContext.addRootCertificate` the handshake succeeds from an XCTest.  However, when removing `NIOSSLContext.addRootCertificate`, and WITH the certificate chain is installed on the machine, `BoringSSLError` throws  `TLSV1_ALERT_UNKNOWN_CA` which does not sit right to me, especially because the other two routes for `testFullVerificationWithCANamesFromCertificate` and `testFullVerificationWithCANamesFromFile` do not end up this way.  To work around this I shimmed in `clientConfig.trustRoots = .certificates([])` only in the test case for `testFullVerificationWithCANamesFromDirectory`.  @Lukasa if you have any other thoughts here on this matter, let me know.

Sources/NIOSSL/SSLContext.swift
agnosticdev: Added in the latest commit.

Sources/NIOSSL/SSLContext.swift
agnosticdev: Done.

Sources/NIOSSL/SSLContext.swift
Lukasa: Sorry, can you clarify the context where we get an unexpected failure?

agnosticdev: @Lukasa sure, the failure existed in the test case for `testFullVerificationWithCANamesFromDirectory`.
If a machine running the test case has the full certificate chain installed on the machine, then the test case will receive a failure for `TLSV1_ALERT_UNKNOWN_CA`.  Now, I suspect that this is only the case for self signed certificates, as that is what I am using here. The strange part about this is that I didn't see this failure with any of the other test cases that test for CA names.  I fixed the failure in the test case, but just wanted to see if there was anything that stuck out to you here.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -614,3 +656,35 @@ extension Optional where Wrapped == String {
...
+    
+    init(path: String) {
+        self.path = path
+        self.dir = opendir(path)
+    }
+    
+    func next() -> String? {
+        if let dirent: UnsafeMutablePointer<dirent> = readdir(self.dir) {
+            let name = withUnsafePointer(to: &dirent.pointee.d_name) { (ptr) -> String in
+                return ptr.withMemoryRebound(to: CChar.self, capacity: Int(ptr.pointee.0) + 1) { String(cString: $0) }
```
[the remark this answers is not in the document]
agnosticdev: Good call

Tests/NIOSSLTests/TLSConfigurationTest.swift
```diff
@@ -464,6 +525,220 @@ class TLSConfigurationTest: XCTestCase {
...
+        
+        // Pulling in the logic for assertHandshakeSucceededEventLoop because there needs to be some custom handling
+        // due to the IP address and the potential for `CERTIFICATE_VERIFY_FAILED`.
+        var clientContext: NIOSSLContext!
+        var serverContext: NIOSSLContext!
+        
+        XCTAssertNoThrow(clientContext = try NIOSSLContext(configuration: clientConfig))
+        XCTAssertNoThrow(serverContext = try NIOSSLContext(configuration: serverConfig))
+
+        let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
```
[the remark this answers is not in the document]
agnosticdev: The reason I am doing this here is because I am using the contents of `assertHandshakeSucceededEventLoop` as a template to account for `NIOSSLError`'s that come through here on the client channel in `finishHandshakeForCANameTests`.

Tests/NIOSSLTests/TLSConfigurationTest.swift
```diff
@@ -464,6 +525,220 @@ class TLSConfigurationTest: XCTestCase {
...
+    
+    func testFullVerificationWithCANamesFromCertificate() throws {
+        // Custom certificates for TLS and client authentication.
+        let root = try NIOSSLCertificate(bytes: .init(customCARoot.utf8), format: .pem)
+        
+        let leaf = try NIOSSLCertificate(bytes: .init(leafCertificateForTLSIssuedFromCustomCARoot.utf8), format: .pem)
+        let leaf_privateKey = try NIOSSLPrivateKey.init(bytes: .init(privateKeyForLeafCertificate.utf8), format: .pem)
+        
+        let client_cert = try NIOSSLCertificate(bytes: .init(leafCertificateForClientAuthenticationIssuedFromCustomCARoot.utf8), format: .pem)
+        let client_privateKey = try NIOSSLPrivateKey.init(bytes: .init(privateKeyForClientAuthentication.utf8), format: .pem)
```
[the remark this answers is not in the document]
agnosticdev: Done.

Tests/NIOSSLTests/TLSConfigurationTest.swift
```diff
@@ -254,6 +254,67 @@ class TLSConfigurationTest: XCTestCase {
...
+                // If this test is run on a remote system where `customCARoot` is not in the trust store or something like /etc/ssl/cert.pem
+                // then this handshake will fail with `CERTIFICATE_VERIFY_FAILED` because of the chain of trust cannot be established.
+                XCTAssertTrue(eventHandler.errors.description.contains(expectedError))
+                
+            }
+            XCTAssertEqual(handshakeHandler.handshakeSucceeded, handshakeSucceed)
+        }
+        try clientChannel.closeFuture.wait()
+        // If we have made it here, return true
+        return true
```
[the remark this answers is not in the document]
agnosticdev: Good call, I can clean this up.

agnosticdev:
@Lukasa addressed feedback.

Tests/NIOSSLTests/TLSConfigurationTest.swift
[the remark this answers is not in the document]
agnosticdev: I looked into this deeper to see if I could ditch `MultiThreadedEventLoopGroup` for the CA name unit tests; the stadard function for `assertHandshakeSucceeded` would not work so I assumed the contents of  `assertHandshakeSucceededEventLoop` two of the test functions and this did work well.

Tests/NIOSSLTests/TLSConfigurationTest.swift
```diff
@@ -464,6 +473,247 @@ class TLSConfigurationTest: XCTestCase {
...
+        
+        XCTAssertNoThrow(try serverChannel.pipeline.addHandler(NIOSSLServerHandler(context: serverContext)).wait())
+        XCTAssertNoThrow(try clientChannel.pipeline.addHandler(NIOSSLClientHandler(context: clientContext, serverHostname: nil)).wait())
+        let handshakeHandler = HandshakeCompletedHandler()
+        XCTAssertNoThrow(try clientChannel.pipeline.addHandler(handshakeHandler).wait())
+
+        // Connect. This should lead to a failed handshake.
+        let addr = try assertNoThrowWithValue(SocketAddress(unixDomainSocketPath: "/tmp/whatever2"))
+        let connectFuture = clientChannel.connect(to: addr)
+        serverChannel.pipeline.fireChannelActive()
```
Lukasa, line 523: I think we have a `connectInMemory` function that can do this dance for you.

Tests/NIOSSLTests/TLSConfigurationTest.swift
```diff
@@ -464,6 +473,247 @@ class TLSConfigurationTest: XCTestCase {
...
+        
+        XCTAssertNoThrow(try serverChannel.pipeline.addHandler(NIOSSLServerHandler(context: serverContext)).wait())
+        XCTAssertNoThrow(try clientChannel.pipeline.addHandler(NIOSSLClientHandler(context: clientContext, serverHostname: nil)).wait())
+        let handshakeHandler = HandshakeCompletedHandler()
+        XCTAssertNoThrow(try clientChannel.pipeline.addHandler(handshakeHandler).wait())
+
+        // Connect. This should lead to a completed handshake.
+        let addr = try assertNoThrowWithValue(SocketAddress(unixDomainSocketPath: "/tmp/whatever2"))
+        let connectFuture = clientChannel.connect(to: addr)
+        serverChannel.pipeline.fireChannelActive()
```
Lukasa, line 580: Same re `connectInMemory` here.

Tests/NIOSSLTests/TLSConfigurationTest.swift
```diff
@@ -254,6 +254,15 @@ class TLSConfigurationTest: XCTestCase {
         return try assertHandshakeSucceededEventLoop(withClientConfig: clientConfig, andServerConfig: serverConfig, file: file, line: line)
         #endif
     }
+    
+    func setupTLSLeafandClientIdentitiesFromCustomCARoot() throws -> (NIOSSLCertificate, NIOSSLPrivateKey , NIOSSLCertificate, NIOSSLPrivateKey) {
```
Lukasa, line 258: As a minor test cleanup, let's use labels for these fields in the tuple so that we can access them by name in the test cases instead of as `.0` and `.1` and so-on. Additionally, in the test cases can you give this tuple a camelCase name?

Lukasa (commented):
Two minor notes, otherwise seems great.

Tests/NIOSSLTests/TLSConfigurationTest.swift
agnosticdev: Same with above.

Tests/NIOSSLTests/TLSConfigurationTest.swift
agnosticdev: @Lukasa this has been addressed in the latest commit.

Lukasa (approved):
LGTM, let's ship it.

Lukasa:
Nightly build timed out building the docker image. Will re-spin.

Lukasa:

Lukasa:
Nightly failed due to #317, not introduced in this PR so we can merge this. Thanks for your work here @agnosticdev!

agnosticdev:
Thank you @Lukasa !

PR #308, opened 2021-08-16
Update BoringSSL to 2e68a05c9943a8dec1758d4a393b2ae906fd3295

Lukasa:
Also tweak the vendor script to add an explicit destination when
building for iOS.


fabianfett (approved):
Just out of curiosity: Why do we now need the `-destination` flag and haven't so far.

Lukasa:
xcodebuild in Xcode 13 got mad at me when I didn't have it, and it never used to. That's the best reason I can give I'm afraid :wink:

PR #311, opened 2021-09-02
Implement tlsext_sni_cb

### LINKED ISSUE #310: Certificate selection for servers is missing
`SSL_CTX_set_tlsext_servername_callback` is not implemented.

To write a modern TLS enabled server, this callback is required. Why? Because a modern server should be capable of being initialized with a default `SSL_CTX` that implements the SNI callback and, if executed (when SNI is present in the client hello), you should be able to fetch or generate a `SSL_CTX` specifically for that host (fetched with `SSL_get_servername(ssl, TLSEXT_NAMETYPE_host_name)`) and then also dynamically change the context in use for that connected client `SSL` object via `SSL_set_SSL_CTX()` before returning from the callback.

Without this, you can only write servers like its 2002 (SNI was introduced in 2003).

I'll see if I can do a PR but it may not come soon. It should be trivial to implement for someone who already knows their way around this library.

TechnikEmpire:
Created the ability to initialize a TLSConfiguration + NIOSSLContext specifically for servers that want to respond to the TLS SNI extension in client hello messages and dynamically create/load a NIOSSLContext per-host.

Fixes #310


TechnikEmpire:
@swift-server-bot Why you spamming bro calm down.

Lukasa:
@TechnikEmpire Sorry about the notification spam, just ignore it. :wink:

Sources/NIOSSL/SSLCallbacks.swift
```diff
@@ -101,6 +100,33 @@ public typealias NIOSSLCustomVerificationCallback = ([NIOSSLCertificate], EventL
...
+///
+/// When set, this callback will be invoked once per TLS hello. The provided `String` will contain the
+/// host name indicated in the TLS client hello, while the provided `NIOSSLContext` will be the current
+/// server context being used for the handshake.
+///
+/// Within this callback, the user can create and return a new server `NIOSSLContext` for the given host,
+/// return the same context provided if it is sufficient to complete the handshake, or return `nil` in the event
+/// of an exception.
+///
+public typealias NIOSSLSniCallback = (String, NIOSSLContext) -> NIOSSLContext?
```
Lukasa, line 113: I'd like to litigate this interface a bit, if we can.

Concretely, this API is not actually restricted to SNI: the underlying API name is not really reflective of what it does. The API itself is called after the client extensions are parsed, and so can be used to modify the context based on any of them. As a result, we should consider this to be a bit more general.

Instead of having this pass a `String` as the first argument, we should probably create a new `struct` that contains the values of some extensions that we want to expose. This will let us evolve the API cleanly by adding new fields to that structure as needed.

For now we can stick to server name, but I'd like to have the hook available. For now this should probably be:

```swift
public struct NIOSSLClientExtensionValues: Hashable {
    public var serverName: String
    
    public init(serverName: String) {
        self.serverName = serverName
    }
}
```

We should also extend the interface to allow the function to throw, probably instead of having it return `nil`.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -411,7 +418,7 @@ extension NIOSSLContext {
             throw BoringSSLError.failedToSetALPN(errorStack)
         }
     }
-
+
```
Lukasa, line 421: Nit: can we remove this whitespace?

Sources/NIOSSL/SSLContext.swift
```diff
@@ -520,6 +527,83 @@ extension NIOSSLContext {
...
+            {
+                return SSL_TLSEXT_ERR_NOACK
+            }
+            
+            // We want to take the SSL pointer and extract the parent Swift object. These force-unwraps are for
+            // safety: a correct NIO program can never fail to set these pointers, and if it does failing loudly is
+            // more useful than failing quietly.
+            let parentCtx = CNIOBoringSSL_SSL_get_SSL_CTX(ssl)!
+            let parentPtr = CNIOBoringSSLShims_SSL_CTX_get_app_data(parentCtx)!
+            let parentSwiftContext: NIOSSLContext = Unmanaged.fromOpaque(parentPtr).takeUnretainedValue()
```
Lukasa, line 559: This pattern is common enough that I think we should probably extract it into a helper function now.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -520,6 +527,83 @@ extension NIOSSLContext {
...
+            // However, SSL_set_SSL_CTX explicitly increases the reference count and
+            // will decrement the count when the SSL* object is destroyed, thus finally
+            // freeing the raw SSL_CTX object, even if the swift reference is orphaned.
+            //
+            // As such, no memory should be leaked or freed too early in the event that
+            // the user fails to retain their created and return NIOSSLContext. In the
+            // event that they do retain it, the reference count will remain at at-least
+            // 1 and thus they can re-use the same context for multiple connections.
+            
+            guard let nativeReturnVal = CNIOBoringSSL_SSL_set_SSL_CTX(ssl, userChosenContext.sslContext) else
```
Lukasa, line 581: The above comment is great, but the code as written is not quite right in the face of refactors. If we stopped using `userChosenContext` below, then it would be allowable for Swift to free the `userChosenContext` after extracting the `sslContext` pointer from it, causing `deinit` to run between the `get` and the call to `SSL_set_SSL_CTX`.

To that end, we should wrap this in a call to `withExtendedLifetime` to make sure things behave properly:

```swift
let nativeReturnVal = withExtendedLifetime(userChosenContext) {
    CNIOBoringSSL_SSL_set_SSL_CTX(ssl, userChosenContext.sslContext)
}
```

That way refactors will never affect the correctness of this code.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -520,6 +527,83 @@ extension NIOSSLContext {
...
+            {
+                return SSL_TLSEXT_ERR_NOACK
+            }
+            
+            // If no action was taken, we indicate that this current context can be used
+            // to attempt to complete the handshake. Given that the user has indicated
+            // that they want control of setting the context when SNI is indicated, if this
+            // indication turns out to be false, the SSL/TLS connection will fail with an
+            // appropriate error and the user will be made aware of it.
+            return SSL_TLSEXT_ERR_OK
```
Lukasa, line 599: I don't think we're quite done here. The `SSLConnection` object itself has a reference to its parent context, which is now different. We need to update that. We also need to update any state derived from it, such as the call to `setRenegotiationSupport` in `SSLConnection.init`. It would be good to add unit tests that validate that any change to that field is correctly reflected when we change the context.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -438,6 +454,33 @@ extension TLSConfiguration {
...
+                                additionalTrustRoots: [])
+    }
+    
+    /// Create a TLS configuration for dynamically generating server-size contexts with SNI.
+    ///
+    /// This provides sensible defaults while requiring that you provide any data that is necessary
+    /// for server-side function. For client use, try `makeClientConfiguration` instead.
+    ///
+    /// For customising fields, modify the returned TLSConfiguration object.
+    public static func makeServerConfiguration(
```
Lukasa, line 468: I don't think we need this function: users can simply call the main version and set the SNI callback as desired. The advantage of the main callback is that it encourages users to set "fallback" certs and keys.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -463,6 +506,7 @@ extension TLSConfiguration {
                                  applicationProtocols: [String] = [],
                                  shutdownTimeout: TimeAmount = .seconds(5),
                                  keyLogCallback: NIOSSLKeyLogCallback? = nil,
+                                 sniCallback: NIOSSLSniCallback? = nil,
```
Lukasa, line 509: We cannot add arguments to existing methods, defaulted or not, as it's a semver major.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -494,7 +539,8 @@ extension TLSConfiguration {
                                  trustRoots: NIOSSLTrustRoots = .default,
                                  applicationProtocols: [String] = [],
                                  shutdownTimeout: TimeAmount = .seconds(5),
-                                 keyLogCallback: NIOSSLKeyLogCallback? = nil) -> TLSConfiguration {
+                                 keyLogCallback: NIOSSLKeyLogCallback? = nil,
+                                 sniCallback: NIOSSLSniCallback? = nil) -> TLSConfiguration {
```
Lukasa, line 543: Same here.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -527,7 +574,8 @@ extension TLSConfiguration {
...
```
Lukasa, line 578: Same here.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -561,6 +610,7 @@ extension TLSConfiguration {
                                  applicationProtocols: [String] = [],
                                  shutdownTimeout: TimeAmount = .seconds(5),
                                  keyLogCallback: NIOSSLKeyLogCallback? = nil,
+                                 sniCallback: NIOSSLSniCallback? = nil,
```
Lukasa, line 613: Same here.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -595,6 +646,7 @@ extension TLSConfiguration {
...
```
Lukasa, line 649: Same here.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -627,7 +680,8 @@ extension TLSConfiguration {
                                  privateKey: NIOSSLPrivateKeySource? = nil,
                                  applicationProtocols: [String] = [],
                                  shutdownTimeout: TimeAmount = .seconds(5),
-                                 keyLogCallback: NIOSSLKeyLogCallback? = nil) -> TLSConfiguration {
+                                 keyLogCallback: NIOSSLKeyLogCallback? = nil,
+                                 sniCallback: NIOSSLSniCallback? = nil) -> TLSConfiguration {
```
Lukasa, line 684: Same here.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -660,6 +715,7 @@ extension TLSConfiguration {
                                  applicationProtocols: [String] = [],
                                  shutdownTimeout: TimeAmount = .seconds(5),
                                  keyLogCallback: NIOSSLKeyLogCallback? = nil,
+                                 sniCallback: NIOSSLSniCallback? = nil,
```
Lukasa, line 718: Same here.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -694,6 +751,7 @@ extension TLSConfiguration {
...
```
Lukasa, line 754: Same here.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -728,6 +787,7 @@ extension TLSConfiguration {
...
```
Lukasa, line 790: Same here.

Tests/NIOSSLTests/ServerSNITests.swift
```diff
@@ -0,0 +1,154 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
```
Lukasa, line 5:
```suggestion
// Copyright (c) 2021 Apple Inc. and the SwiftNIO project authors
```

Tests/NIOSSLTests/ServerSNITests.swift
```diff
@@ -0,0 +1,154 @@
...
+import NIO
+import NIOTLS
+import NIOSSL
+
+
+class ServerSNITests: XCTestCase {
+    
+    class SniServerContext {
+    
+        private var _tlsConfig : TLSConfiguration?;
```
Lukasa, line 25:
```suggestion
        private var _tlsConfig : TLSConfiguration?
```

Tests/NIOSSLTests/ServerSNITests.swift
```diff
@@ -0,0 +1,154 @@
...
+        {
+            get
+            {
+                return _ctx!
+            }
+        }
+        
+        private var _indicatedHostname : String?
+        
+        public var IndicatedHostname : String
```
Lukasa, line 41: Rather than have the getters/setters, should we just use `public private(set)` and expose their optionality?

Tests/NIOSSLTests/ServerSNITests.swift
```diff
@@ -0,0 +1,154 @@
...
+
+class ServerSNITests: XCTestCase {
+    
+    class SniServerContext {
+    
+        private var _tlsConfig : TLSConfiguration?;
+        
+        private var _ctx : NIOSSLContext?
+        
+        private(set) var SniExpectation : XCTestExpectation
```
Lukasa, line 29:
```suggestion
        private(set) var sniExpectation : XCTestExpectation
```

Tests/NIOSSLTests/ServerSNITests.swift
```diff
@@ -0,0 +1,154 @@
...
+            SniExpectation = sniExpectation
+        }
+     
+        private func sniCallback (host: String, ctx: NIOSSLContext) -> NIOSSLContext? {
+            self._indicatedHostname = host
+            self.SniExpectation.fulfill()
+            return ctx
+        }
+        
+        public static func Create(sniExpectation: XCTestExpectation) -> SniServerContext {
```
Lukasa, line 59: This can just be folded into `init`.

Tests/NIOSSLTests/ServerSNITests.swift
```diff
@@ -0,0 +1,154 @@
...
+        let expectation = expectation(description: "Notification Raised")
+        return SniServerContext.Create(sniExpectation: expectation)
+    }
+
+    private func assertSniResult(sniField: String?, expectedResult: String) throws {
+        let clientContext = try configuredClientSSLContext()
+
+        let serverContext = try configuredServerSSLContext()
+        
+        let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
```
Lukasa, line 100: Can we try to run these using `EmbeddedChannel` and the `connectInMemory`/`interactInMemory` functions defined in other test helper files? This reduces FD pressure in CI machines and also makes the tests vastly more reproducible, as well as faster.

Tests/NIOSSLTests/ServerSNITests.swift
```diff
@@ -0,0 +1,154 @@
...
+        
+        XCTAssertEqual(sniResult, expectedResult)
+    }
+
+    func testSNIIsTransmitted() throws {
+        try assertSniResult(sniField: "httpbin.org", expectedResult: "httpbin.org")
+    }
+
+    func testNoSNILeadsToNoSNI() throws {
+        try assertSniResult(sniField: nil, expectedResult: String())
```
Lukasa, line 141: Nit: 

```suggestion
        try assertSniResult(sniField: nil, expectedResult: "")
```

Tests/NIOSSLTests/ServerSNITests.swift
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    func testSNIIsNotRejectedForAnyOfTheFirst1000CodeUnits() throws {
+        for testString in (1...Int(1000)).compactMap({ UnicodeScalar($0).map { String($0) } }) {
```
Lukasa, line 145:
```suggestion
        for testString in (1...Int(1000)).map({ String($0, radix: 10)! }) {
```

Tests/NIOSSLTests/TLSConfigurationTest.swift
```diff
@@ -542,9 +542,9 @@ class TLSConfigurationTest: XCTestCase {
         var differentConfig = config
         differentConfig.privateKey = .file("fake2.file")
         XCTAssertFalse(config.bestEffortEquals(differentConfig))
-    }
+    }
```
Lukasa, line 545:
```suggestion
    }
```

Lukasa (changes requested):
Thanks for this patch! It's extremely thorough and I think it's in really good shape.

I've left a number of comments in the diff. Many of these are simply nits or style notes that I'd like to see cleaned up. The biggest suggestions are to make the interface more general, and to fix up some internal consistency issues that we can get ourselves into here.

To add to those I'd like to add some more testing that confirms that changes to the SSLContext itself are actually reflected in the underlying connection. That is, can we confirm that we get the certificates we expect if we tweak the connection, or that we get the expected ALPN protocols, and so on? I identified a specific one that I think this testing would have revealed, but I think it'd be good to get a few more tests to cover those cases.

Sources/NIOSSL/SSLCallbacks.swift
TechnikEmpire: I appreciate the vision to try to make this into a more general interface, but that goes explicitly against the way in which the native library (boring/open/ssl) actually uses these callbacks. This callback is executed at a time *before* the client hello is fully processed and there are multiple callbacks that are called in order and *must* be called in order explicitly for different purposes. From the [docs](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_tlsext_servername_callback.html):

```
NOTES

Several callbacks are executed during ClientHello processing, including the ClientHello, ALPN, and servername callbacks. The ClientHello callback is executed first, then the servername callback, followed by the ALPN callback.
```

With respect to throwing, I believe I left comments elsewhere that failures are propagated elsewhere automatically on failure and those errors manifest in a more correct fashion than a hard throw. Failures arising from an incorrect configuration created in this callback will cause the client to fail to connect at the handshake stage.

Sources/NIOSSL/SSLContext.swift
TechnikEmpire: Agreed. I copy and pasted existing code.

Sources/NIOSSL/SSLContext.swift
TechnikEmpire: I'll take a look at this. I've never used swift before and I'm trying to imagine how the destructor could be invoked in the same line the object still exists and is being accessed. We don't care at all to retain `nativeReturnResult` and SSL_set_SSL_CTX is a blocking call that immediately runs an atomic increment of the reference count stored in the native object, which will prevent its freeing even if the `userChosenContext` `deinit` is executed.

Sources/NIOSSL/SSLContext.swift
TechnikEmpire: Good point regarding updating the native state object. However, I don't want to get too crazy with testing here beyond making sure the linked swift objects change because then we're venturing into testing the correctness of the native crypto library itself. 

To resolve any possible confusion, what's actually happening here is we're changing the `SSL_CTX` for **this single client connection alone**, represented by the `SSL*` parameter that BoringSSL is feeding us in this callback.

The original server `SSLContext`, the one bound to the `TlsConfiguration` that has the SNI callback set on it, never goes away or changes. It remains as the default context for the stream wrapping the TCP listener on the server.

Something like this pseudo:

```
defaultSslCtx = new sslctx + sni callback
let clientConn = new sslstream<clientTcpConn, defaultSslCtx>() from tcpListener.accept
clientConn.handshake
snicallback invoked, SSL* == clientConn
create brand new context with entirely tailored settings, crafted for the given host
clientConn.context = brand new context // original context remains in use for all new client connections
```

So yeah I think for sure it's a great call to verify that the relationships between the swift objects and the native objects are being correctly maintained. But if we get into doing things like verifying that BoringSSL has actually done its job then I think we're leaving our boundary and overlapping into testing the native library itself.

Sources/NIOSSL/SSLCallbacks.swift
Lukasa: OpenSSL's docs don't bind us here, as we aren't using OpenSSL. :smile: Additionally, while the underlying library has a callback series, I'd rather that we didn't. In my view there are only two sensible places to run a callback: as soon as possible after the extensions have been parsed, or before any processing is done at all.

The early callback is set by `SSL_CTX_set_select_certificate_cb`. This is a neat callback. It lets us do things that we can't do with any other, such as change the proposed version range. However, due to its awkwardness we probably don't want to use it right now. Then there are a number of others. All of these others are used in a specific order. However, all of them are handled _after_ the extensions are parsed.

The next best callback to use is the one I'd like to propose we use: `SSL_CTX_set_cert_cb`. This sets a callback that is called after the extensions are processed, but before a resumption decision is made. In fact, it runs immediately after the callback you're attaching here. However, the `SSL_CTX_set_cert_cb` callback has one specific advantage: it can be asynchronous. That is, we can return a value that asks for the handshake to be paused, and then resume it later. This gives us some real utility, as we can now do asynchronous work in order to load certs and keys.

To that end, I'd like to propose we use `SSL_CTX_set_cert_cb`, and consider it the primary customisation point. There is only one thing that this callback cannot do that the one we're using here can, which is to explicitly NACK the SNI extension. I don't think that's valuable, and I think it's substantially _less_ valuable than having the asynchronous support, so I'd like to push in that direction.

Sources/NIOSSL/SSLContext.swift
Lukasa: The way to envision it is this: Swift treats all `get` accessors _as if_ they performed a shallow copy of the element. Essentially, `var tempBar = foo.bar` is conceptually equivalent to `var tempBar = foo.getBar()`. Consider, then, this code: `someFunction(foo.bar)`. This is conceptually equivalent to:

```swift
var temporaryBar = foo.getBar()
someFunction(temporaryBar)
```

Here it should be clear why there might be a risk: after `getBar` completes there is no longer a reference to `foo`, and so we can deinit it. This is not a hypothetical: Swift has over time made the ARC checker increasingly aggressive, and I have personally observed it deinit an object in this way.

Sources/NIOSSL/SSLContext.swift
Lukasa: :+1: I think that's entirely reasonable here. Because of the limited type visibility I think the best way to do this is to use the only thing we know changes, which is the renegotiation support. If we add a single test to validate this, say by twiddling from a context that _does_ support renegotiation to one that doesn't, it should be sufficient to prove we've done the right thing I think.

Sources/NIOSSL/SSLCallbacks.swift
TechnikEmpire: Ok so you want me to change this whole callback interface to use SSL_CTX_set_cert_cb? Or you want me to do this generalization so that these other callbacks can be implemented in the future?

Also I was linking in ossl docs cause I figured this has not changed in fork but I will double check.

I'm trying to get this basic feature in asap to pitch using this lib at work so I would prefer to take up any expansion on this capability in subsequent patches, just so I can pitch this with a POC. Let me know your thoughts.

Sources/NIOSSL/SSLContext.swift
TechnikEmpire: Got it. Thanks.

Sources/NIOSSL/SSLCallbacks.swift
Lukasa: I'd like to change the whole callback interface to `SSL_CTX_set_cert_cb`.

In general I'm fine with the idea of following up in subsequent patches, but it puts this library in a tricky spot if we merge partial work, because we then cannot easily tag releases. My suggestion for rapid pitching at work would be to rely on your fork in your PoC, which can contain incomplete work, while we work through this patch. Does that sound reasonable?

TechnikEmpire: Ok, I'm going to have to basically redo all of this if we're switching the interface. IMO we're actually restricting the interface going this way and adding work for no reason. 

By switching to the cert callback, then we're moving away from being able to return fully customized NIOSSLContext objects on a per-connection basis, to having to use the same context for every connection and then we're basically going to reimplement the whole purpose of a NIOSSLContext by expanding a new struct over time that will force users to making BoringSSL calls on raw pointers or worse yet, force us to keep wrapping those calls in the library.

For clarity:

`public typealias NIOSSLSniCallback = (String, NIOSSLContext) -> NIOSSLContext?`

becomes

`public typealias NIOSSLCertCallback = (EverExpandingStructOfThingsInsideNIOSSLContext, SSL*) -> void`

and then we're either forcing users to abuse a certificate select callback to do a bunch of stuff, or we're creating a new callback for everything you can do inside a `NIOSSLContext` and forcing them to have a bunch of useless/irrelevant information inside of `EverExpandingStructOfThingsInsideNIOSSLContext`.

I appreciate that you can pause the handshake with `SSL_CTX_set_cert_cb` and resume it and then we can tell people it's async. I've been writing multithreaded SSL/TLS servers for a decade and have never once needed so badly to spare a few microseconds of CPU cycles that I had to pause the handshake and come back to it later.

Anyway I don't have time to basically delete everything I did and start over so I think we can consider this PR dead.

Lukasa: > Anyway I don't have time to basically delete everything I did and start over so I think we can consider this PR dead.

I'm sorry to hear that. 😞 

But, for what it's worth, I said only that we should change the _first_ argument from `String` to a `struct`. That is, the change would be from:

```swift
public typealias NIOSSLSniCallback = (String, NIOSSLContext) -> NIOSSLContext?
```

to

```swift
public typealias NIOSSLSniCallback = (NIOSSLClientExtensionValues, NIOSSLContext) -> EventLoopFuture<NIOSSLContext>
```

TechnikEmpire: @Lukasa Yes but it seems that there's a misunderstanding about what I'm doing here. If I go and implement your change, then there's no sense in receiving or returning  a `NIOSSLContext`, because the context *will never change*. The context will never change because we're now using `SSL_CTX_set_cert_cb`, which is explicitly for modifying *only the certificate being used in the handshake*.

The purpose of this callback is to modify the cert and we're abusing the API to make changes to the `SSL` connection, changes we will expand on later by appending to the new struct you want to create. What you're doing is manually reimplementing the sole purpose of `SSL_CTX`, which we already have. But instead of using it, we're going to reuse the same context for every server connection and manually apply changes on the `SSL` struct.

So while `SSL_CTX_set_cert_cb` might be "asynchronous" (not really, we're just suspending and resuming within the same IO reactor and pretending its aynchronous), this change is 1. A massive deviation from the intent of the pull request 2. API abuse 3. Code duplication.

Lukasa: None of the above is my proposal. `SSL_CTX_set_cert_cb` is explicitly given the permission to call `SSL_set_SSL_CTX`, which I am proposing we continue to do. We are using only a newly-added callback that allows for asynchronous loading of contexts: otherwise its behaviour is identical.

And, to be clear, it is absolutely asynchronous: we can pause the handshake indefinitely until we have a context we want to use.

TechnikEmpire: Thanks @Lukasa, I respectfully disagree. Anyway I've just found a bunch of other inadequacies that have now forced my hand to maintain an incompatible fork. Thanks for your time on this.

TechnikEmpire:
Closing as too many additional changes necessary to make this into a viable product and too much work to try to back port or integrate.

Sources/NIOSSL/SSLCallbacks.swift
Lukasa: You're welcome and thanks for your contribution. Please don't hesitate to reach out if you have further questions or changes.

PR #312, opened 2021-09-03
Support building in the Swift toolchain

Lukasa:
Motivation:

The Swift toolchain has an unusual build environment that attempts to
check out all package dependencies locally before building begins. To
support that flow we need to detect that we're in the toolchain
environment and defer to a local checkout accordingly.

Modifications:

- Added a check for building in the toolchain.

Result:

Better builds in the toolchain.


Davidde94 (approved):

glbrntt:
FYI you may need https://github.com/apple/swift-nio/pull/1950 (pushing this around the various repos is on my todo list)

fabianfett (approved):

PR #315, opened 2021-09-08
Update BoringSSL to 25773430c07075a368416c3646fa4b07daf4968a

Lukasa:
Motivation:

BoringSSL has begun opaquifying their X.509 library. This unfortunately
manifests as source code changes in Swift due to the way OpaquePointer
behaves. As a result, we need to change the places that held an X.509
pointer directly.

While we're here, we've been playing a bit fast-and-loose with the
lifetime of owned pointers. We should take this opportunity to clean
things up and be a bit better behaved.

Modifications:

- Replace the X509 pointers with OpaquePointer
- Use scoped access to the X.509 access outside the NIOSSLCertificate
  class.

Result:

Code compiles with better pointer hygeine.


glbrntt (approved):

glbrntt (changes requested):
Oops -- approved in wrong tab

Lukasa:
API breakage failed but it seems transient.

Lukasa:

Davidde94 (approved):

glbrntt (approved):
Looks good.

Lukasa:
Nightly is failing due to #317, merging over the failure.

PR #319, opened 2021-09-14
Clean up imports and dependencies

Lukasa:
Motivation:

With NIO 2.32.0 we broke the core NIO module up into modules that split
apart the POSIX layer and the core abstractions. As a result, this
package no longer needs to express a hard dependency on the POSIX layer.

Modifications:

- Rewrote imports of NIO to NIOCore.
- Added NIOEmbedded and NIOPosix imports where necessary in tests.
- Extended soundness script to detect NIO imports.
- Note that the main modules still depend on NIO, which is necessary
for backwards-compatibility reasons. This dependency is unused.

Result:

No need to use NIOPosix.

### DIFF
--- a/Sources/NIOSSL/IdentityVerification.swift
+++ b/Sources/NIOSSL/IdentityVerification.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftNIO open source project
 //
-// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
+// Copyright (c) 2017-2021 Apple Inc. and the SwiftNIO project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
@@ -12,7 +12,15 @@
 //
 //===----------------------------------------------------------------------===//
 
-import NIO
+import NIOCore
+
+#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+import Darwin.C
+#elseif os(Linux) || os(FreeBSD) || os(Android)
+import Glibc
+#else
+let badOS = { fatalError("unsupported OS") }()
+#endif
 
 
 private let asciiIDNAIdentifier: ArraySlice<UInt8> = Array("xn--".utf8)[...]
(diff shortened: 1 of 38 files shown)


Sources/NIOSSL/IdentityVerification.swift
```diff
@@ -12,7 +12,15 @@
...
 
-import NIO
+import NIOCore
+
+#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+import Darwin.C
+#elseif os(Linux) || os(FreeBSD) || os(Android)
+import Glibc
+#else
+let badOS = { fatalError("unsupported OS") }()
```
Davidde94, line 22: Can't we do #error ?

Sources/NIOSSL/NIOSSLClientHandler.swift
```diff
@@ -12,7 +12,15 @@
...
 
-import NIO
+import NIOCore
+
+#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+import Darwin.C
+#elseif os(Linux) || os(FreeBSD) || os(Android)
+import Glibc
+#else
+let badOS = { fatalError("unsupported OS") }()
```
Davidde94, line 22: Same

Sources/NIOSSL/PosixPort.swift
```diff
@@ -20,7 +20,15 @@
 //
 // The code is an exact port from SwiftNIO, so if that version ever becomes public we
 // can lift anything missing from there and move it over without change.
-import NIO
+#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+import Darwin.C
+#elseif os(Linux) || os(FreeBSD) || os(Android)
+import Glibc
+#else
+let badOS = { fatalError("unsupported OS") }()
```
Davidde94, line 28: Same, I won't comment if I see it again

Davidde94 (approved):

fabianfett (approved):
Great fix!

PR #320, opened 2021-09-16
Fixup CA name tests on Darwin

Lukasa:
Motivation:

The recently-added CA name tests consistently fail on Darwin machines.
This is becuase they expect the handshake to fail due to certificate
verification errors, but that doesn't happen synchronously on Darwin.
This also introduces thread-unsafety into the tests. Neither is good.

Modifications:

- Modify the test configuration to use custom trust roots, which
  disables the Security.framework callbacks
- Modify the test configuration to disable hostname verification.

Result:

These tests now successfully perform their handshake on all platforms.

### DIFF
--- a/Tests/NIOSSLTests/TLSConfigurationTest.swift
+++ b/Tests/NIOSSLTests/TLSConfigurationTest.swift
@@ -484,6 +484,11 @@ class TLSConfigurationTest: XCTestCase {
         let digitalIdentities = try setupTLSLeafandClientIdentitiesFromCustomCARoot()
         
         // Client Configuration.
+        //
+        // This configuration disables hostname verification because the hostname verification
+        // code requires IP addresses, which we don't have in EmbeddedChannel. We override the
+        // trust roots to prevent execution of the SecurityFramework verification code, which doesn't
+        // work with EmbeddedChannel.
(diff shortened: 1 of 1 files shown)


fabianfett (approved):
LGTM

glbrntt (approved):
LGTM -- thanks for fixing this!

Tests/NIOSSLTests/TLSConfigurationTest.swift
```diff
@@ -480,19 +480,31 @@ class TLSConfigurationTest: XCTestCase {
...
+        //
+        // This configuration disables hostname verification because the hostname verification
+        // code requires IP addresses, which we don't have in EmbeddedChannel. We override the
+        // trust roots to prevent execution of the SecurityFramework verification code, which doesn't
+        // work with EmbeddedChannel.
         var clientConfig = TLSConfiguration.makeClientConfiguration()
         clientConfig.renegotiationSupport = .none
         clientConfig.certificateChain = [.certificate(digitalIdentities.clientCert)]
         clientConfig.privateKey = .privateKey(digitalIdentities.clientKey)
+        clientConfig.trustRoots = .certificates([root])
```
agnosticdev, line 492: Good call adding the custom root certificate on the client configuration.

PR #321, opened 2021-09-16
Remove the implementationOnly dance

Lukasa:
Motivation:

When we added implementationOnly imports a while back, we supported
Swift 5.0, and so we needed to guard their availability behind a
compiler check. We only support 5.2 and later now, so that check is
redundant and can go away.

Modifications:

- Removed compiler checks around imports of our C libraries.

Result:

Cleaner code

### DIFF
--- a/Sources/NIOSSL/ByteBufferBIO.swift
+++ b/Sources/NIOSSL/ByteBufferBIO.swift
@@ -13,11 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 import NIOCore
-#if compiler(>=5.1)
 @_implementationOnly import CNIOBoringSSL
-#else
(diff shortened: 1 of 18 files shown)


fabianfett (approved):

fabianfett:
Do we maybe want to test in our soundness script that CNIOBoringSSL is always imported `@_implementationOnly`?

Lukasa:
We don't need to: the Swift compiler will warn if we're inconsistent, and that'll produce an error in CI.

glbrntt (approved):
LGTM

fabianfett:
> We don't need to: the Swift compiler will warn if we're inconsistent, and that'll produce an error in CI.

Right! I forgot!

PR #322, opened 2021-09-20
Add callback-based custom private keys.

Lukasa:
Motivation:

NIOSSL traditionally handles private keys be requiring that they can be
decoded into a BoringSSL EVP_PKEY structure. This requires the key to
have an in-memory representation that BoringSSL understands, and
ultimately forces the bytes of the private key into memory.

In many cases, however, this is either undesirable or not possible.
For security reasons it is common to want to offload a private key into
secure storage, such as the iPhone's Secure Enclave or a portable smart
card. In other circumstances it may be useful to put the private key on
an entirely different machine and have the signing operations occur over
an RPC mechanism.

In all of these cases it is necessary to have a programmatic interface
to the private key that can be implemented in an essentially arbitrary
manner.

Modifications:

- Defined a new protocol, `NIOSSLCustomPrivateKey`, that allows users to
  implement the private key operations needed for TLS.
- Plumbed support for this interface through `NIOSSLPrivateKey`.
- Added support for the BoringSSL interface to `SSLConnection`.

Result:

Custom TLS private keys backed by function calls can now be provided to
NIOSSL.


Sources/NIOSSL/CustomPrivateKey.swift
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    @inlinable func hash(into hasher: inout Hasher) {
+        self._hashFunction(&hasher)
+    }
+
+    @inlinable static func ==(lhs: AnyNIOSSLCustomPrivateKey, rhs: AnyNIOSSLCustomPrivateKey) -> Bool {
+        return lhs._equalsFunction(rhs._value)
+    }
+}
```
fabianfett, line 111: If we only use it in tests, could this live in the test target?

Lukasa: Ah, no: the `Hashable` requirement is on the class that holds this. What I mean is that NIOSSL itself doesn't use the conformance, but we added it as API and can't take it away.

Sources/NIOSSL/SSLPrivateKey.swift
```diff
@@ -266,6 +277,18 @@ public class NIOSSLPrivateKey {
...
+    /// Create a `NIOSSLPrivateKey` from a custom private key callback.
+    ///
+    /// The private key, in addition to needing to conform to `NIOSSLCustomPrivateKey`,
+    /// is also required to be `Hashable`. This is because `NIOSSLPrivateKey`s are `Hashable`.
+    ///
+    /// - parameters:
+    ///     - customPrivateKey: The custom private key to use with the TLS certificate.
+    @inlinable
+    public init<CustomKey: NIOSSLCustomPrivateKey & Hashable>(customPrivateKey: CustomKey) {
+        self.representation = .custom(AnyNIOSSLCustomPrivateKey(customPrivateKey))
```
agnosticdev, line 289: Just a question; would this be the entry point for any key that is created outside the context of BoringSSL's `EVP_PKEY` type?

Lukasa: Yup

Sources/NIOSSL/CustomPrivateKey.swift
fabianfett: Is there a benefit to use `@inlinable` if we operate on an existential anyway?

Lukasa: Not really, no. For `hash` and `==` it's _moderately_ useful simply to hoist the logic a bit, but realistically it's not. The `@inlinable`s are a holdover from a previous design. I can remove them if you think it would improve things.

fabianfett: I think, I'd prefer if we only used `@inlinable`, if we can actually assume a performance win.

fabianfett (approved):
Looks good to my very untrained eye.

PR #323, opened 2021-09-21
Tolerate zeroReturnError in closed/unwrapped

Lukasa:
Motivation:

In some rare cases we've seen production applications crash due to us
getting a zeroReturnError in either closed or perhaps unwrapped. This is
pretty hard to get to trigger in realistic situations, but I've found a
contrived-but-possible way to hit it in tests. In both of these cases we
are able to simply ignore the error and move on.

Modifications:

- Added a test to replicate the issue
- Tolerate .unwrapped and .closed on zeroReturnError.

Result:

Fewer weird crashes


fabianfett (approved):
LGTM

Issue #316, opened 2021-09-08
OpenSSL Directory Format


agnosticdev:
Add functionality so that certificates can be read and loaded from the file system based on the hashed subject name of the certificate.  This would be based on how [c_rehash](https://www.openssl.org/docs/man1.0.2/man1/c_rehash.html) scans and finds certificates that are on disk to be loaded into a program.  An immediate use for this would be in the `loadVerifyLocations` function from `SSLContext`.

Lukasa:
Concretely, we already support that. What we should do is write our own tools to be able to understand what BoringSSL is _going_ to load, for use-cases like populating CA names.

Lukasa:
Resolved by #327.

PR #327, opened 2021-09-30
issue-316: Support c_rehash Directory Format

agnosticdev:
This is an enhancement to address #316 to support c_rehash directory format.

When loading certificates into the `SSL_CTX` to send the client CA names a directory path can be used for `trustRoots`.
When a directory path is used, it will be assumed that the OpenSSL tool for [c_rehash](https://www.openssl.org/docs/man1.0.2/man1/c_rehash.html) is run on this directory first to create the symlinks and truncated hash names in the directory first.
This PR checks for the rehash formatted certificates in that directory path and then resolves the symlink to load those certificates so the CA name can be extracted and set with `SSL_CTX_add_client_CA`.

Note that I removed the test for `testFullVerificationWithCANamesFromDirectory` in favor of the new test for `testRehashFormatToPopulateCANamesFromDirectory`.
The reason for this is because `testFullVerificationWithCANamesFromDirectory` assumes the previous style of picking up certificates directly from the directory and does not place them in the directory with the rehash format like `testRehashFormatToPopulateCANamesFromDirectory` does.


agnosticdev:
Looks like the nightly build failed here with:
```
16:18:13 [496/514] Compiling NIOCore Codec.swift
16:18:13 [497/514] Compiling NIOCore ConvenienceOptionSupport.swift
16:18:13 error: fatalError
16:18:13 1
```
Let me know if I need to rebase or resubmit here with a fix for this.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -579,6 +625,10 @@ extension NIOSSLContext {
         }
         return CNIOBoringSSL_sk_X509_NAME_num(caNameList)
     }
+    /// Testing stub to expose the rehash format.
+    static func getRehashFormatResult(path: String) -> Bool {
```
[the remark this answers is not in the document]
agnosticdev: Done.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -569,6 +582,39 @@ extension NIOSSLContext {
...
+        
+        // Double check that the extension did not fail to cast to an integer.
+        // Make sure that the filename is an 8 character hex based file name.
+        guard fileExtension != nil,
+              filename.count == 8,
+            filename.allSatisfy(\.isHexDigit) else { return false }
+        
+        // Check if the element is a symlink. If it is not, return false.
+        var buffer = stat()
+        let _ = lstat(path, &buffer)
```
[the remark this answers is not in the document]
agnosticdev: Done

Sources/NIOSSL/SSLContext.swift
```diff
@@ -508,15 +508,28 @@ extension NIOSSLContext {
...
-            for path in pemFilePaths {
+            // Match the c_rehash directory format and load the certificate based on this criteria.
+            let certificateFilePaths = DirectoryContents(path: path).filter {
+                self.isRehashFormat(path: $0)
+            }
+            // Load only the certificates that resolve to an existing certificate in the directory.
+            for symPath in certificateFilePaths {
+                let bufSize = Int(PATH_MAX)
+                // Resolve the symlink, and load the contents into an NIOSSLCertificate
+                let resolvedPathBuffer = UnsafeMutablePointer<CChar>.allocate(capacity: bufSize + 1)
```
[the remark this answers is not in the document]
agnosticdev: Thank you for pointing this out.  I added this to a function to `deinitialize` the pointer.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -508,15 +508,28 @@ extension NIOSSLContext {
...
+                let bytesRead = readlink(symPath, resolvedPathBuffer, bufSize)
```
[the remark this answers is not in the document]
agnosticdev: Done

agnosticdev:
@Lukasa thanks for the review.  I have updated the PR based on your feedback.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -508,15 +508,23 @@ extension NIOSSLContext {
...
-            let pemFilePaths = DirectoryContents(path: path).filter { $0.suffix(4) == ".pem" || $0.suffix(4) == ".cer" }
-            // Create the PEM files one by one and use `addCACertificateNameToList` to add the CA name to the STACK_OF(X509_NAME).
-            for path in pemFilePaths {
+            // Match the c_rehash directory format and load the certificate based on this criteria.
+            let certificateFilePaths = try DirectoryContents(path: path).filter {
+                try self._isRehashFormat(path: $0)
+            }
+            // Load only the certificates that resolve to an existing certificate in the directory.
+            for symPath in certificateFilePaths {
+                guard let resolvedPath = try self.resolveSymlink(path: symPath) else { continue }
```
Lukasa, line 517: Sorry I didn't spot this earlier, but I actually don't think we need this. We'll follow the symlink when we try to open it. Additionally, I believe c_rehash only supports PEM files, not DER, so I don't think that should be an issue either.

agnosticdev: @Lukasa you are correct about the DER support for `c_rehash`.  I just ran another test with a `PEM` and a `CER` file and only seen one hash output in the directory.  I guess I was confused by the docs for [c_rehash](https://www.openssl.org/docs/man1.0.2/man1/c_rehash.html) here:

> c_rehash scans directories and calculates a hash value of each .pem, .crt, .cer, or .crl file in the specified directory list and creates symbolic links for each file, where the name of the link is the hash value.

I will refactor this.

agnosticdev:
@Lukasa addressed latest feedback.

Lukasa:
Nope, nightly is breaking on all NIO builds at the moment because we haven't done our Sendable audit yet.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -569,6 +567,40 @@ extension NIOSSLContext {
...
+        
+        // The links created are of the form HHHHHHHH.D, where each H is a hexadecimal character and D is a single decimal digit.
+        let filename = String(filenameElement)
+        let fileExtension = Int(fileExtensionElement)
+        
+        // Double check that the extension did not fail to cast to an integer.
+        // Make sure that the filename is an 8 character hex based file name.
+        guard fileExtension != nil,
+              filename.count == 8,
+              filename.utf8.allSatisfy( { Character(UnicodeScalar($0)).isHexDigit }) else { return false }
```
Lukasa, line 592: Promoting things back up to `Character` isn't really getting us a meaningful performance win. We'd be better served by rewriting `isHexDigit` against `UTF8.CodeUnit`:

```swift
extension UTF8.CodeUnit {
    private static let asciiZero = UInt8(ascii: "0")
    private static let asciiNine = UInt8(ascii: "9")
    private static let asciiLowercaseA = UInt8(ascii: "a")
    private static let asciiLowercaseF = UInt8(ascii: "f")
    private static let asciiUppercaseA = UInt8(ascii: "A")
    private static let asciiUppercaseF = UInt8(ascii: "F")

    var isHexDigit: Bool {
        switch self {
        case (.asciiZero)...(.asciiNine),
             (.asciiLowercaseA)...(.asciiLowercaseF),
             (.asciiUppercaseA)...(.asciiUppercaseF):
            return true
        default:
            return false
        }
    }
}
```

Sources/NIOSSL/SSLContext.swift
```diff
@@ -569,6 +567,40 @@ extension NIOSSLContext {
...
+            return false
+        }
+        
+        // The links created are of the form HHHHHHHH.D, where each H is a hexadecimal character and D is a single decimal digit.
+        let filename = String(filenameElement)
+        let fileExtension = Int(fileExtensionElement)
+        
+        // Double check that the extension did not fail to cast to an integer.
+        // Make sure that the filename is an 8 character hex based file name.
+        guard fileExtension != nil,
```
Lukasa, line 590: I believe the file extension is required to be `0`, so we could just do a String comparison instead of turning it into an Int.

Tests/NIOSSLTests/TLSConfigurationTest.swift
```diff
@@ -650,92 +662,59 @@ class TLSConfigurationTest: XCTestCase {
...
+        // Only setup the serverContext here to define that our two certificate CA names were populated to the SSL_CTX.
         let countAfter = serverContext.getX509NameListCount()
         XCTAssertEqual(countAfter, 2, "CA Name List should be 2 after the Server Context is created")
         
-        XCTAssertNoThrow(clientChannel = try clientTLSChannel(context: clientContext, preHandlers:[], postHandlers: [eventHandler, handshakeWatcher, handshakeHandler], group: group, connectingTo: serverChannel.localAddress!))
+        // Delete all files that were created for this test.
+        XCTAssertNoThrow(try FileManager.default.removeItem(at: rootCAURLOne))
+        XCTAssertNoThrow(try FileManager.default.removeItem(at: rootCAURLTwo))
+        XCTAssertNoThrow(try FileManager.default.removeItem(at: URL(string: "file://" + rehashSymlinkNameOne)!))
+        XCTAssertNoThrow(try FileManager.default.removeItem(at: URL(string: "file://" + rehashSymlinkNameTwo)!))
```
Lukasa, line 675: Can we aim to put these in a defer block so that we can be quite religious about cleaning them up?

Tests/NIOSSLTests/TLSConfigurationTest.swift
```diff
@@ -650,92 +662,59 @@ class TLSConfigurationTest: XCTestCase {
...
+        XCTAssertNoThrow(try FileManager.default.createSymbolicLink(atPath: rehashSymlinkName, withDestinationPath: rootCAFilenameOne))
+        
+        // Test the success case for the symlink
+        let successSymlink = try NIOSSLContext._isRehashFormat(path: rehashSymlinkName)
+        XCTAssertTrue(successSymlink)
+        // Delete all files that were created for this test.
         XCTAssertNoThrow(try FileManager.default.removeItem(at: URL(string: "file://" + rootCAPathOne)!))
-        XCTAssertNoThrow(try FileManager.default.removeItem(at: URL(string: "file://" + rootCAPathTwo)!))
+        XCTAssertNoThrow(try FileManager.default.removeItem(at: URL(string: "file://" + rehashSymlinkName)!))
+        XCTAssertNoThrow(try FileManager.default.removeItem(at: URL(string: "file://" + newPath)!))
```
Lukasa, line 717: Same note here about defer blocks.

Lukasa (changes requested):
Great, this is moving forward nicely! I have some more notes.

Sources/NIOSSL/SSLContext.swift
agnosticdev: Wow, that is nice.  Added it to the bottom of `SSLContext` and refactored the usage of `Character`.

Tests/NIOSSLTests/TLSConfigurationTest.swift
agnosticdev: This is an **EXCELLENT** idea.

agnosticdev: This will prevent the need to clean up the directories by hand when something fails.  Thanks for calling that out.

agnosticdev:
@Lukasa feedback addressed.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -569,6 +567,39 @@ extension NIOSSLContext {
             parentSwiftContext.keyLogManager!.log(linePointer)
         }
     }
+    
+    /// Takes a path and determines if the file at this path is of c_rehash format .
+    internal static func _isRehashFormat(path: String) throws -> Bool {
+        // Check if the element’s name matches the c_rehash symlink name format.
+        // Note that `lastPathComponent` is not available here.
+        guard let lastPathComponent = path.split(separator: "/").last else { return false }
```
Lukasa, line 575: So it's up to you whether you want to take this, but an optional refactor here would be to do all of this computation on the `UTF8View` of `path`. That is, do `let utf8View = path.utf8` at the top and then do all of the rest of the `split` and similar operations on the subsequences of the `UTF8View`.

Aside from a few changes to `UInt8(ascii:)` instead of string separators this should mostly be equivalent. We can also delay constructing the `filename` string until just before the call to `lstat`.

In fact, that doesn't even need to be a string: you could do:

```swift
let filename = filenameElement.map { CInt(bitPattern: $0) }
```

It's a bit cheaper to get the pointer out of Array than out of String so that'll help a bit.

Totally up to you though, and we can always do it in a follow-up PR.

Lukasa (commented):
Ok, looks great! Only one note here. Let me know if you'd like to do the extra work, or if you want to call time on this patch here.

Sources/NIOSSL/SSLContext.swift
agnosticdev: Thanks @Lukasa.  I'll take an additional look here to include this refactor.  As a side note, I believe you also mentioned this earlier: 
> For performance reasons we should probably rewrite all of this code to work on String.UTF8View

And I had thought you were referring to only this snippet of code:
```swift
guard fileExtension != nil,
              filename.count == 8,
            filename.allSatisfy(\.isHexDigit) else { return false }
```

PR #45, opened 2018-10-01
fix Android support

albertaleksieiev:
### Android support

**Motivation:**
fix `swift-nio-ssl` compatible with Android. This is a continuation(apple/swift-nio#609) of porting `Vapor/WebSocket` and all related packages to Android.

**Modifications:**
fix `FileSystemObject::pathType` and `sysStat` Android API inconsistency

**Result:**
now `swift-nio-ssl` compatible with Android!

PR #328, opened 2021-10-12
Fix build for Android armv7

finagolfin:
Motivation:

Fix compilation error because of a type mismatch between `st_mode` and `mode_t` on 32-bit Android.

Modifications:

- Zero-extend `mode_t` constants on Android armv7 to match the types.
- Scope an earlier modification in this vein to armv7 only.

Result:

The code builds on Android armv7 again.

This line just added in #327 yesterday [broke the armv7 build on my daily Android CI](https://github.com/buttaface/swift-android-sdk/runs/3868413805?check_suite_focus=true#step:23:569). Adding these conversions fixes it and I went ahead and scoped the earlier change from #45 more narrowly too.

### DIFF
--- a/Sources/NIOSSL/SSLContext.swift
+++ b/Sources/NIOSSL/SSLContext.swift
@@ -41,7 +41,7 @@ internal enum FileSystemObject {
             return nil
         }
 
-#if os(Android)
+#if os(Android) && arch(arm)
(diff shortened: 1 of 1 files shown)


Lukasa:
Thanks! :sparkles:

PR #331, opened 2021-11-12
Trust additional roots and defaults on Darwin

Lukasa:
Motivation:

When we added additional trust roots on Darwin, we didn't add much
testing to cover that use-case because testing the default validator is
hard. This meant a bug slipped through: we called a pair of APIs in the
wrong order. As a result, users using additional trust roots on Darwin
_only_ trusted the additional ones: the defaults were unavailable.

Modifications:

- Fix the call so that we trust both the defaults and the additional
  roots.
- Refactor the security framework validator to make it possible to test
  it directly.
- Add unit tests for the security framework validator.

Result:

We'll correctly trust the additional roots and the defaults on Darwin.


glbrntt (approved):
Easy mistake... nice catch though!

PR #332, opened 2021-11-25
Provide better error on EOF during handshake

Lukasa:
Motivation:

When we get channelInactive during a handshake, we currently produce a
fairly useless error:
.handshakeFailed(NIOSSL.BoringSSLError.sslError([])). This error isn't
helpful to anyone, and it doesn't really communicate what's happened.

Sadly, due to the tightly constrained types it's really hard for us to
add new errors to handshakeFailed. Regardless, it would be good for us
to give as much diagnostic information as we can here.

Modifications:

- Changed the representation of BoringSSLInternalError to allow
  "synthetic" errors that don't come from BoringSSL at all.
- Added a new synthetic error for eof during handshake.
- Sent that error instead of a BoringSSL error stack that will always be
  empty.

Result:

- Easier to diagnose the error.

### DIFF
--- a/Sources/NIOSSL/NIOSSLHandler.swift
+++ b/Sources/NIOSSL/NIOSSLHandler.swift
@@ -111,7 +111,9 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
             // In this case the channel is going through the doHandshake steps and
             // a channelInactive is fired taking down the connection.
             // This case propogates a .handshakeFailed instead of an .uncleanShutdown.
-            channelError = NIOSSLError.handshakeFailed(.sslError(BoringSSLError.buildErrorStack()))
+            // We use a synthetic error here as the error stack will be empty, and we should try to
+            // provide some diagnostic help.
+            channelError = NIOSSLError.handshakeFailed(.sslError([.eofDuringHandshake]))
         default:
             // This is a ragged EOF: we weren't sent a CLOSE_NOTIFY. We want to send a user
             // event to notify about this before we propagate channelInactive. We also want to fail all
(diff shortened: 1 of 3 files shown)


Sources/NIOSSL/SSLErrors.swift
```diff
@@ -16,14 +16,33 @@
...
 public struct BoringSSLInternalError: Equatable, CustomStringConvertible {
-    let errorCode: UInt32
-
-    var errorMessage: String? {
-        // TODO(cory): This should become non-optional in the future, as it always succeeds.
-        var scratchBuffer = [CChar](repeating: 0, count: 512)
-        return scratchBuffer.withUnsafeMutableBufferPointer { pointer in
-            CNIOBoringSSL_ERR_error_string_n(self.errorCode, pointer.baseAddress!, pointer.count)
-            return String(cString: pointer.baseAddress!)
+    private enum Backing: Hashable {
```
glbrntt, line 19: Doesn't matter much but this is `Hashable` and `BoringSSLInternalError` is only `Equatable` -- do we want these to line up?

glbrntt (approved):
LGTM

Sources/NIOSSL/SSLErrors.swift
Lukasa: For now I didn't think it was worth making the change to the public type.

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -111,7 +111,9 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
             // In this case the channel is going through the doHandshake steps and
             // a channelInactive is fired taking down the connection.
             // This case propogates a .handshakeFailed instead of an .uncleanShutdown.
-            channelError = NIOSSLError.handshakeFailed(.sslError(BoringSSLError.buildErrorStack()))
+            // We use a synthetic error here as the error stack will be empty, and we should try to
+            // provide some diagnostic help.
+            channelError = NIOSSLError.handshakeFailed(.sslError([.eofDuringHandshake]))
```
agnosticdev, line 116: Nice addition!

Issue #337, opened 2022-01-07
Get the final handshake TLS version


Chenyibo:
When using swift-nio-ssl to carry out TLS handshake with the server, I need to know the final handshake TLS version. But currently swift-nio-ssl does not support showing the result, I hope you can expose this function, thanks!

agnosticdev:
I did a bit of research on this today and exposing this functionality does not seem like something that would fit nicely into our APIs for `NIOSSLContext`.  For example, using `NIOSSLHTTP1Client` as a test bed, and `badssl.com` as test API, I was able to establish a connection with TLSv1, and then use the `NIOSSLContext` object available in `NIOSSLHTTP1Client` to extract the TLS version, but I had to expose a public function (getTLSVersion) in `NIOSSLContext` to get the final TLS version data:

`NIOSSLHTTP1Client`
```swift
private final class HTTPResponseHandler: ChannelInboundHandler {

    ...

    func channelRead(context: ChannelHandlerContext, data: NIOAny) {
        let httpResponsePart = unwrapInboundIn(data)
        switch httpResponsePart {
        case .head(let httpResponseHeader):
            print("\(httpResponseHeader.version) \(httpResponseHeader.status.code) \(httpResponseHeader.status.reasonPhrase)")
            for (name, value) in httpResponseHeader.headers {
                print("\(name): \(value)")
            }
        case .body(var byteBuffer):
            if let data = byteBuffer.readData(length: byteBuffer.readableBytes) {
                FileHandle.standardOutput.write(data)
            }
        case .end(_):
            closeFuture = context.channel.close()
            // Extract the version from the cached SSL OpaquePointer for this connection that is saved in NIOSSLContext.
            print("Context TLS version \(sslContext.getTLSVersion())")
            promise.succeed(())
        }
    }

    ...
}

if let u = arg1 {
    url = URL(string: u)!
} else {
    url = URL(string: "https://tls-v1-0.badssl.com:1010")!
}

var tlsConfiguration = TLSConfiguration.makeClientConfiguration()
tlsConfiguration.maximumTLSVersion = .tlsv11
tlsConfiguration.minimumTLSVersion = .tlsv1

let sslContext = try! NIOSSLContext(configuration: tlsConfiguration)
```
[...]

PR #338, opened 2022-01-16
Enhancement to address issue-337 to query the final `TLSVersion`

agnosticdev:
Motivation:

To provide an enhancment to resolve issue-337 to query the final `TLSVersion`.
This allows a client channel to grab the final TLSVersion used for the connection.

Modifications:

* Added extension points to allow querying the `TLSVersion` from the `SSLConnection` based on the `Channel` and `ChannelPipeline`.

* Added an `unknown` type to `TLSVersion` to account for a default case where the `TLSVersion` cannot be derived from the conenction.

* Added a unit test `testObtainingTLSVersionOnClientChannel` to query the final `TLSVersion` when the client channel is closed.

Result:

2 APIs added to query the `TLSVersion` when on a channel.

Additional Notes:

Note that I have 2 concerns about this PR that I would like some additional feedback on:
1. First is the `unknown` case for `TLSVersion`.  This may not be needed as BoringSSL may just return `tlsv12` each time the version cannot be derived from the connection.  Let me know your thoughts.
2. I am not wild about the way this is being unit-tested, but I wanted to do it this way because it provides the `TLSVersion` on the `Channel` after it is closed.  Let me know your thoughts.

### DIFF
--- a/Sources/NIOSSL/NIOSSLHandler.swift
+++ b/Sources/NIOSSL/NIOSSLHandler.swift
@@ -547,34 +547,25 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
 }
 
 extension NIOSSLHandler {
-    /// Variable that can be queried during the connection lifecycle to grab the `TLSVesion` used on the `SSLConnection`.
+    /// Variable that can be queried during the connection lifecycle to grab the `TLSVersion` used on the `SSLConnection`.
     public var tlsVersion: TLSVersion? {
         return self.connection.getTLSVersionForConnection()
     }
 }
 
 extension Channel {
-    ///  API to extract the `TLSVersion` from the channel `whenComplete` is fired.
-    ///  This API is intended to provide the final `TLSVersion` on the channel.
-    public func nioSSL_tlsVersion() -> TLSVersion? {
+    ///  API to extract the `TLSVersion` from an EventLoopFuture.
+    ///  This calls into the `ChannelPipeline.SynchronousOperations` to extract the `TLSVersion`.
+    public func nioSSL_tlsVersion() -> EventLoopFuture<TLSVersion?> {
         let handler = self.pipeline.handler(type: NIOSSLHandler.self)
-        var tlsVersion: TLSVersion?
-        handler.whenComplete { sslHandlerResult in
-            switch sslHandlerResult {
-            case .success(let nioSSLHandler):
-                tlsVersion = nioSSLHandler.tlsVersion
-            case .failure(_):
-                tlsVersion = nil
-            }
-        }
-        return tlsVersion
+        return handler.eventLoop.makeSucceededFuture(try? self.pipeline.syncOperations.nioSSL_tlsVersion())
     }
 }
 
 extension ChannelPipeline.SynchronousOperations {
     /// API to query the `TLSVersion` directly from the `ChannelPipeline`.
-    public func nioSSL_tlsVersion() -> TLSVersion? {
-        let handler = try! self.handler(type: NIOSSLHandler.self)
+    public func nioSSL_tlsVersion() throws -> TLSVersion? {
+        let handler = try self.handler(type: NIOSSLHandler.self)
         return handler.tlsVersion
     }
 }
--- a/Sources/NIOSSL/NIOSSLHandler.swift
+++ b/Sources/NIOSSL/NIOSSLHandler.swift
@@ -548,24 +548,34 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
 
 extension NIOSSLHandler {
     /// Variable that can be queried during the connection lifecycle to grab the `TLSVesion` used on the `SSLConnection`.
-    public var tlsVersionForHandler: TLSVersion {
+    public var tlsVersion: TLSVersion? {
         return self.connection.getTLSVersionForConnection()
     }
 }
 
 extension Channel {
-    ///  API to add the `NIOSSLHandler` to a channel to extract the `TLSVersion` during
-    ///  lifecycle events such as `whenSuccess` or `whenComplete`.
-    public func nioSSL_tlsVersionOnChannel() -> EventLoopFuture<NIOSSLHandler> {
-        return self.pipeline.handler(type: NIOSSLHandler.self)
+    ///  API to extract the `TLSVersion` from the channel `whenComplete` is fired.
+    ///  This API is intended to provide the final `TLSVersion` on the channel.
+    public func nioSSL_tlsVersion() -> TLSVersion? {
+        let handler = self.pipeline.handler(type: NIOSSLHandler.self)
+        var tlsVersion: TLSVersion?
+        handler.whenComplete { sslHandlerResult in
+            switch sslHandlerResult {
+            case .success(let nioSSLHandler):
+                tlsVersion = nioSSLHandler.tlsVersion
+            case .failure(_):
+                tlsVersion = nil
+            }
+        }
+        return tlsVersion
     }
 }
 
 extension ChannelPipeline.SynchronousOperations {
     /// API to query the `TLSVersion` directly from the `ChannelPipeline`.
-    public func nioSSL_tlsVersionOnChannelPipeline() -> TLSVersion {
+    public func nioSSL_tlsVersion() -> TLSVersion? {
         let handler = try! self.handler(type: NIOSSLHandler.self)
-        return handler.tlsVersionForHandler
+        return handler.tlsVersion
     }
 }
 
--- a/Sources/NIOSSL/NIOSSLHandler.swift
+++ b/Sources/NIOSSL/NIOSSLHandler.swift
@@ -546,6 +546,28 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
     }
 }
 
+extension NIOSSLHandler {
+    /// Variable that can be queried during the connection lifecycle to grab the `TLSVesion` used on the `SSLConnection`.
+    public var tlsVersionForHandler: TLSVersion {
+        return self.connection.getTLSVersionForConnection()
+    }
+}
+
+extension Channel {
+    ///  API to add the `NIOSSLHandler` to a channel to extract the `TLSVersion` during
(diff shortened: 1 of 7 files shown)


agnosticdev:
Scratch the `unknown` case comments above, this created an API Breakage which will never do.  I refactored this code to default to `tlsv12` as this is what BoringSSL will default to anyways.

agnosticdev:
Fix for #337

Sources/NIOSSL/SSLConnection.swift
```diff
@@ -407,6 +407,24 @@ internal final class SSLConnection {
...
+          case TLS1_3_VERSION:
+            return .tlsv13
+          case TLS1_2_VERSION:
+            return .tlsv12
+          case TLS1_1_VERSION:
+            return .tlsv11
+          case TLS1_VERSION:
+            return .tlsv1
+        default:
+            return .tlsv12 // This is the default for BoringSSL
```
Lukasa, line 425: This should probably return `nil` if we don't have an answer.

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -546,6 +546,28 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
...
+    /// Variable that can be queried during the connection lifecycle to grab the `TLSVesion` used on the `SSLConnection`.
+    public var tlsVersionForHandler: TLSVersion {
+        return self.connection.getTLSVersionForConnection()
+    }
+}
+
+extension Channel {
+    ///  API to add the `NIOSSLHandler` to a channel to extract the `TLSVersion` during
+    ///  lifecycle events such as `whenSuccess` or `whenComplete`.
+    public func nioSSL_tlsVersionOnChannel() -> EventLoopFuture<NIOSSLHandler> {
```
Lukasa, line 559: This returns the handler, not the version. It should probably return the version!

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -546,6 +546,28 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
     }
 }
 
+extension NIOSSLHandler {
+    /// Variable that can be queried during the connection lifecycle to grab the `TLSVesion` used on the `SSLConnection`.
+    public var tlsVersionForHandler: TLSVersion {
```
Lukasa, line 551:
```suggestion
    public var tlsVersion: TLSVersion {
```

Sources/NIOSSL/NIOSSLHandler.swift
Lukasa, line 559:
```suggestion
    public func nioSSL_tlsVersion() -> EventLoopFuture<NIOSSLHandler> {
```

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -546,6 +546,28 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
...
+    ///  API to add the `NIOSSLHandler` to a channel to extract the `TLSVersion` during
+    ///  lifecycle events such as `whenSuccess` or `whenComplete`.
+    public func nioSSL_tlsVersionOnChannel() -> EventLoopFuture<NIOSSLHandler> {
+        return self.pipeline.handler(type: NIOSSLHandler.self)
+    }
+}
+
+extension ChannelPipeline.SynchronousOperations {
+    /// API to query the `TLSVersion` directly from the `ChannelPipeline`.
+    public func nioSSL_tlsVersionOnChannelPipeline() -> TLSVersion {
```
Lukasa, line 566:
```suggestion
    public func nioSSL_tlsVersion() -> TLSVersion {
```

Tests/NIOSSLTests/TLSConfigurationTest.swift
```diff
@@ -1077,6 +1096,50 @@ class TLSConfigurationTest: XCTestCase {
...
+            certificateChain: [.certificate(TLSConfigurationTest.cert1)],
+            privateKey: .privateKey(TLSConfigurationTest.key1)
+        )
+        serverConfig.maximumTLSVersion = .tlsv11
+        serverConfig.certificateVerification = .none
+        
+        let clientContext = try assertNoThrowWithValue(NIOSSLContext(configuration: clientConfig))
+        let serverContext = try assertNoThrowWithValue(NIOSSLContext(configuration: serverConfig))
+
+        let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
```
Lukasa, line 1117: Rather than use MTELG can we use one of the `BackToBackEmbeddedChannel` constructions? These tests run much faster and are less flaky and weird.

Lukasa (changes requested):
Great, this looks really good! I've left a few notes in the diff. Mostly these are small, naming changes and such, but the two larger notes are a suggestion for an alternative testing strategy and a recommendation that we don't use TLSv1.2 as a default.

Tests/NIOSSLTests/TLSConfigurationTest.swift
agnosticdev: Sure, no problem. I added an initial refactor using `BackToBackEmbeddedChannel`, let me know your thoughts.

Sources/NIOSSL/SSLConnection.swift
agnosticdev: I refactored this not to use `.tlsv12` in the latest and to return `nil` instead on the default case.  This of course changes `TLSVersion` to be an optional returned here but that didn't have too much impact on the final usage of this API.  Let me know your thoughts.

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -546,6 +546,38 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
     }
 }
 
+extension NIOSSLHandler {
+    /// Variable that can be queried during the connection lifecycle to grab the `TLSVesion` used on the `SSLConnection`.
```
Lukasa, line 550:
```suggestion
    /// Variable that can be queried during the connection lifecycle to grab the `TLSVersion` used on the `SSLConnection`.
```

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -546,6 +546,38 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
...
+            }
+        }
+        return tlsVersion
+    }
+}
+
+extension ChannelPipeline.SynchronousOperations {
+    /// API to query the `TLSVersion` directly from the `ChannelPipeline`.
+    public func nioSSL_tlsVersion() -> TLSVersion? {
+        let handler = try! self.handler(type: NIOSSLHandler.self)
```
Lukasa, line 577: This should just `throw`.

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -546,6 +546,38 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
...
+    /// Variable that can be queried during the connection lifecycle to grab the `TLSVesion` used on the `SSLConnection`.
+    public var tlsVersion: TLSVersion? {
+        return self.connection.getTLSVersionForConnection()
+    }
+}
+
+extension Channel {
+    ///  API to extract the `TLSVersion` from the channel `whenComplete` is fired.
+    ///  This API is intended to provide the final `TLSVersion` on the channel.
+    public func nioSSL_tlsVersion() -> TLSVersion? {
```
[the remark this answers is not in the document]
agnosticdev: I made a change here to return `EventLoopFuture<TLSVersion?>` and then `.map` to extract the `TLSVersion` from the `EventLoopFuture`.  However, to construct the `EventLoopFuture` I used `makeSucceededFuture` and then used `self.pipeline.syncOperations.nioSSL_tlsVersion()` to grab the actual value.  I have a feeling that this isn't what you were looking for though.  Let me know your thoughts.

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -546,6 +546,29 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
...
+        return self.connection.getTLSVersionForConnection()
+    }
+}
+
+extension Channel {
+    ///  API to extract the `TLSVersion` from an EventLoopFuture.
+    ///  This calls into the `ChannelPipeline.SynchronousOperations` to extract the `TLSVersion`.
+    public func nioSSL_tlsVersion() -> EventLoopFuture<TLSVersion?> {
+        let handler = self.pipeline.handler(type: NIOSSLHandler.self)
+        return handler.eventLoop.makeSucceededFuture(try? self.pipeline.syncOperations.nioSSL_tlsVersion())
```
Lukasa, line 561: What I meant here was this:

```swift
return self.pipeline.handler(type: NIOSSLHandler.self).map {
    $0.tlsVersion
}
```

Lukasa:
This can happen if the builder takes this opportunity to rebuild the docker image. I can usually just kick it again.

Lukasa:

Tests/NIOSSLTests/TLSConfigurationTest.swift
```diff
@@ -1077,6 +1077,54 @@ class TLSConfigurationTest: XCTestCase {
...
+        XCTAssertTrue(b2b.server.handshakeSucceeded)
+        
+        var tlsVersion: TLSVersion?
+        XCTAssertNoThrow(tlsVersion = try b2b.client.pipeline.syncOperations.nioSSL_tlsVersion())
+        XCTAssertEqual(tlsVersion!, .tlsv11)
+        
+        let tlsVersionForChannel = b2b.client.nioSSL_tlsVersion()
+        let _ = tlsVersionForChannel.map { tlsVersion in
+            XCTAssertEqual(tlsVersion, .tlsv11)
+        }
```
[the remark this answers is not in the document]
agnosticdev: Yeah, that works well.  Refactored in 7116737

Lukasa (approved):

Lukasa:
Nice one!

agnosticdev:
Thank you @Lukasa !

PR #345, opened 2022-03-08
Add RandomAccessCollection conformance to `SubjectAlternativeNames` and make it public

dnadoba:
We want to expose `SubjectAlternativeNames` without promising API stability. Therefore `SubjectAlternativeNames` and related symbols are now public but have a leading underscore to mark them as unstable. 

### Modifications
- Add `RandomAccessCollection` conformance to `_SubjectAlternativeNames`
- Add separate `Element` type to `_SubjectAlternativeNames` that moves OpenSSL to Swift enum type conversations from the collection to `.get()` on `Element`
- make `_SubjectAlternativeNames` and related symbols public and mark them as unstable

### Result
Users of the library can access subject alternative names from a `NIOSSLCertificate`


Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -490,13 +493,15 @@ extension NIOSSLCertificate: CustomStringConvertible {
             let commonName = String(decoding: commonNameBytes, as: UTF8.self)
             desc += ";common_name=" + commonName
         }
-        if let alternativeName = self.subjectAlternativeNames() {
-            let altNames = alternativeName.map { name in
-                switch name {
+        if let alternativeName = self._subjectAlternativeNames() {
+            let altNames = alternativeName.compactMap { name in
+                switch try? name.get() {
```
dnadoba, line 498: Should we swallow or throw here? The previous implementation did swallow the error deeper in the stack.

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -407,73 +407,76 @@ extension NIOSSLCertificate: Hashable {
...
+    public struct Element {
+        private let collection: _SubjectAlternativeNames
+        private let index: Int
+        
+        fileprivate init(collection: _SubjectAlternativeNames, index: Int) {
+            self.collection = collection
+            self.index = index
+        }
+        
+        func get() throws -> NIOSSLCertificate._AlternativeName? {
```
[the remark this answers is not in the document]
dnadoba: Sounds good to me. Where would we then move the IPAddress construction logic? Should we still keep the current `NIOSSLCertificate._AlternativeName` enum?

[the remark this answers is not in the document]
Lukasa: We can move the construction logic to the point where it is actually used. And nah, I'd throw that enum away.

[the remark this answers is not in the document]
dnadoba: @Lukasa PR update with the new Element type

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -407,73 +397,108 @@ extension NIOSSLCertificate: Hashable {
...
-    private var nextIdx: Int
-    private let stackSize: Int
+    public struct NameType: Hashable {
+        var rawValue: Int
 
-    init(nameStack: OpaquePointer) {
-        self.nameStack = nameStack
-        self.stackSize = CNIOBoringSSLShims_sk_GENERAL_NAME_num(nameStack)
-        self.nextIdx = 0
+        fileprivate init(_ rawCode: Int) {
```
Lukasa, line 407: This, and `var rawValue`, may as well be `public`: the API is more useful if users can refer to other types.

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -407,73 +397,108 @@ extension NIOSSLCertificate: Hashable {
...
-        switch bytes.count {
+extension _SubjectAlternativeName.Contents: RandomAccessCollection {
+    public var startIndex: Int { buffer.startIndex }
+    public var endIndex: Int { buffer.endIndex }
+    
+    public subscript(position: Int) -> UInt8 {
+        get {
+            buffer[position]
+        }
+    }
```
Lukasa, line 446: Can we make the computed properties and the subscript `@inlinable`?

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -407,73 +397,108 @@ extension NIOSSLCertificate: Hashable {
...
+
+extension _SubjectAlternativeName.IPAddress {
+    init?(_ subjectAlternativeName: _SubjectAlternativeName) {
+        guard subjectAlternativeName.nameType == .ipAddress else {
+            return nil
+        }
+        switch subjectAlternativeName.contents.count {
         case 4:
-            let addr = bytes.baseAddress?.withMemoryRebound(to: in_addr.self, capacity: 1) {
+            let addr = subjectAlternativeName.contents.buffer.baseAddress?.withMemoryRebound(to: in_addr.self, capacity: 1) {
```
Lukasa, line 456: Instead of reaching to `buffer` directly, can we use a `withUnsafePointer` function? It's possible to drop the reference to `_SubjectAlternativeNames` when we use a direct member access like this.

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -407,73 +397,108 @@ extension NIOSSLCertificate: Hashable {
...
-        }
+/// Collection of all Subject Alternative Names from a `NIOSSLCertificate`
+public final class _SubjectAlternativeNames: RandomAccessCollection {
+    public typealias Element = _SubjectAlternativeName
 
-        guard let name = CNIOBoringSSLShims_sk_GENERAL_NAME_value(self.nameStack, self.nextIdx) else {
-            fatalError("Unexpected null pointer when unwrapping SAN value")
-        }
+    public var startIndex: Int { 0 }
+    public var endIndex: Int { stackSize }
```
Lukasa, line 482: Can we make these inlinable too?

Sources/NIOSSL/SSLCertificate.swift
```diff
@@ -407,73 +397,108 @@ extension NIOSSLCertificate: Hashable {
...
         }
     }
+}
 
-    func next() -> NIOSSLCertificate.AlternativeName? {
-        guard self.nextIdx < self.stackSize else {
-            return nil
-        }
+/// Collection of all Subject Alternative Names from a `NIOSSLCertificate`
+public final class _SubjectAlternativeNames: RandomAccessCollection {
```
Lukasa, line 478: Can we wrap this `class` in a `struct`, where the `struct` is public and forwards to equivalent methods on the `class`?

Lukasa (commented):
Cool, looking really good so far!

dnadoba (commented):
@Lukasa all review comments should be addressed. I have now moved `_SubjectAlternativeName` and related types to a separate file as well because it did become sufficiently large.

Sources/NIOSSL/SubjectAlternativeName.swift
```diff
@@ -0,0 +1,196 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+@_implementationOnly import CNIOBoringSSL
+@_implementationOnly import CNIOBoringSSLShims
+
+/// Collection of all Subject Alternative Names from a `NIOSSLCertificate`
+public struct _SubjectAlternativeNames {
+    @usableFromInline
+    internal final class Storage: RandomAccessCollection {
```
Lukasa, line 21: I don't think we need to conform this to RAC.

Sources/NIOSSL/SubjectAlternativeName.swift
```diff
@@ -0,0 +1,196 @@
...
+        public typealias Element = _SubjectAlternativeName
+        
+        @inlinable public var startIndex: Int { 0 }
+        @inlinable public var endIndex: Int { stackSize }
```
Lukasa, line 25: We don't need these two fields.

Sources/NIOSSL/SubjectAlternativeName.swift
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        deinit {
+            CNIOBoringSSL_GENERAL_NAMES_free(self.nameStack)
+        }
+    }
+    
+    
+    @usableFromInline internal var storage: Storage
+    internal init(nameStack: OpaquePointer) {
```
Lukasa, line 54: Nit: whitespace above this point.

Sources/NIOSSL/SubjectAlternativeName.swift
```diff
@@ -0,0 +1,196 @@
...
+        storage[position]
+    }
+    
+    @inlinable public var startIndex: Int {
+        storage.startIndex
+    }
+    
+    @inlinable public var endIndex: Int {
+        storage.endIndex
+    }
```
Lukasa, line 71: Nit: we still prefer explicit `self` here, even though older code doesn't use it. That applies throughout this file.

dnadoba:
Soundness fails because of year 2022 is currently not a accepted year in license headers. Fixed in https://github.com/apple/swift-nio-ssl/pull/346

Sources/NIOSSL/SubjectAlternativeName.swift
```diff
@@ -0,0 +1,201 @@
...
+    public var nameType: NameType
+    public var contents: Contents
+}
+
+extension _SubjectAlternativeName.Contents: RandomAccessCollection {
+    
+    @inlinable public var startIndex: Int { self.buffer.startIndex }
+    @inlinable public var endIndex: Int { self.buffer.endIndex }
+    
+    @inlinable public subscript(position: Int) -> UInt8 {
```
Lukasa, line 126: We should add a bounds check here.

Sources/NIOSSL/SubjectAlternativeName.swift
```diff
@@ -0,0 +1,201 @@
...
+        
+        fileprivate let nameStack: OpaquePointer
+        @usableFromInline internal let stackSize: Int
+
+        internal init(nameStack: OpaquePointer) {
+            self.nameStack = nameStack
+            self.stackSize = CNIOBoringSSLShims_sk_GENERAL_NAME_num(nameStack)
+        }
+        
+        public subscript(position: Int) -> Element {
```
Lukasa, line 40: We should add a bounds check here.

Sources/NIOSSL/SubjectAlternativeName.swift
```diff
@@ -0,0 +1,201 @@
...
+extension _SubjectAlternativeName.IPAddress {
+    
+    internal init?(_ subjectAlternativeName: _SubjectAlternativeName) {
+        guard subjectAlternativeName.nameType == .ipAddress else {
+            return nil
+        }
+        switch subjectAlternativeName.contents.count {
+        case 4:
+            let addr = subjectAlternativeName.contents.withUnsafeBufferPointer {
+                $0.baseAddress?.withMemoryRebound(to: in_addr.self, capacity: 1) {
```
Lukasa, line 140: Oh we just cleaned this up in NIO, let's do the same here: this should be:

```
$0.baseAddress.map { UnsafeRawPointer($0) }.load(in_addr.self)
```

The same transformation should go on below as well.

Lukasa (commented):
Teeny tiny nits and then we're good to go.

Lukasa (approved):

Lukasa:
Nightly has a failure in the integration tests, probably the same one that affected NIO. Merging over it.

PR #352, opened 2022-03-30
Add `NIOSSLObjectIdentifier` (OID) type

dnadoba:
This PR adds a wrapper around `ASN1_OBJECT` that represents an Object Identifier (OID). It manages memory through ARC and implements `Equatable`, `Hashable`, `CustomStringConvertible` and `LosslessStringConvertible`.

### DIFF
--- a/Sources/NIOSSL/ObjectIdentifier.swift
+++ b/Sources/NIOSSL/ObjectIdentifier.swift
@@ -1,0 +1,110 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2022 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+@_implementationOnly import CNIOBoringSSL
+@_implementationOnly import CNIOBoringSSLShims
+
+/// Object Identifier (OID)
+public struct NIOSSLObjectIdentifier: Hashable {
+    fileprivate final class Storage {
+        
+        /// All operations accessing `reference` need to be implemented while guaranteeing that the enclosing `self` has a reference count of at least one.
+        /// Otherwise `reference` could already be freed. This would result in undefined behaviour as we access a dangling pointer.
+        /// One way we can guarantee this is by implementation all operations as methods/computed properties on `Storage`.
+        /// Swift guarantees that `self` is not deallocated during the execution of a method/computed property on `self`.
+        private let reference: OpaquePointer
+        
+        fileprivate init?(_ string: String) {
+            let result = string.withCString { string in
+                CNIOBoringSSL_OBJ_txt2obj(string, 0)
+            }
+            guard let reference = result else {
+                return nil
+            }
+            self.reference = reference
+        }
+        
+        fileprivate init(reference: OpaquePointer) {
+            self.reference = reference
+        }
+        
+        deinit {
+            CNIOBoringSSL_ASN1_OBJECT_free(reference)
+        }
+    }
+    
+    private let storage: Storage
+    
+    /// Creates a Object Identifier (OID) from its textual dotted representation (e.g. `1.2.3`)
+    /// - Parameter string: textual dotted representation of an OID
+    public init?(_ string: String) {
+        guard let storage = Storage(string) else {
+            return nil
+        }
+        self.storage = storage
+    }
+    
+    /// Creates an Object Identifier (OID) from an OpenSSL reference.
+    /// - Note: initialising an ``NIOSSLObjectIdentifier`` takes ownership of the reference and will free it after the reference count drops to zero
+    /// - Parameter reference: reference to a valid OpenSSL OID aka OBJ
+    internal init(reference: OpaquePointer) {
+        self.storage = Storage(reference: reference)
+    }
+}
+
+extension NIOSSLObjectIdentifier.Storage: Equatable {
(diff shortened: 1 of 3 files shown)


Sources/NIOSSL/ObjectIdentifier.swift
```diff
@@ -0,0 +1,110 @@
...
+        
+        /// All operations accessing `reference` need to be implemented while guaranteeing that the enclosing `self` has a reference count of at least one.
+        /// Otherwise `reference` could already be freed. This would result in undefined behaviour as we access a dangling pointer.
+        /// One way we can guarantee this is by implementation all operations as methods/computed properties on `Storage`.
+        /// Swift guarantees that `self` is not deallocated during the execution of a method/computed property on `self`.
+        private let reference: OpaquePointer
+        
+        fileprivate init?(_ string: String) {
+            let result = string.withCString { string in
+                CNIOBoringSSL_OBJ_txt2obj(string, 0)
```
Lukasa, line 30: Are we deliberately setting this to 0? Should we only tolerate the dotted sequence of numbers?

Sources/NIOSSL/ObjectIdentifier.swift
```diff
@@ -0,0 +1,110 @@
...
+        guard let storage = Storage(string) else {
+            return nil
+        }
+        self.storage = storage
+    }
+    
+    /// Creates an Object Identifier (OID) from an OpenSSL reference.
+    /// - Note: initialising an ``NIOSSLObjectIdentifier`` takes ownership of the reference and will free it after the reference count drops to zero
+    /// - Parameter reference: reference to a valid OpenSSL OID aka OBJ
+    internal init(reference: OpaquePointer) {
```
Lukasa, line 61: Elsewhere in the module we use the pattern `init(takingOwnershipOf reference:)`.

Sources/NIOSSL/ObjectIdentifier.swift
```diff
@@ -0,0 +1,110 @@
...
+            let result = string.withCString { string in
+                CNIOBoringSSL_OBJ_txt2obj(string, 0)
+            }
+            guard let reference = result else {
+                return nil
+            }
+            self.reference = reference
+        }
+        
+        fileprivate init(reference: OpaquePointer) {
```
Lukasa, line 38: This should also use `takingOwnershipOf reference`.

Sources/NIOSSL/ObjectIdentifier.swift
```diff
@@ -0,0 +1,110 @@
...
+    fileprivate var description: String {
+        var failed = false
+        let oid = String(unsafeUninitializedCapacity: 80) { buffer in
+            // OBJ_obj2txt() is awkward and messy to use: it doesn't follow the convention of other OpenSSL functions where the buffer can be set to NULL to determine the amount of data that should be written. Instead buf must point to a valid buffer and buf_len should be set to a positive value. A buffer length of 80 should be more than enough to handle any OID encountered in practice.
+            // source: https://linux.die.net/man/3/obj_obj2txt
+            let result = buffer.withMemoryRebound(to: CChar.self) { buffer in
+                CNIOBoringSSL_OBJ_obj2txt(buffer.baseAddress, Int32(buffer.count), self.reference, 1)
+            }
+            guard result >= 0 else {
+                // result of -1 indicates that it could not successfully
```
Lukasa, line 92: That it could not what successfully?

Sources/NIOSSL/ObjectIdentifier.swift
```diff
@@ -0,0 +1,110 @@
...
+extension NIOSSLObjectIdentifier.Storage: Hashable {
+    fileprivate func hash(into hasher: inout Hasher) {
+        let length = CNIOBoringSSL_OBJ_length(self.reference)
+        let data = CNIOBoringSSL_OBJ_get0_data(self.reference)
+        let buffer = UnsafeRawBufferPointer(start: data, count: length)
+        hasher.combine(bytes: buffer)
+    }
+}
+
+@available(macOS 11, iOS 14, tvOS 14, watchOS 7, *)
```
Lukasa, line 81: I don't think we need an availability guard here: let's bring over `customUnsafeUninitializedCapacity` from https://github.com/apple/swift-nio-http2/blob/dbfc8fcd1a0cc4d7254a060501120d378e91f16c/Sources/NIOHPACK/HuffmanCoding.swift#L267-L289

Sources/NIOSSL/ObjectIdentifier.swift
dnadoba: The docs just say that -1 is an error and are not more specific. Change the comment to reflect that.

Sources/NIOSSL/ObjectIdentifier.swift
dnadoba: I think we only need to support the dotted sequence of numbers. Changed the setting to `1` and copied the documentation from OpenSSL inline.

Sources/NIOSSL/String+unsafeUninitializedCapacity.swift
```diff
@@ -0,0 +1,57 @@
...
+
+        self = String(decoding: UnsafeMutableBufferPointer(start: buffer.baseAddress!, count: initializedCount), as: UTF8.self)
+    }
+}
+
+// Frustratingly, Swift 5.3 shipped before the macOS 11 SDK did, so we cannot gate the availability of
+// this declaration on having the 5.3 compiler. This has caused a number of build issues. While updating
+// to newer Xcodes does work, we can save ourselves some hassle and just wait until 5.4 to get this
+// enhancement on Apple platforms.
+#if (compiler(>=5.3) && !(os(macOS) || os(iOS) || os(tvOS) || os(watchOS))) || compiler(>=5.4)
```
Lukasa, line 39: As we've just dropped 5.4, I think we can clean this conditional up now.

PR #356, opened 2022-04-22
Update BoringSSL to f299342e38fd34b589604ef5b730623da13ee235

glbrntt:


Lukasa (approved):

Lukasa:
We can merge over 5.7 and main, that's a real issue but we can fix it separately.

PR #358, opened 2022-04-27
Add support for owned references to `NIOSSLObjectIdentifier`

dnadoba:
### Motivation
BoringSSL can give us a pointer to an OID which is part of a larger allocation that can not separately be deallocated. This is the case for Certificate Extensions where each extension has an OID but the memory of this OID is actually owned by the certificate. 

### Modification
- add support for specifying an owner of the memory which we need to retain to keep the memory alive
- add support for copying the memory and taking ownership of that copy

### Result
We can vend users a `NIOSSLObjectIdentifier` for e.g. a Certificate Extension OID without needing to do a copy first.


Tests/NIOSSLTests/ObjectIdentifierTests+XCTest.swift
```diff
@@ -0,0 +1,38 @@
...
+//
+import XCTest
+
+///
+/// NOTE: This file was generated by generate_linux_tests.rb
+///
+/// Do NOT edit this file directly as it will be regenerated automatically when needed.
+///
+
+extension ObjectIdentifierTests {
```
dnadoba, line 25: Just noticed that I accidentally forgot to add the `Tests` suffix to this file. Renamed the file now and run `./scripts/generate_linux_tests.rb`

dnadoba, line 25: Just noticed that I accidentally forgot to add the `Tests` suffix to this file. Renamed the file now and run `./scripts/generate_linux_tests.rb`

Sources/NIOSSL/ObjectIdentifier.swift
```diff
@@ -16,35 +16,45 @@
...
-        
-        fileprivate init?(_ string: String) {
-            let result = string.withCString { string in
-                // If no_name (the last parameter of CNIOBoringSSL_OBJ_txt2obj) is 0 then long names and
-                // short names will be interpreted as well as numerical forms.
-                // If no_name is 1 only the numerical form is acceptable.
-                // source: https://www.openssl.org/docs/manmaster/man3/OBJ_txt2obj.html
-                CNIOBoringSSL_OBJ_txt2obj(string, 1)
+public struct NIOSSLObjectIdentifier {
+    enum Storage {
```
Lukasa, line 20: Should this be `private`?

Sources/NIOSSL/ObjectIdentifier.swift
```diff
@@ -16,35 +16,45 @@
...
-        fileprivate init(takingOwnershipOf reference: OpaquePointer) {
-            self.reference = reference
+        case owned(Deallocator)
+        case unowned(reference: OpaquePointer!, owner: AnyObject)
+        
+        init(takeOwnershipOf reference: OpaquePointer!) {
+            self = .owned(.init(takeOwnershipOf: reference))
+        }
+        
+        init(takeUnowned reference: OpaquePointer!, owner: AnyObject) {
```
Lukasa, line 40: I think a better set of labels is:

```suggestion
        init(borrowing reference: OpaquePointer!, owner: AnyObject) {
```

Sources/NIOSSL/ObjectIdentifier.swift
```diff
@@ -53,65 +63,89 @@ public struct NIOSSLObjectIdentifier: Hashable {
...
-extension NIOSSLObjectIdentifier.Storage: Equatable {
-    fileprivate static func == (lhs: NIOSSLObjectIdentifier.Storage, rhs: NIOSSLObjectIdentifier.Storage) -> Bool {
-        CNIOBoringSSL_OBJ_cmp(lhs.reference, rhs.reference) == 0
+    
+    /// Creates an Object Identifier (OID) from an OpenSSL reference.
+    /// - Note: initialising an ``NIOSSLObjectIdentifier`` does **not** take ownership of the memory. Instead ``NIOSSLObjectIdentifier`` keeps a reference to the owning object which it will retain for the lifetime of itself.
+    /// - Parameters
+    ///   - reference: reference to a valid OpenSSL OID aka OBJ
+    ///   - owner: owner of the memory `reference` is pointing to which it will retain.
+    internal init(takeUnowned reference: OpaquePointer!, owner: AnyObject) {
```
Lukasa, line 91: I think a better set of labels is 

```suggestion
    internal init(borrowing reference: OpaquePointer!, owner: AnyObject) {
```

Sources/NIOSSL/ObjectIdentifier.swift
```diff
@@ -53,65 +63,89 @@ public struct NIOSSLObjectIdentifier: Hashable {
...
+        self.storage = .init(takeOwnershipOf: reference)
     }
-}
-
-extension NIOSSLObjectIdentifier.Storage: Equatable {
-    fileprivate static func == (lhs: NIOSSLObjectIdentifier.Storage, rhs: NIOSSLObjectIdentifier.Storage) -> Bool {
-        CNIOBoringSSL_OBJ_cmp(lhs.reference, rhs.reference) == 0
+    
+    /// Creates an Object Identifier (OID) from an OpenSSL reference.
+    /// - Note: initialising an ``NIOSSLObjectIdentifier`` does **not** take ownership of the memory. Instead ``NIOSSLObjectIdentifier`` keeps a reference to the owning object which it will retain for the lifetime of itself.
```
Lukasa, line 87: We should say that initialising this with _this_ constructor does not take ownership.

PR #359, opened 2022-04-27
Add `NIOSSLCertficiateExtensions`

dnadoba:
**Note:** All changes to `NIOObjectIdentifier` in `NIOObjectIdentifier.swift` and `NIOObjectIdentifierTests.swift` are part of #358 and should be reviewed there if possible.

### Motivation
We want to allow users to iterate and inspect certificate extensions

### Modification
- expose `STACK(X509_EXTENSION)` of a `NIOSSLCertificate` through a memory managed and type-safe   `NIOSSLCertficiateExtensions` which conforms to `RandomAccessCollection`
- `NIOSSLCertficiateExtensions` `Element` type is a type safe and memory managed version of `X509_EXTENSION` which gives access to `objectIdentifier`, `isCritical`, and `data` of an extension.

### Result
Users can inspect certificate extensions of a  `NIOSSLCertificate`


Sources/NIOSSL/SSLCertificateExtensions.swift
```diff
@@ -0,0 +1,197 @@
...
+        init(takeOwnershipOf reference: OpaquePointer!) {
+            self = .owned(.init(takeOwnershipOf: reference))
+        }
+        
+        init(borrowing reference: OpaquePointer!, owner: AnyObject) {
+            self = .borrowed(reference: reference, owner: owner)
+        }
+        
+        /// The owner of the memory to which the reference points
+        var owner: AnyObject {
```
Lukasa, line 44: Do we need this?

Lukasa (commented):
Nice, looks really good! Just a few cleanup notes.

Sources/NIOSSL/SSLCertificateExtensions.swift
```diff
@@ -0,0 +1,197 @@
...
+                self.reference = reference
+            }
+            
+            deinit {
+                CNIOBoringSSL_sk_X509_EXTENSION_free(self.reference)
+            }
+        }
+        
+        case owned(Deallocator)
+        case borrowed(reference: OpaquePointer!, owner: AnyObject)
```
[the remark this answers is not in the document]
dnadoba: It either needs to be optional/IUO or we need to handle the empty case differently [because `CNIOBoringSSL_X509_get0_extensions` returns `nil` if no extensions are present](https://www.openssl.org/docs/man1.1.1/man3/X509_get0_extensions.html).
The good thing is that `CNIOBoringSSL_sk_X509_EXTENSION_num` returns `0` if we give it a `nil` pointer so we should be safe. That being said, I'm wondering if we should rely on that behaviour or if we should add our own additional guard. WDYT?

Sources/NIOSSL/SSLCertificateExtensions.swift
dnadoba: Yes because we need to give `_SSLCertificateExtension`, `_SSLCertificateExtension.Data` and `NIOSSLObjectIdentifier` an owner of type `AnyObject`. `_SSLCertificateExtensions` and `_SSLCertificateExtensions.Storage` are not classes and therefore we need this property to get an `AnyObject` we can pass as an owner. We could relax the type constraint to `Any` but I think this will have some negative performance effect (only a theory and not tested) and someone could more easily pass us something wrong.

Sources/NIOSSL/SSLCertificateExtensions.swift
```diff
@@ -0,0 +1,197 @@
...
+        self.storage = .init(borrowing: reference, owner: owner)
+        self.stackSize = CNIOBoringSSL_sk_X509_EXTENSION_num(reference)
+    }
+    
+    internal init(copyOf reference: OpaquePointer!) {
+        self.init(takeOwnershipOf: CNIOBoringSSL_sk_X509_EXTENSION_dup(reference))
+    }
+    
+    internal init() {
+        self.init(takeOwnershipOf: nil)
```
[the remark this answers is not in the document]
dnadoba: We need to support the empty case [because CNIOBoringSSL_X509_get0_extensions returns nil if no extensions are present](https://www.openssl.org/docs/man1.1.1/man3/X509_get0_extensions.html). The other option we have would be to make the `NIOSSLCertificate._extensions` optional which I don't like.

Sources/NIOSSL/SSLCertificateExtensions.swift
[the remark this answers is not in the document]
Lukasa: In this case we don't want it to be an IUO, right, we want it to be an actual optional.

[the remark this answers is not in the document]
dnadoba: fixed in https://github.com/apple/swift-nio-ssl/pull/359/commits/de44d779536fcbb48e0b4b698da05e9439f2b5a6

Sources/NIOSSL/SSLCertificateExtensions.swift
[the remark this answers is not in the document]
Lukasa: Do we need an empty `init` like this at `internal` scope?

Sources/NIOSSL/SSLCertificateExtensions.swift
```diff
@@ -0,0 +1,221 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+@_implementationOnly import CNIOBoringSSL
+@_implementationOnly import CNIOBoringSSLShims
+
+public struct _NIOSSLCertificateExtensions {
+    private enum Storage {
+        final class Deallocator {
+            let reference: OpaquePointer?
```
Lukasa, line 21: Can we explain the desire to support a `nil` pointer with a code comment here?

Sources/NIOSSL/SSLCertificateExtensions.swift
```diff
@@ -0,0 +1,221 @@
...
+            self.stackSize = 0
+        }
+    }
+    
+    internal init(copyOf reference: OpaquePointer?) {
+        guard let reference = reference else {
+            self.init()
+            return
+        }
+        self.init(takeOwnershipOf: CNIOBoringSSL_sk_X509_EXTENSION_dup(reference))
```
Lukasa, line 98: It is possible for this to return `nil`. I'm inclined to say we should trap in that case: getting a silent copy of nothing seems not great.

Lukasa: Actually do we use this at all? If not we should probably just drop it.

Sources/NIOSSL/SSLCertificateExtensions.swift
```diff
@@ -0,0 +1,221 @@
...
+        _NIOSSLCertificateExtensions(borrowing: CNIOBoringSSL_X509_get0_extensions(self._ref), owner: self)
+    }
+}
+
+extension _NIOSSLCertificateExtensions: RandomAccessCollection {
+    public subscript(position: Int) -> _NIOSSLCertificateExtension {
+        precondition(self.indices.contains(position), "index \(position) out of bounds")
+        return self.storage.withReference { reference in
+            guard let reference = reference else {
+                fatalError("unexpected null pointer when unwrapping reference for value at \(position)")
```
Lukasa, line 117: This is just a fancy force-unwrap, so we can use a force-unwrap and save the code.

Sources/NIOSSL/SSLCertificateExtensions.swift
```diff
@@ -0,0 +1,221 @@
...
+            }
+            guard let value = CNIOBoringSSLShims_sk_X509_EXTENSION_value(reference, position) else {
```
Lukasa, line 119: Same here.

Sources/NIOSSL/SSLCertificateExtensions.swift
```diff
@@ -0,0 +1,221 @@
...
+            
+            return .init(borrowing: value, owner: self.storage.owner)
+        }
+    }
+    
+    @inlinable public var startIndex: Int { 0 }
+    @inlinable public var endIndex: Int { self.stackSize }
+}
+
+public struct _NIOSSLCertificateExtension {
```
Lukasa, line 131: I'm wondering if this and `_NIOSSLCertificateExtensions` should be nested types within `NIOSSLCertificate`.

Lukasa (approved):

Issue #230, opened 2020-07-02
Want to enable Pre-Shared Key communication using SSL Service.


anil291987:
Hi,
 I am connecting with  my server with certificate and PSK communication. Able to connect with certificate with server. But not able to figure out how to do [PSK](https://en.wikipedia.org/wiki/Pre-shared_key) communication using [swift-nio-ssl.](https://github.com/apple/swift-nio-ssl) If you have required more information from my side. Please let me know.

Lukasa:
Yes, we should support this. Pre-Shared Keys are complex enough that we probably want to do this with a delegate protocol for users to implement. The trickiest part of this will be arranging for the PSKs to be able to secure stored on the heap. For that use-case, we probably need to bring over [`SecureBytes`](https://github.com/apple/swift-crypto/blob/master/Sources/Crypto/Util/SecureBytes.swift) from `swift-crypto`.

The delegate would be something like:

```swift
public protocol NIOSSLPreSharedKeyCache {
    func preSharedKey(identity: [UInt8]) -> SecureBytes
}
```

Once we have that, wiring up the rest of the interface should be easy enough, I think.

However, this is not a terribly high priority feature for the NIO core team, so we are unlikely to rush to implement this feature. It's on our roadmap, we'll keep this issue open, but if anyone else feels like making the contribution they should go for it.

Lukasa:
As a breadcrumb, if you're only running your application on Apple platforms, [SwiftNIO Transport Services](https://github.com/apple/swift-nio-transport-services) gives you access to Network.framework with SwiftNIO, which already has PSK support.

Lukasa:
Resolved using #369.

PR #360, opened 2022-04-29
Add TLS-PSK Support

agnosticdev:
Add TLS-PSK Support

Motivation:

Add TLS-PSK Support to address #230.

Modifications:

This is a large PR so I can understand if it needs to be broken up.
This PR includes the addition of SecureBytes from SwiftCrypto, to which I had to alter to make work.
To add SecureBytes I also had to rearrange the Package.swift file based on what I seen from the SwiftCrypto project.  Please let me know if this is correct.
I altered the access level on SecureBytes to be used in the return value for the callback.  Please let me know your thoughts on this.

Result:

TLS-PSK support for TLS 1.2. TLS 1.3 is not addressed due to API support in BoringSSL.


agnosticdev:
It looks like I broke the soundness check.  Do I just need to add the `testThrowingInitialization` back to the `SecureBytesTests`?
```
+                ("testThrowingInitialization", testThrowingInitialization),
```

Tests/NIOSSLTests/SecureBytesTests.swift
```diff
@@ -0,0 +1,228 @@
...
+        }
+
+        XCTAssertGreaterThanOrEqual(base.backing.capacity, 8)
+        XCTAssertEqual(Array(base), [1, 2, 3, 4])
+
+        func testThrowingInitialization() throws {
+            _ = try SecureBytes(unsafeUninitializedCapacity: 5) { (_, _) in
+                throw CryptoKitError.incorrectKeySize
+            }
+        }
```
dnadoba, line 174: `scripts/generate_linux_tests.rb` doesn't seem to handle inline functions correctly. I'm not quite sure how much work it would be to correctly detect inline functions an skip them during.
For now, you should be able to just rewrite your inline function as a closure:
```suggestion
        let testThrowingInitialization: () -> throws () = {
            _ = try SecureBytes(unsafeUninitializedCapacity: 5) { (_, _) in
                throw CryptoKitError.incorrectKeySize
            }
        }
```

agnosticdev: Thank you @dnadoba!  Looks like I have it all fixed up now in cd37a41

Package.swift
```diff
@@ -45,6 +45,20 @@ func generateDependencies() -> [Package.Dependency] {
...
+if development {
+    swiftSettings = [
+        .define("CRYPTO_IN_SWIFTPM"),
+        .define("CRYPTO_IN_SWIFTPM_FORCE_BUILD_API"),
+    ]
+} else {
+    swiftSettings = [
+        .define("CRYPTO_IN_SWIFTPM"),
+    ]
+}
```
Lukasa, line 60: None of this should be necessary: the Package.swift in swift-crypto exists to support the same code being part of CryptoKit and swift-crypto, which is not a problem that exists here.

Sources/NIOSSL/SwiftCrypto/CryptoKitErrors.swift
```diff
@@ -0,0 +1,60 @@
...
+    /// An ASN.1 field was truncated and could not be decoded.
+    case truncatedASN1Field
+
+    /// The encoding used for the field length is not supported.
+    case unsupportedFieldLength
+
+    /// It was not possible to parse a string as a PEM document.
+    case invalidPEMDocument
+}
+#endif
```
Lukasa, line 60: We don't need this file, we can use our own errors.

Sources/NIOSSL/SwiftCrypto/SafeCompare_boring.swift
```diff
@@ -0,0 +1,35 @@
...
+
+/// A straightforward constant-time comparison function for any two collections of bytes.
+@inlinable
+internal func constantTimeCompare<LHS: Collection, RHS: Collection>(_ lhs: LHS, _ rhs: RHS) -> Bool where LHS.Element == UInt8, RHS.Element == UInt8 {
+    guard lhs.count == rhs.count else {
+        return false
+    }
+
+    return zip(lhs, rhs).reduce(into: 0) { $0 |= $1.0 ^ $1.1 } == 0
+}
```
Lukasa, line 35: We can flatten this into a single `SafeCompare.swift` file above as the other code path doesn't exist.

Sources/NIOSSL/SwiftCrypto/SecureBytes.swift
```diff
@@ -0,0 +1,494 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+#if (os(macOS) || os(iOS) || os(watchOS) || os(tvOS)) && CRYPTO_IN_SWIFTPM && !CRYPTO_IN_SWIFTPM_FORCE_BUILD_API
+@_exported import CryptoKit
+#else
+import Foundation
+
+public struct SecureBytes {
```
Lukasa, line 19: This will need to be namespaced:

```suggestion
public struct NIOSSLSecureBytes {
```

Sources/NIOSSL/SwiftCrypto/SecureBytes.swift
```diff
@@ -0,0 +1,494 @@
...
+    public init() {
+        self = .init(count: 0)
+    }
+
+    @usableFromInline
+    init(count: Int) {
+        self.backing = SecureBytes.Backing.create(randomBytes: count)
+    }
+
+    init<D: ContiguousBytes>(bytes: D) {
```
Lukasa, line 33: We can't have this constructor: we must drop the `Foundation` dependency here.

Sources/NIOSSL/SwiftCrypto/SecureBytes.swift
```diff
@@ -0,0 +1,494 @@
...
+        } else {
+            // We have room, and a unique pointer. Ask the backing storage to shuffle around.
+            let offsetRange = subrange.lowerBound.offset..<subrange.upperBound.offset
+            self.backing.replaceSubrangeFittingWithinCapacity(offsetRange, with: newElements)
+        }
+    }
+}
+
+// MARK: - ContiguousBytes conformance
+extension SecureBytes: ContiguousBytes {
```
Lukasa, line 167: We'll have to drop this conformance too.

Sources/NIOSSL/SwiftCrypto/SecureBytes.swift
```diff
@@ -0,0 +1,494 @@
...
+        if !isKnownUniquelyReferenced(&self.backing) {
+            self.backing = Backing.create(copying: self.backing)
+        }
+
+        return try self.backing.withUnsafeMutableBytes(body)
+    }
+}
+
+// MARK: - DataProtocol conformance
+extension SecureBytes: DataProtocol {
```
Lukasa, line 184: And this one

Sources/NIOSSL/SwiftCrypto/SecureBytes.swift
```diff
@@ -0,0 +1,494 @@
...
+    @inlinable
+    public var regions: CollectionOfOne<SecureBytes> {
+        return CollectionOfOne(self)
+    }
+}
+
+// MARK: - MutableDataProtocol conformance
+extension SecureBytes: MutableDataProtocol { }
```
Lukasa, line 192: And this one

Sources/NIOSSL/SwiftCrypto/SecureBytes.swift
```diff
@@ -0,0 +1,494 @@
...
+        let byteRange = offset..<(offset + bytes.count)
+
+        self._withVeryUnsafeMutableBytes { backingPtr in
+            let dest = UnsafeMutableRawBufferPointer(rebasing: backingPtr[byteRange])
+            dest.copyBytes(from: bytes)
+        }
+    }
+}
+
+extension SecureBytes.Backing: ContiguousBytes {
```
Lukasa, line 392: And this conformance.

Sources/NIOSSL/SwiftCrypto/SecureBytes.swift
```diff
@@ -0,0 +1,494 @@
...
+        n |= n >> 16
+        if n != max {
+            n += 1
+        }
+
+        return n
+    }
+}
+
+extension Data {
```
Lukasa, line 454: And this extension.

Lukasa (changes requested):
Thanks for this! As you proposed splitting the PR I think that's a good idea. The most natural first split is to land the `SecureBytes` port as a separate patch, by itself. That will remove most of the lines of code, and let us more thoroughly scrutinise how we want to bring that across.

agnosticdev:
Thank you for the feedback @Lukasa.  I will start on the migration of `SecureBytes` to `NIOSSLSecureBytes` and will close this PR once that is ready for review.

agnosticdev:
Closing to submit a new PR for `NIOSSLSecureBytes` as a first step here.

PR #361, opened 2022-05-02
support additional certificate validation

dnadoba:
### Motivation
A user should be able to inspect a certificate chain after default verification and accept or reject that connection asynchronously. We do not yet have access to the complete verified certificate chain in SwiftNIO SSL but we can already prepare the state machine in `NIOSSLHandler` to support an additional verification  step.

### Modification
- add `additionalVerification` state to `ConnectionState`
- add `additionalCertificateChainVerification` callback to `NIOSSLContext`

### Result
`NIOSSLHandler` supports an asynchronous user defined verification step after default verification that delays connection establishment until success. Everything is still internal as we will need to alter the type signature of `NIOSSLContext. CertificateChainVerificationCallback` once we have access to the verified certificate chain.


Sources/NIOSSL/SSLContext.swift
```diff
@@ -123,14 +123,20 @@ private func alpnCallback(ssl: OpaquePointer?,
 ///
 /// - Warning: Avoid creating `NIOSSLContext`s on any `EventLoop` because it does _blocking disk I/O_.
 public final class NIOSSLContext {
+    internal typealias CertificateChainVerificationCallback = () -> EventLoopFuture<Void>
```
Lukasa, line 126: Can we add the word additional here?

```suggestion
    internal typealias AdditionalCertificateChainVerificationCallback = () -> EventLoopFuture<Void>
```

Lukasa: Also, it's probably a good idea for us to try to pass the `Channel` in to this callback, so that users have access to both the correct `EventLoop` _and_ to a way to find out what the context of the callback is.

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -266,6 +267,43 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
...
+    }
+    
+    private func completedAdditionalCertificateChainVerification(result: Result<Void, Error>, context: ChannelHandlerContext) {
+        context.eventLoop.preconditionInEventLoop()
+        switch self.state {
+        case .idle, .handshaking, .active:
+            preconditionFailure("invalid state \(self.state)")
+        case .additionalVerification:
+            do {
+                try result.get()
```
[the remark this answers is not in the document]
dnadoba: I have now used a `switch` but can change it to a `if/gard case` if you still want.

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -241,19 +249,19 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
                 channelClose(context: context, reason: error)
                 return
             }
+            
+            if let additionalCertificateChainVerification = connection.parentContext.additionalCertificateChainVerification {
+                state = .additionalVerification
+                additionalCertificateChainVerification(context.channel)
+                    .hop(to: context.eventLoop)
+                    .whenComplete { result in
+                        self.completedAdditionalCertificateChainVerification(result: result, context: context)
```
Lukasa, line 258: As we store the `context` on `self`, we don't need to close over it here, which will make this closure "free" to allocate (as it will only close over a single `class`).

Lukasa (commented):
Great, one minor change and we're good to go!

Lukasa (approved):

PR #362, opened 2022-05-02
Flush only after handshake

dnadoba:
### Motivation
We have a new custom certification chain validation step after the normal BoringSSL certificate chain validation that BoringSSL is not aware of. Therefore we can no longer rely on BoringSSLs rejection of writes and need to delay writes until the complete handshake is complete. 

### Modification
- ignore calls to `flush(context:)` during handshaking
- correctly flush everything after successful handshake

### Result
Writes are correctly buffered during custom certification chain validation.


Lukasa (approved):
Nice one, great patch @dnadoba

PR #366, opened 2022-05-06
Port SecureBytes from Swift Crypto to use with TLS-PSK

agnosticdev:
Motivation:

Porting `SecureBytes` to `NIOSSLSecureBytes` for using in (Add TLS-PSK Support #360).

Modifications:

Addition and modification of `SecureBytes` to work within SwiftNIO SSL.
Added additional test cases.

Result:

`NIOSSLSecureBytes` will be ported for usage in a later PR with TLS-PSK.


Sources/NIOSSL/SSLErrors.swift
```diff
@@ -157,6 +157,12 @@ internal extension BoringSSLError {
     }
 }
 
+public enum NIOSSLSecureBytesError: Error {
```
Lukasa, line 160: We only use this in the tests: can we just define it there instead of adding this as new API?

Sources/NIOSSL/SwiftCrypto/NIOSSLSecureBytes.swift
```diff
@@ -0,0 +1,429 @@
...
+        let newBacking = Backing.create(capacity: n)
+        newBacking._appendBytes(self.backing, inRange: 0..<self.count)
+        self.backing = newBacking
+    }
+}
+
+// MARK: - Equatable conformance, constant-time
+extension NIOSSLSecureBytes: Equatable {
+    public static func == (lhs: NIOSSLSecureBytes, rhs: NIOSSLSecureBytes) -> Bool {
+        //return safeCompare(lhs, rhs)
```
Lukasa, line 75: We can remove the commented out code.

Sources/NIOSSL/SwiftCrypto/NIOSSLSecureBytes.swift
```diff
@@ -0,0 +1,429 @@
...
+        }
+    }
+
+    @inlinable
+    public func index(after index: Index) -> Index {
+        return index.advanced(by: 1)
+    }
+    
+    @inlinable
+    public var regions: CollectionOfOne<NIOSSLSecureBytes> {
```
Lukasa, line 126: We don't need this: this is part of the `DataProtocol` conformance we don't have anymore.

Sources/NIOSSL/SwiftCrypto/NIOSSLSecureBytes.swift
```diff
@@ -0,0 +1,429 @@
...
+                constantTimeCompare(lhsPtr, rhsPtr)
+            }
+        }
+    }
+}
+
+// MARK: - Collection conformance
+extension NIOSSLSecureBytes: Collection {
+    
+    public struct Index {
```
Lukasa, line 87: I'm inclined to want us to diverge from Crypto here a bit: while the opaque index seemed like a good idea at the time I don't think it's serving us here.

We can delete `struct Index` and use an `Int` as the `Index` directly. Our `startIndex` and `endIndex` stay the same values, they just stop being wrapped in `struct Index`. This will also let us remove our `Strideable` and `Hashable` conformances for `struct Index` and also remove `func index(after:)` and `func index(before:)`, and lets us collapse our `Collection`/`BidirectionalCollection`/`RandomAccessCollection` conformances into a single conformance to `RandomAccessCollection`.

Sources/NIOSSL/SwiftCrypto/RNG_boring.swift
```diff
@@ -0,0 +1,40 @@
+//===----------------------------------------------------------------------===//
```
Lukasa, line 1: We can rename this file to just `RNG`.

Sources/NIOSSL/SwiftCrypto/Zeroization_boring.swift
```diff
@@ -0,0 +1,26 @@
+//===----------------------------------------------------------------------===//
```
Lukasa, line 1: We can just call this `Zeroization` and drop the `_boring`.

Lukasa (commented):
Great, looks really good! A few notes in the diff, mostly to simplify things.

Sources/NIOSSL/SwiftCrypto/Zeroization_boring.swift
agnosticdev: Done in 8007b79

Sources/NIOSSL/SwiftCrypto/RNG_boring.swift
agnosticdev: Done in 8007b79

Sources/NIOSSL/SwiftCrypto/NIOSSLSecureBytes.swift
agnosticdev: Done in 8007b79

Sources/NIOSSL/SSLErrors.swift
agnosticdev: Forgot to add this as part of 8007b79, coming right up.

agnosticdev: Done

agnosticdev:
Ping @Lukasa for review feedback on this one.

Sources/NIOSSL/SwiftCrypto/NIOSSLSecureBytes.swift
```diff
@@ -0,0 +1,375 @@
...
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+
+public struct NIOSSLSecureBytes {
```
Lukasa, line 16: Can we begin by removing all the `public` access qualifiers from this type and then add them in as we need in follow-up PRs? I'd like us to avoid exposing too much public API unnecessarily.

Lukasa (commented):
Broadly looks good to me, one minor tweak to the public API surface introduced by this PR and we're golden.

agnosticdev:
@Lukasa do you want me to rebase the updates from https://github.com/apple/swift-nio-ssl/pull/368 ?

Lukasa (approved):
Nice one, LGTM.

agnosticdev:
Thank you!

Issue #367, opened 2022-05-09
Tests don't succeed when running in parallel


Davidde94:
The same scratch file is used in multiple tests, which introduces a race condition where one test may try to write to a file that's just been "cleaned up" by a concurrently-running test.

Is this something we care about? Possible/probably not, but worth checking regardless. 

Caught by the release tools.

```
Test Suite 'Selected tests' started at 2022-05-09 14:11:12.061Test Suite 'swift-nio-sslPackageTests.xctest' started at 2022-05-09 14:11:12.063Test Suite 'TLSConfigurationTest' started at 2022-05-09 14:11:12.063Test Case '-[NIOSSLTests.TLSConfigurationTest testRehashFormat]' started./private/tmp/nio-release-validator/C9DDB4EC-7CDC-4190-AC23-F47E15A7934B/repos/swift-nio-ssl/Tests/NIOSSLTests/TLSConfigurationTest.swift:711: error: -[NIOSSLTests.TLSConfigurationTest testRehashFormat] : XCTAssertNoThrow failed: threw error "Error Domain=NSCocoaErrorDomain Code=516 "The file “7f44456a.0” couldn’t be saved in the folder “T” because a file with the same name already exists." UserInfo={NSFilePath=/var/folders/d7/t2tw0mc97b557jnn0hv9gflm0000gn/T/7f44456a.0, NSUnderlyingError=0x6000018fc090 {Error Domain=NSPOSIXErrorDomain Code=17 "File exists"}}"/private/tmp/nio-release-validator/C9DDB4EC-7CDC-4190-AC23-F47E15A7934B/repos/swift-nio-ssl/Tests/NIOSSLTests/TLSConfigurationTest.swift:716: error: -[NIOSSLTests.TLSConfigurationTest testRehashFormat] : XCTAssertNoThrow failed: threw error "Error Domain=NSCocoaErrorDomain Code=4 "“7f44456a.0” couldn’t be removed." UserInfo={NSUserStringVariant=(
    Remove
[...]

PR #368, opened 2022-05-11
Parallel test fix for TLSConfigurationTest

agnosticdev:
Motivation:

Address the issue the race condition brought up in #367

Modifications:

Created directories based on the test name so that each file would be isolated.

Result:

Avoid race condition in parallel test.

### DIFF
--- a/Tests/NIOSSLTests/SSLCertificateTest.swift
+++ b/Tests/NIOSSLTests/SSLCertificateTest.swift
@@ -142,8 +142,14 @@ kVuVyNH7NBMh6YOuTL1dh55bvDjvgkuzudepsZnpfjgQKE1aZ7dL32Xi000gBM8=
 -----END CERTIFICATE-----
 """
 
-func makeTemporaryFile(fileExtension: String = "") -> String {
-    let template = "\(FileManager.default.temporaryDirectory.path)/niotestXXXXXXX\(fileExtension)"
+func makeTemporaryFile(fileExtension: String = "", customPath: String = "") throws -> String {
+    var template = "\(FileManager.default.temporaryDirectory.path)/niotestXXXXXXX\(fileExtension)"
+    // If a custom file path is passed in then a new directory has to also be created.  Then the file can be written to that directory.
+    if !customPath.isEmpty {
+        let path = "\(FileManager.default.temporaryDirectory.path)/\(customPath)/"
+        try FileManager.default.createDirectory(at: URL(fileURLWithPath: path), withIntermediateDirectories: true)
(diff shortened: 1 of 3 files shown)


Tests/NIOSSLTests/SSLCertificateTest.swift
```diff
@@ -161,6 +161,27 @@ internal func dumpToFile(data: Data, fileExtension: String = "") throws  -> Stri
     return filename
 }
 
+internal func dumpToFileWithCustomPath(data: Data, fileExtension: String = "", customPath: String = "") throws  -> String {
```
glbrntt, line 164: Can we just update `makeTemporaryFile` and `dumpToFile` to have a `customPath` or am I missing something?

Tests/NIOSSLTests/SSLCertificateTest.swift
```diff
@@ -161,6 +161,27 @@ internal func dumpToFile(data: Data, fileExtension: String = "") throws  -> Stri
...
+    var template = "\(FileManager.default.temporaryDirectory.path)/niotestXXXXXXX\(fileExtension)"
+    // If a custom file path is passed in then a new directory has to also be created.  Then the file can be written to that directory.
+    if !customPath.isEmpty {
+        let path = "\(FileManager.default.temporaryDirectory.path)/\(customPath)/"
+        try FileManager.default.createDirectory(at: URL(string: "file://"  + path)!, withIntermediateDirectories: true)
```
glbrntt, line 169:
```suggestion
        try FileManager.default.createDirectory(at: URL(fileURLWithPath: path)!, withIntermediateDirectories: true)
```

glbrntt:
(Also thanks for the patch!)

Tests/NIOSSLTests/SSLCertificateTest.swift
agnosticdev: Addressed in [5d546a8](https://github.com/apple/swift-nio-ssl/pull/368/commits/5d546a8f830aab00132d33f2c210d20ed0b26713)

Tests/NIOSSLTests/SSLCertificateTest.swift
agnosticdev: Nope, you did not miss anything, we can go that route too.

agnosticdev: Addressed in [5d546a8](https://github.com/apple/swift-nio-ssl/pull/368/commits/5d546a8f830aab00132d33f2c210d20ed0b26713)

glbrntt (approved):
Great, thanks for fixing this @agnosticdev

agnosticdev:
> Great, thanks for fixing this @agnosticdev

No problem at all!

PR #369, opened 2022-05-20
Add TLS-PSK Support

agnosticdev:
Motivation:

Add TLS-PSK Support to address #230.

Modifications:

Modified SSLContext to take a passed in PSK callback to perform key and identity operations during the handshake.

Result:

TLS-PSK support for TLS 1.2. TLS 1.3 is not addressed due to API support in BoringSSL.


agnosticdev:
@Lukasa for review.  Thank you!

Sources/NIOSSL/SSLCallbacks.swift
```diff
@@ -130,6 +130,18 @@ extension KeyLogCallbackManager {
     }
 }
 
+/// PSK Identity Callback Manager
+public typealias NIOPSKIdentityCallback = () -> (key: NIOSSLSecureBytes, identity: NIOSSLSecureBytes)
+
+
+/// An object that provides helpers for clients / servers wanting to expose PSKs and Identities to their context.
+public struct PSKIdentityCallbackManager {
```
Lukasa, line 138: Does this type need to be `public`?

Sources/NIOSSL/SSLCallbacks.swift
```diff
@@ -130,6 +130,18 @@ extension KeyLogCallbackManager {
     }
 }
 
+/// PSK Identity Callback Manager
+public typealias NIOPSKIdentityCallback = () -> (key: NIOSSLSecureBytes, identity: NIOSSLSecureBytes)
```
Lukasa, line 134: I think we need two callbacks, not one. The server is supposed to have access to the identity asserted by the peer so it can choose. Additionally, the client should be given an identity hint. Finally, we should probably let the server be able to configure an identity hint as well.

Sources/NIOSSL/SwiftCrypto/NIOSSLSecureBytes.swift
```diff
@@ -262,7 +262,7 @@ extension NIOSSLSecureBytes.Backing {
 
     /// Appends the bytes of a slice of another backing buffer to this storage, crashing if there
     /// is not enough room.
-    /* private but inlinable */ func _appendBytes(_ backing: NIOSSLSecureBytes.Backing, inRange range: Range<Int>) {
+    /* private but inlinable */ public func _appendBytes(_ backing: NIOSSLSecureBytes.Backing, inRange range: Range<Int>) {
```
Lukasa, line 265: This should be `@inlinable`, not public,

Sources/NIOSSL/SwiftCrypto/NIOSSLSecureBytes.swift
```diff
@@ -247,7 +247,7 @@ extension NIOSSLSecureBytes.Backing {
     }
 
     /// Appends the bytes of a collection to this storage, crashing if there is not enough room.
-    /* private but inlinable */ func _appendBytes<C: Collection>(_ bytes: C) where C.Element == UInt8 {
+    /* private but inlinable */ public func _appendBytes<C: Collection>(_ bytes: C) where C.Element == UInt8 {
```
Lukasa, line 250: This should be `@inlinable`, not public.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -321,7 +331,9 @@ public struct TLSConfiguration {
                  keyLogCallback: NIOSSLKeyLogCallback?,
                  renegotiationSupport: NIORenegotiationSupport,
                  additionalTrustRoots: [NIOSSLAdditionalTrustRoots],
-                 sendCANameList: Bool = false) {
+                 sendCANameList: Bool = false,
+                 pskCallback: PSKIdentityCallbackManager? = nil,
+                 client: Bool = false) {
```
Lukasa, line 336: What is the `client` argument for?

Lukasa (changes requested):
Great start! I have some suggestions about an alternative implementation direction.

Sources/NIOSSL/TLSConfiguration.swift
agnosticdev: The `client` argument is being used to delineate between how the `NIOSSLContext` sets the callback for the particular context, e.g., `CNIOBoringSSL_SSL_CTX_set_psk_client_callback` or `CNIOBoringSSL_SSL_CTX_set_psk_server_callback`.

Now, above you mentioned:
> I think we need two callbacks, not one. The server is supposed to have access to the identity asserted by the peer so it can choose. Additionally, the client should be given an identity hint.

If you wanted to go in the direction of adding a client or server based callback to `TLSConfiguration` e.g., `NIOPSKServerCallback` and `NIOPSKClientCallback`, then this `client` flag can go away entirely.

Thoughts @Lukasa ?

agnosticdev: I still have some work left on this PR, but I split up the callbacks into a client and server side callback and put them both under `PSKIdentityCallbackManager` in [7742aa0](https://github.com/apple/swift-nio-ssl/pull/369/commits/7742aa0cc15f7b191af71578fd2227c7041a7cba).  This removes the `client` flag also.  Wanted to get your thoughts on this @Lukasa ?

Sources/NIOSSL/SwiftCrypto/NIOSSLSecureBytes.swift
agnosticdev: resolved in [7742aa0](https://github.com/apple/swift-nio-ssl/pull/369/commits/7742aa0cc15f7b191af71578fd2227c7041a7cba)

Sources/NIOSSL/SwiftCrypto/NIOSSLSecureBytes.swift
agnosticdev: resolved in [7742aa0](https://github.com/apple/swift-nio-ssl/pull/369/commits/7742aa0cc15f7b191af71578fd2227c7041a7cba)

Sources/NIOSSL/SSLCallbacks.swift
agnosticdev: I still have work here, but the callbacks are split now in [7742aa0](https://github.com/apple/swift-nio-ssl/pull/369/commits/7742aa0cc15f7b191af71578fd2227c7041a7cba).

agnosticdev: I added two different callbacks here on 943aab0. I also tried to address the following: 
1. Send the hint and possibly identity to callback in the`serverPSKCallback`.
2. Configure a hint for the client
3. Configure a hint for server.

agnosticdev:
Looks like I have some null-terminating string issues on for 5.7.  I can look at how to address this.

Sources/NIOSSL/SSLCallbacks.swift
```diff
@@ -130,6 +130,22 @@ extension KeyLogCallbackManager {
     }
 }
 
+/// PSK Identity Callback Manager
+public typealias NIOPSKClientIdentityCallback = (NIOSSLSecureBytes?) -> (key: NIOSSLSecureBytes, identity: NIOSSLSecureBytes)
+public typealias NIOPSKServerIdentityCallback = (NIOSSLSecureBytes?, NIOSSLSecureBytes?) -> (key: NIOSSLSecureBytes, identity: NIOSSLSecureBytes)
```
Lukasa, line 135: Relatedly, only the keys need to be in `NIOSSLSecureBytes`: the rest can be regular `String` objects.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -112,6 +112,119 @@ private func alpnCallback(ssl: OpaquePointer?,
...
+    
+    // Take the hint and the possible identity and pass them down to the callback to get associated PSK
+    let contextPSKIdentityCallback = serverCallback(clientHint, incomingPSKIndentity)
+    let contextPSK = contextPSKIdentityCallback.key // From the context
+    let contextPSKIdentity = contextPSKIdentityCallback.identity // From the context
+    
+    // Equate both identities above to make sure both are the same (they should be as the context was copied into the incoming identity).
+    if contextPSKIdentity != incomingPSKIndentity {
+        // Warn about this here but do not fail.
+        print("PSK warning: client identity not what we expected (got \(Array(incomingPSKIndentity)) expected \(Array(contextPSKIdentity))")
```
Lukasa, line 164: We shouldn't be making any assertions about what the client's identity is, we're not the boss of the client. If we didn't like the identity the client presented we should refuse to return a key.

Sources/NIOSSL/SSLCallbacks.swift
Lukasa, line 135: These callbacks need error reporting paths: they should be allowed to throw.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -112,6 +112,119 @@ private func alpnCallback(ssl: OpaquePointer?,
...
+    // Initial implementation only supports TLS 1.2 due API support exposed from BoringSSL.
+    // TODO (meaton) add TLS 1.3 support when available.
+    
+    // We want to take the SSL pointer and extract the parent Swift object.
+    let parentCtx = CNIOBoringSSL_SSL_get_SSL_CTX(ssl)!
+    let parentPtr = CNIOBoringSSLShims_SSL_CTX_get_app_data(parentCtx)!
+    let parentSwiftContext: NIOSSLContext = Unmanaged.fromOpaque(parentPtr).takeUnretainedValue()
+    
+    guard let clientCallback = parentSwiftContext.pskIdentityCallbackManager?.clientCallback,
+          let incomingIdentity = identity, // Incoming PSK identity. (Could be empty)
```
Lukasa, line 191: This comment isn't right: `identity` is an _out_ pointer and may not be `nil`.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -112,6 +112,119 @@ private func alpnCallback(ssl: OpaquePointer?,
...
+          let incomingPSK = psk // Incoming PSK key.
```
Lukasa, line 192: Again, this isn't true, the PSK here is an _output_.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -112,6 +112,119 @@ private func alpnCallback(ssl: OpaquePointer?,
...
+          else {
+        return 0
+    }
+    var incomingPSKIndentity = NIOSSLSecureBytes()
+    let strBasedIncomingIdentity = String(validatingUTF8: incomingIdentity) ?? ""
```
Lukasa, line 197: This is UB: there is no guarantee this pointer has valid bytes inside it. We're supposed to write to this, not to read it.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -112,6 +112,119 @@ private func alpnCallback(ssl: OpaquePointer?,
...
+
+    // If set, build out a server hint to pass into the client callback.
+    var serverHint = NIOSSLSecureBytes()
```
Lukasa, line 200: We can just leave this as a `String`, it isn't secret.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -112,6 +112,119 @@ private func alpnCallback(ssl: OpaquePointer?,
...
+    if let strServerHint = hint,
+       let derivedHint = String(validatingUTF8: strServerHint) {
+        serverHint.append(derivedHint.utf8)
+    }
+    
+    let contextPSKIdentityCallback = clientCallback(serverHint)
+    let contextPSK = contextPSKIdentityCallback.key // From the context
+    let contextPSKIdentity = contextPSKIdentityCallback.identity // From the context
+    
+    // If the server is sending the identity, store it first as incomingPSKIndentity and do not fail because of this.
```
Lukasa, line 210: The server never sends the identity.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -112,6 +112,119 @@ private func alpnCallback(ssl: OpaquePointer?,
...
+    if strBasedIncomingIdentity.count > 0 {
+        incomingPSKIndentity.append(strBasedIncomingIdentity.utf8)
+    } else if let sslIdentity = CNIOBoringSSL_SSL_get_psk_identity(ssl),
```
Lukasa, line 213: This branch is also not relevant: the only identity that matters is the one the user returned.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -112,6 +112,119 @@ private func alpnCallback(ssl: OpaquePointer?,
...
+              let sslIdentityStr = String(validatingUTF8: sslIdentity) {
+        incomingPSKIndentity.append(sslIdentityStr.utf8)
+    }
+    // Equate both identities above to make sure both are the same (they should be as the context was copied into the incoming identity).
+    if contextPSKIdentity != incomingPSKIndentity {
+        print("PSK warning: server identity not what we expected (got \(Array(incomingPSKIndentity)) expected \(Array(contextPSKIdentity))")
+    }
+    if max_psk_len > UInt32.max || contextPSK.count > max_psk_len {
```
Lukasa, line 221: Same notes on this check.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -173,6 +287,25 @@ public final class NIOSSLContext {
...
+            if let clientSideCallback = pskCallback.clientCallback {
+                let defaultClientPSKInfo = clientSideCallback(nil)
+                let indentityBytes = defaultClientPSKInfo.identity.backing.withUnsafeBytes{  Array($0) }
+                CNIOBoringSSL_SSL_CTX_use_psk_identity_hint(context, String(decoding: indentityBytes, as: Unicode.UTF8.self))
+                CNIOBoringSSL_SSL_CTX_set_psk_client_callback(context, { clientPSKCallback(ssl: $0, hint: $1, identity: $2, max_identity_len: $3, psk: $4, max_psk_len: $5 ) } )
+            // Ohterwise see if the server side callback is present.
+            } else if let serverSideCallback = pskCallback.serverCallback {
+                let defaultServerPSKInfo = serverSideCallback(nil, nil) // nil, nil being passed in as a way to pass back default data.
+                let indentityBytes = defaultServerPSKInfo.identity.backing.withUnsafeBytes{  Array($0) }
+                CNIOBoringSSL_SSL_CTX_use_psk_identity_hint(context, String(decoding: indentityBytes, as: Unicode.UTF8.self))
```
Lukasa, line 305: This isn't great: we should almost certainly just let the PSK hint be configured on the `TLSConfiguration` directly.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -276,6 +280,9 @@ public struct TLSConfiguration {
 
     /// The private key associated with the leaf certificate.
     public var privateKey: NIOSSLPrivateKeySource?
+    
+    /// PSK Callback to get the key and identity from either the client or server.
+    public var pskCallback: PSKIdentityCallbackManager? = nil
```
Lukasa, line 285: We want to let users configure the callbacks, not the manager. In this instance we're happy to let them configure both at once.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -446,7 +456,34 @@ extension TLSConfiguration {
...
+                                pskCallback: nil)
+    }
+    
+    /// Create a TLS configuration for use with server-side or client-side contexts that uses Pre-Shared Keys for TLS 1.2 and below.
+    ///
+    /// This provides sensible defaults while requiring that you provide any data that is necessary
+    /// for server-side or client-side functionality.  This configuration uses Pre-Shared Keys instead of certificates.
+    ///
+    /// For customising fields, modify the returned TLSConfiguration object.
+    public static func makePreSharedKeyConfiguration() -> TLSConfiguration {
```
Lukasa, line 469: What value are we expecting users to get from this?

Sources/NIOTLSServer/main.swift
```diff
@@ -29,6 +30,17 @@ private final class EchoHandler: ChannelInboundHandler {
     }
 }
 
+
+func pskCallback(hint: NIOSSLSecureBytes?, identity: NIOSSLSecureBytes?) -> (key: NIOSSLSecureBytes, identity: NIOSSLSecureBytes) {
+    var psk = NIOSSLSecureBytes()
+    var pskIdentity = NIOSSLSecureBytes()
+    psk.append(contentsOf: Data("hello".utf8))
```
Lukasa, line 37: We don't need `Data` here, we can pass `hello.utf8` directly.

Lukasa (changes requested):
Thanks Matt! I've left some notes in the diff.

Sources/NIOSSL/SSLCallbacks.swift
```diff
@@ -130,6 +130,22 @@ extension KeyLogCallbackManager {
...
+
+
+/// An object that provides helpers for clients / servers wanting to expose PSKs to their context.
+public struct PSKIdentityCallbackManager {
```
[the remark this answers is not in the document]
agnosticdev: The reason I have left this as public and it's members as public is because when I want to consume these types from, let's say `NIOTLSServer` I get the following errors:
```
Cannot find 'PSKIdentityCallbackManager' in scope
```
or even:
```
'Some type'is inaccessible due to 'internal' protection level
```
With the following code just placed in `NIOTLSServer`:
```swift
func pskCallback(hint: NIOSSLSecureBytes?, identity: NIOSSLSecureBytes?) -> PSKIdentityResponse {
    var psk = NIOSSLSecureBytes()
    var pskIdentity = NIOSSLSecureBytes()
    psk.append(contentsOf: "hello".utf8)
    pskIdentity.append(contentsOf: "world".utf8)
    return PSKIdentityResponse(key: psk, identity: pskIdentity)
}

let clientPSKCallbackManager = PSKIdentityCallbackManager(clientCallback: nil, serverCallback: pskCallback)
```
Beings how and application like `NIOTLSServer` would be the consumer here of the PSK functionality, then I leave this block of code in to make sure these types work well here too.  
Thoughts?

Sources/NIOTLSServer/main.swift
agnosticdev: Addressed in 308299a

Sources/NIOSSL/SSLCallbacks.swift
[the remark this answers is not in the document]
agnosticdev: Addressed in 308299a

agnosticdev:
Note that the latest commit for 308299a is only a WIP partial.  I will be addressing more bit by bit.

Sources/NIOSSL/SSLCallbacks.swift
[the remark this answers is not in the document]
Lukasa: My proposal is that instead of saving a `PSKIdentityCallbackManager` on the `TLSConfiguration` we have two variables, one with each callback.

Sources/NIOSSL/SSLContext.swift
agnosticdev: Resolved in 409e1c1

Sources/NIOSSL/SSLContext.swift
agnosticdev: Resolved in 409e1c1

Sources/NIOSSL/SSLContext.swift
```diff
@@ -112,6 +112,119 @@ private func alpnCallback(ssl: OpaquePointer?,
...
+    var clientHint = NIOSSLSecureBytes()
+    if let strClientHint = hint,
+       let derivedHint = String(validatingUTF8: strClientHint) {
+        clientHint.append(derivedHint.utf8)
+    }
+
+    // Attempt to get the identity from the incoming argument or from the SSL OpaquePointer
+    if strBasedIncomingIdentity.count > 0 {
+        incomingPSKIndentity.append(strBasedIncomingIdentity.utf8)
+    } else if let sslIdentity = CNIOBoringSSL_SSL_get_psk_identity(ssl),
```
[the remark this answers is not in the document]
agnosticdev: Resolved in 409e1c1

Sources/NIOSSL/SSLCallbacks.swift
agnosticdev: Resolved in 409e1c1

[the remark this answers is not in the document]
agnosticdev: Resolved in 409e1c1

Sources/NIOSSL/SSLContext.swift
```diff
@@ -112,6 +112,119 @@ private func alpnCallback(ssl: OpaquePointer?,
...
+    let contextPSKIdentityCallback = serverCallback(clientHint, incomingPSKIndentity)
+    let contextPSK = contextPSKIdentityCallback.key // From the context
+    let contextPSKIdentity = contextPSKIdentityCallback.identity // From the context
+    
+    // Equate both identities above to make sure both are the same (they should be as the context was copied into the incoming identity).
+    if contextPSKIdentity != incomingPSKIndentity {
+        // Warn about this here but do not fail.
+        print("PSK warning: client identity not what we expected (got \(Array(incomingPSKIndentity)) expected \(Array(contextPSKIdentity))")
+    }
+    if max_psk_len > UInt32.max || contextPSK.count > max_psk_len {
```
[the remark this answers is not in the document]
agnosticdev: Resolved in 409e1c1

Sources/NIOSSL/SSLContext.swift
agnosticdev: Resolved in 409e1c1

Sources/NIOSSL/SSLContext.swift
agnosticdev: Resolved in 409e1c1

Sources/NIOSSL/SSLContext.swift
agnosticdev: Resolved in 409e1c1

Sources/NIOSSL/SSLContext.swift
agnosticdev: Resolved in 409e1c1

Sources/NIOSSL/SSLContext.swift
agnosticdev: Resolved in 409e1c1

Sources/NIOSSL/SSLContext.swift
agnosticdev: Resolved in 409e1c1

Sources/NIOSSL/SSLContext.swift
```diff
@@ -112,6 +112,119 @@ private func alpnCallback(ssl: OpaquePointer?,
...
+        return 0
+    }
+    memcpy(incomingPSK, Array(contextPSK), contextPSK.count)
```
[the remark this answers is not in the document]
agnosticdev: Great point.  Addressed in e3c15ef

Sources/NIOSSL/SSLCallbacks.swift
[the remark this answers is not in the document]
agnosticdev: @Lukasa I wanted to ask about this before I proceeded. Did to take:  
```swift
public typealias NIOPSKClientIdentityCallback = (String, String) -> PSKIdentityResponse
public typealias NIOPSKServerIdentityCallback = (String, String) -> PSKIdentityResponse
```
And add them both as options to configure on `TLSConfiguration` struct?  So `TLSConfiguration` would then have:
```swift
public var pskClientCallback: NIOPSKClientIdentityCallback? = nil

public var pskServerCallback: NIOPSKServerIdentityCallback? = nil
```
Let me know your thoughts?

agnosticdev:
Another WIP commit for [e3c15ef](https://github.com/apple/swift-nio-ssl/pull/369/commits/e3c15efc494826101d76890ef6b51f724d1ec8d0). Before I addressed the rest of the feedback I wanted to double check my understanding of the `PSKIdentityCallbackManager` feedback.

Sources/NIOSSL/TLSConfiguration.swift
Lukasa: Gotcha.

Sources/NIOSSL/TLSConfiguration.swift
agnosticdev: Resolved in 516a89f.

Sources/NIOSSL/SSLCallbacks.swift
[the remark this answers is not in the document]
agnosticdev: Thank you. Resolved in 516a89f.

Sources/NIOSSL/SSLCallbacks.swift
agnosticdev: Resolved in 516a89f.

Lukasa (changes requested):

Sources/NIOSSL/SSLContext.swift
```diff
@@ -112,6 +112,101 @@ private func alpnCallback(ssl: OpaquePointer?,
...
+    
+    // Make sure the key is returned by the callback and it is of proper length, otherwise, fail.
+    if serverPSK.isEmpty || serverPSK.count > max_psk_len {
+        return 0
+    }
+    var pskBuffer: [UInt8] = []
+    serverPSK.backing.withUnsafeBytes {
+        pskBuffer.append(contentsOf: $0)
+    }
+    memcpy(outputPSK, pskBuffer, pskBuffer.count)
```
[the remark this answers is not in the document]
agnosticdev: `withUnsafeBytes` is on the `backing` of `NIOSSLSecureBytes`.  The reason I copied this into a byte array first is because if I do:  
```swift
 memcpy(outputPSK, serverPSK.backing.withUnsafeBytes, pskBuffer.count)
```
I get:
```
Cannot convert value of type '((UnsafeRawBufferPointer) throws -> T) throws -> T' to expected argument type 'UnsafeRawPointer?'
Generic parameter 'T' could not be inferred
```

Sources/NIOSSL/SSLCallbacks.swift
```diff
@@ -130,6 +130,17 @@ extension KeyLogCallbackManager {
...
+/// PSK Identity response type used in the callback.
+public struct PSKIdentityResponse {
+    public var key: NIOSSLSecureBytes
+    public init(key: NIOSSLSecureBytes) {
+        self.key = key
+    }
+}
+
+/// PSK Identity Callback Manager
+public typealias NIOPSKClientIdentityCallback = (String, String) throws -> PSKIdentityResponse
```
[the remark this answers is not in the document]
agnosticdev: Great point. Addressed in c709232

Sources/NIOSSL/SSLContext.swift
```diff
@@ -112,6 +112,101 @@ private func alpnCallback(ssl: OpaquePointer?,
...
+    guard let serverCallback = parentSwiftContext.pskServerConfigurationCallback,
+          let outputIdentity = identity, // Output PSK identity.
+          let strOutputIdentity = String(validatingUTF8: outputIdentity), // Map identity to str.
+          let outputPSK = psk // Output PSK key.
+          else {
+        return 0
+    }
+    
+    // Get PSK client identity hint based on the SSL connection
+    let hint = CNIOBoringSSL_SSL_get_psk_identity_hint(ssl)
```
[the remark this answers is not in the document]
agnosticdev: Addressed in c709232

Sources/NIOSSL/SSLContext.swift
```diff
@@ -112,6 +112,101 @@ private func alpnCallback(ssl: OpaquePointer?,
...
+    // TODO (meaton) add TLS 1.3 support when available.
+    
+    // We want to take the SSL pointer and extract the parent Swift object.
+    let parentCtx = CNIOBoringSSL_SSL_get_SSL_CTX(ssl)!
+    let parentPtr = CNIOBoringSSLShims_SSL_CTX_get_app_data(parentCtx)!
+    let parentSwiftContext: NIOSSLContext = Unmanaged.fromOpaque(parentPtr).takeUnretainedValue()
+    
+    guard let clientCallback = parentSwiftContext.pskClientConfigurationCallback,
+          let outputIdentity = identity, // Output PSK identity. (Could be empty)
+          let strOutputIdentity = String(validatingUTF8: outputIdentity), // Map identity to str.
```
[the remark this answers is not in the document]
agnosticdev: Addressed in c709232

Sources/NIOSSL/SSLContext.swift
```diff
@@ -112,6 +112,101 @@ private func alpnCallback(ssl: OpaquePointer?,
...
+        return 0
+    }
+    let serverPSK = contextPSKIdentityCallback.key // From the callback
+    
+    // Make sure the key is returned by the callback and it is of proper length, otherwise, fail.
+    if serverPSK.isEmpty || serverPSK.count > max_psk_len {
+        return 0
+    }
+    var pskBuffer: [UInt8] = []
+    serverPSK.backing.withUnsafeBytes {
```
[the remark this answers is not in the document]
Lukasa: Sure, but that means we should add a wrapper method on `NIOSSLSecureBytes` that calls through. Ideally `.backing` would be `private` to the type.

Sources/NIOSSL/SSLContext.swift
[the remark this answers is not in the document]
Lukasa: Sorry, what I meant is:

```swift
serverPSK.withUnsafeBytes {
    outputPSK, $0.baseAddress, $0.count
}
```

Sources/NIOSSL/SSLContext.swift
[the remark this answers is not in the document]
agnosticdev: Addressed in 732e883

Sources/NIOSSL/SSLContext.swift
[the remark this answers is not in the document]
agnosticdev: Addressed in 732e883

agnosticdev:
I reverted a change I made because this caused a breakage.  I am still in need of researching a technique for copying the output key using `withUnsafeBytes`.  I will update when I've made progress on this.

agnosticdev:
@Lukasa 23f6baa should have the latest changes in it and is ready for review again.

agnosticdev:
Updating my branch with the latest changes on `main`.

Sources/NIOSSL/SSLCallbacks.swift
```diff
@@ -130,6 +130,19 @@ extension KeyLogCallbackManager {
     }
 }
 
+/// PSK Identity response type used in the callback.
+public struct PSKIdentityResponse {
```
Lukasa, line 134: I think in this instance we may as well have two types, one for each callback to return. That way we don't have to deal with the question of what `identity` should be set to in the server callback.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -112,6 +112,101 @@ private func alpnCallback(ssl: OpaquePointer?,
...
+        return 0
+    }
+    let serverPSK = contextPSKIdentityCallback.key // From the callback
+    
+    // Make sure the key is returned by the callback and it is of proper length, otherwise, fail.
+    if serverPSK.isEmpty || serverPSK.count > max_psk_len {
+        return 0
+    }
+    let _ = serverPSK.withUnsafeBytes { (body: UnsafeRawBufferPointer) -> Void in
+        guard let unsafeRawPointer = body.baseAddress else { return }
```
Lukasa, line 151: We can force-unwrap `baseAddress` here: it's required to be non-`nil` if the count of the pointer is greater than zero, so this branch is unreachable.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -112,6 +112,101 @@ private func alpnCallback(ssl: OpaquePointer?,
...
+          else {
+        return 0
+    }
+    
+    // If set, build out a hint to pass into the client callback.
+    guard let clientHint = hint,
+          let derivedHint = String(validatingUTF8: clientHint) else {
+        return 0
+    }
+    // Take the hint and the possible identity and pass them down to the callback to get associated PSK from callback
```
Lukasa, line 186: This comment is out of date I think.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -112,6 +112,101 @@ private func alpnCallback(ssl: OpaquePointer?,
...
+    }
+    let clientPSK = pskIdentityCallback.key // Key from the callback
+    let clientIdentity = pskIdentityCallback.identity ?? ""
+    
+    if clientIdentity.isEmpty {
+        return 0
+    }
+    // Map the output identity from the one passed back from the callback.
+    // This helps populate the server callback for the key exchange.
+    memcpy(unwrappedIdentity, clientIdentity, clientIdentity.count)
```
Lukasa, line 198: This algorithm isn't right, we need to use `withCString` here. This has two issues: first, it uses `clientIdentity.count`, which is a count of `String.Character`, not of bytes; and second, we are required to write a null-terminator into this buffer per the doc-string.

Relatedly, we also aren't comparing against `max_identity_len`, so this can trigger a buffer overrun.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -112,6 +112,101 @@ private func alpnCallback(ssl: OpaquePointer?,
...
+    }
+    
+    // Take the hint and the possible identity and pass them down to the callback to get associated PSK from callback
+    guard let contextPSKIdentityCallback = try? serverCallback(configurationHint, strIdentity) else {
+        return 0
+    }
+    let serverPSK = contextPSKIdentityCallback.key // From the callback
+    
+    // Make sure the key is returned by the callback and it is of proper length, otherwise, fail.
+    if serverPSK.isEmpty || serverPSK.count > max_psk_len {
```
Lukasa, line 147: We can't use this `count`, it is asking the wrong question. We should check against `serverPSK.utf8.count`.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -112,6 +112,101 @@ private func alpnCallback(ssl: OpaquePointer?,
...
+        return 0
+    }
+    let _ = serverPSK.withUnsafeBytes { (body: UnsafeRawBufferPointer) -> Void in
+        guard let unsafeRawPointer = body.baseAddress else { return }
+        memcpy(outputPSK, unsafeRawPointer, body.count)
+    }
+    return UInt32(serverPSK.count)
```
Lukasa, line 154: This also needs to use `serverPSK.utf8.count`.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -112,6 +112,101 @@ private func alpnCallback(ssl: OpaquePointer?,
...
+    let clientIdentity = pskIdentityCallback.identity ?? ""
+    
+    if clientIdentity.isEmpty {
+        return 0
+    }
+    // Map the output identity from the one passed back from the callback.
+    // This helps populate the server callback for the key exchange.
+    memcpy(unwrappedIdentity, clientIdentity, clientIdentity.count)
+    
+    if clientPSK.isEmpty || clientPSK.count > max_psk_len {
```
Lukasa, line 200: This needs to use `clientPSK.utf8.count`.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -112,6 +112,101 @@ private func alpnCallback(ssl: OpaquePointer?,
...
+        return 0
+    }
+    let _ = clientPSK.withUnsafeBytes { (body: UnsafeRawBufferPointer) -> Void in
+        guard let unsafeRawPointer = body.baseAddress else { return }
+        memcpy(outputPSK, unsafeRawPointer, body.count)
+    }
+    return UInt32(clientPSK.count)
```
Lukasa, line 207: This needs to use `clientPSK.utf8.count` as well.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -173,6 +270,22 @@ public final class NIOSSLContext {
...
+            self.pskClientConfigurationCallback = pskClientConfigurationsCallback
+            CNIOBoringSSL_SSL_CTX_set_psk_client_callback(context, { clientPSKCallback(ssl: $0, hint: $1, identity: $2, max_identity_len: $3, psk: $4, max_psk_len: $5 ) } )
+        }
+        
+        if let pskServerConfigurationCallback = configuration.pskServerCallback {
+            self.pskServerConfigurationCallback = pskServerConfigurationCallback
+            CNIOBoringSSL_SSL_CTX_set_psk_server_callback(context, { serverPSKCallback(ssl: $0, identity: $1, psk: $2, max_psk_len: $3 )})
+        }
+        
+        // Set the hint no matter if it is client of server side.
```
Lukasa, line 285:
```suggestion
        // Set the hint no matter if it is client or server side.
```

Lukasa (changes requested):
Thanks @agnosticdev! Sorry for the delay in reviewing, I was on vacation.

Sources/NIOSSL/SSLContext.swift
agnosticdev: In this context `serverPSK` is `NIOSSLSecureBytes` and not a string, so count would work here, no?

Sources/NIOSSL/SSLContext.swift
agnosticdev: In this context `clientPSK` is `NIOSSLSecureBytes` and not a string, so count would work here, no?

Sources/NIOSSL/SSLContext.swift
agnosticdev: In this context `clientPSK` is `NIOSSLSecureBytes` and not a string, so count would work here, no?

Sources/NIOSSL/SSLContext.swift
agnosticdev: In this context `serverPSK` is `NIOSSLSecureBytes` and not a string, so count would work here, no?

Sources/NIOSSL/SSLContext.swift
agnosticdev: Addressed in 0101dcb

Sources/NIOSSL/SSLCallbacks.swift
agnosticdev: Addressed in 0101dcb

Sources/NIOSSL/SSLContext.swift
agnosticdev: Addressed in 0101dcb

Sources/NIOSSL/SSLContext.swift
agnosticdev: Addressed in 0101dcb, let me know if this is what you had in mind.

agnosticdev:
@Lukasa feedback addressed in 0101dcb and all other comments inline.

agnosticdev:
Merging recent updates from `main` into this branch.

Sources/NIOSSL/SSLContext.swift
Lukasa: Oh yes, good catch.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -112,6 +112,106 @@ private func alpnCallback(ssl: OpaquePointer?,
...
+        return 0
+    }
+    // Take the hint and pass it down to the callback to get associated PSK from callback
+    guard let pskIdentityCallback = try? clientCallback(derivedHint) else {
+        return 0
+    }
+    let clientPSK = pskIdentityCallback.key // Key from the callback
+    let clientIdentity = pskIdentityCallback.identity
+    
+    if clientIdentity.isEmpty {
```
Lukasa, line 192: This check is redundant with the one below it.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -112,6 +112,106 @@ private func alpnCallback(ssl: OpaquePointer?,
...
+        return 0
+    }
+    // Use max_identity_len so it does not trigger an overrun.
+    if clientIdentity.isEmpty || clientIdentity.utf8.count > max_identity_len {
```
Lukasa, line 196: Can we use `isEmpty` on `.utf8` as well?

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -446,7 +468,38 @@ extension TLSConfiguration {
...
+                                shutdownTimeout: .seconds(5),
+                                keyLogCallback: nil,
+                                renegotiationSupport: .none,
+                                additionalTrustRoots: [],
+                                sendCANameList: false,
+                                pskClientCallback: nil,
+                                pskServerCallback: nil,
+                                pskHint: nil)
     }
 }
```
Lukasa, line 504: Can you extend `bestEffortEquals` and `bestEffortHash` here, and add appropriate testing? See also #386 (my fault for not catching that in review last time).

Lukasa (changes requested):

Sources/NIOSSL/TLSConfiguration.swift
agnosticdev: Addressed in [f961f4a](https://github.com/apple/swift-nio-ssl/pull/369/commits/f961f4a6d34a6f5f40bc79b33e38e90ae76a12e3)

Sources/NIOSSL/SSLContext.swift
agnosticdev: Addressed in [f961f4a](https://github.com/apple/swift-nio-ssl/pull/369/commits/f961f4a6d34a6f5f40bc79b33e38e90ae76a12e3)

agnosticdev:
@Lukasa updated based on latest feedback.  Thanks for pointing out the `TLSConfiguration` equatable scenario.

Lukasa (approved):
Awesome, great work @agnosticdev! I appreciate that this was a real marathon patch, thanks so much for bearing with me through it all.

agnosticdev:
Thank you @Lukasa for all of the help here, it absolutely was a team effort!

PR #375, opened 2022-06-27
Include file path in `IOError`s thrown by `fopen`

dnadoba:
### Motivation
If a certificate, private key or bundle can not be loaded from the file system, we get an IOError with a description that looks like this:
```
fopen(file:mode:): No such file or directory (errno: 2)
```
It would be nice if the error description would also include the file path of the file that could not be found.

### Modifications
- change `fopen` file argument from `UnsafePointer<CChar>` to `String`. All existing call sites already have a `String`. If we want to include the file path in the error description we also want it as a String. We could convert `UnsafePointer<CChar>` to a `String` through `String(cString:)` but to reduce allocations to a minimum we can also just pass in a `String` directly.
- same as above for `mode`
- include the file path in the `reason` string of all `IOError`s thrown by `Posix.fopen`

### Result
Better error description:
```
fopen(file: "/nonexistent/path", mode: "rb"): No such file or directory (errno: 2)
```

### DIFF
--- a/Sources/NIOSSL/PosixPort.swift
+++ b/Sources/NIOSSL/PosixPort.swift
@@ -101,8 +101,10 @@ internal func wrapErrorIsNullReturnCall<T>(errorReason: @autoclosure () -> Strin
 internal enum Posix {
     @inline(never)
     internal static func fopen(file: String, mode: String) throws -> FILEPointer {
-        return try wrapErrorIsNullReturnCall(errorReason: "fopen(file: \"\(file)\", mode: \"\(mode)\")") {
-            sysFopen(file, mode)
+        try file.withCString { fileCString in
+            try wrapErrorIsNullReturnCall(errorReason: "fopen(file: \"\(file)\", mode: \"\(mode)\")") {
+                sysFopen(fileCString, mode)
+            }
         }
     }
 


Sources/NIOSSL/PosixPort.swift
```diff
@@ -100,8 +100,8 @@ internal func wrapErrorIsNullReturnCall<T>(where function: String = #function, _
 // MARK:- Our functions
 internal enum Posix {
     @inline(never)
-    internal static func fopen(file: UnsafePointer<CChar>, mode: UnsafePointer<CChar>) throws -> FILEPointer {
-        return try wrapErrorIsNullReturnCall {
+    internal static func fopen(file: String, mode: String) throws -> FILEPointer {
```
Lukasa, line 103: Can we add an indirection here? The `String` to `char *` pointer transformation potentially dirties up `errno`, so we don't want it to implicitly happen within the body of `wrapErrorIsNullReturnCall`.

dnadoba: Done in https://github.com/apple/swift-nio-ssl/pull/375/commits/199bdac198e98320dacf1ad8b9c8d2d0f0c0b53f

Lukasa (approved):

PR #376, opened 2022-06-29
Fix non-modular header compile error

makleso6:
I am using vapor and there is error when compiling swift-nio-ssl

`Include of non-modular header inside framework module 'CNIOBoringSSL.CNIOBoringSSL_bn': '/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/clang/include/inttypes.h'`

### DIFF
--- a/Sources/CNIOBoringSSL/crypto/x509/t_x509.c
+++ b/Sources/CNIOBoringSSL/crypto/x509/t_x509.c
@@ -67,7 +67,7 @@
 #include <CNIOBoringSSL_x509v3.h>
 
 #include "internal.h"
-
+#include <inttypes.h>  // for PRIu64 and friends
 
 int X509_print_ex_fp(FILE *fp, X509 *x, unsigned long nmflag,
                      unsigned long cflag)
--- a/Sources/CNIOBoringSSL/include/CNIOBoringSSL_bn.h
+++ b/Sources/CNIOBoringSSL/include/CNIOBoringSSL_bn.h
@@ -126,7 +126,6 @@
 #include "CNIOBoringSSL_base.h"
 #include "CNIOBoringSSL_thread.h"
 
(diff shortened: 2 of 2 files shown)


makleso6:
same issue

https://github.com/AFNetworking/AFNetworking/issues/2205

glbrntt (commented):
Thanks for the PR @makleso6. Unfortunately we can't accept it in its current form: all of the code in the `CNIOBoringSSL` module is vendored in from BoringSSL using https://github.com/apple/swift-nio-ssl/blob/main/scripts/vendor-boringssl.sh.

This script can, and does, apply patches to the vendored code though. I think we could either adjust the existing inttypes patch or add another one: https://github.com/apple/swift-nio-ssl/blob/181802810e2af1f5fc5d26c066ac45030c07418e/scripts/vendor-boringssl.sh#L322

This is quite a gnarly process to go through so we can pick this up from here.

makleso6:
Thanks for reply! I'll try to add patch in my project.

glbrntt:
Ah, sorry I meant that the patch needs to be updated and applied by running the script to update the version of BoringSSL used by NIOSSL, we'll then tag a new release of NIOSSL. You don't need to apply the patch to your project.

glbrntt:
@makleso6 how are you building this? I don't get an error when compiling so I want to make sure we actually fix this.

makleso6:
@glbrntt https://github.com/tuist/tuist/issues/4582

makleso6:
@glbrntt You can reproduce by `swift package generate-xcodeproj` and build project

glbrntt:
Thanks @makleso6 -- this should be fixed by https://github.com/apple/swift-nio-ssl/pull/378 and we'll get a release tagged soon after.

makleso6:
Thanks @ glbrntt. Great!

PR #378, opened 2022-07-04
Update BoringSSL to c239ffd0552179f358de31517391679e9b62ccd3

glbrntt:
Various fixes to the vending script:

- Fix various 'gsed' occurrences with '$sed'
- Remove the delocate patch, it is no longer required
- Add another patch to include inttypes in
  CNIOBoringSSL/crypto/x509/t_x509.c where PRIu64 is used and remove it
  from CNIOBoringSSL_bn.h. this works around a build failure with
  xcodeproj.
- Use the host platform build directory when mangling C++ symbols
- Fix Python bytes/string writing in build-asm.py


Lukasa (changes requested):
Can you also bring over the missing changes to `build-asm`/`vendor-boringssl` from https://github.com/apple/swift-crypto/pull/121?

glbrntt:
We already have those changes:
- The `build-asm.py` change was done in https://github.com/apple/swift-nio-ssl/commit/36880de81050d0416bf523e505a0b1da429a1663.
- The `vendor-boringssl` changes were done in https://github.com/apple/swift-nio-ssl/commit/83c20018523c1a1418773f73157628228527fef2 and https://github.com/apple/swift-nio-ssl/commit/e71a0f38df6c7705925e68d352f543991480069f.

Lukasa:
Oh good note!

Lukasa (approved):

PR #379, opened 2022-07-12
Unbuffer reads after handshake

FranzBusch:
# Motivation
We missed unbuffering reads after the handshake completed. This could result in reads that were buffered while the handshake was in progress to not unbuffer.

# Modification
This PR unbuffers any reads after the handshake is completed.

# Result
We are now unbuffering reads correctly.

### DIFF
--- a/Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
+++ b/Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
@@ -1599,6 +1599,76 @@ class NIOSSLIntegrationTest: XCTestCase {
         XCTAssertThrowsError(try handshakeResultPromise.futureResult.wait())
     }
 
+    func testReadsAreUnbufferedAfterHandshake() throws {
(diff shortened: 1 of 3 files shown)


Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
```diff
@@ -1599,6 +1599,75 @@ class NIOSSLIntegrationTest: XCTestCase {
         XCTAssertThrowsError(try handshakeResultPromise.futureResult.wait())
     }
 
+    func testReadsAreUnbufferedAfterHandshake() throws {
```
weissi, line 1602: please comment that this is a regression test with a link to the bug.

weissi (approved):
LGTM, @FranzBusch did you check that this test _fails_ before the fix and passes after?

FranzBusch:
@weissi Yes this test fails without the fix

PR #383, opened 2022-07-25
Update the expired leaf certificate

dnadoba:
### Motivation
The test suite in the package was failing because the leaf cert apple.com expired on Jul-22.

### Modification
Fetched the new leaf cert by executing `openssl s_client -connect apple.com:443 -showcerts` and copying the resulting cert chain into `SecurityFrameworkVerificationTests.swift`.

### Result
Tests are now passing, as verified by running the test suite locally.


Lukasa (approved):
Thanks!

PR #384, opened 2022-07-26
Tolerate massive writes

Lukasa:
Motivation:

In rare cases, users may attempt to write more data than can be
expressed in 2**31 bytes. This should not crash our program.

Modifications:

- Break up extremely long writes into smaller chunks
- Make sure to clamp on the read path too

Result:

Users who send hilariously long writes can do so.

### DIFF
--- a/Sources/NIOSSL/NIOSSLHandler.swift
+++ b/Sources/NIOSSL/NIOSSLHandler.swift
@@ -693,6 +693,17 @@ extension NIOSSLHandler {
     private typealias BufferedWrite = (data: ByteBuffer, promise: EventLoopPromise<Void>?)
 
     private func bufferWrite(data: ByteBuffer, promise: EventLoopPromise<Void>?) {
+        var data = data
+
+        // Here we guard against the possibility that any of these writes are larger than CInt.max.
+        // This is very unusual but it can happen. To work around it, we just pretend that there were
+        // multiple writes.
+        let maxWriteSize = Int(CInt.max)
+        while let slice = data.readSlice(length: maxWriteSize) {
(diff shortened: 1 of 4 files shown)


FranzBusch (approved):
Nice fix!

glbrntt (approved):

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -693,6 +693,17 @@ extension NIOSSLHandler {
...
+        // Here we guard against the possibility that any of these writes are larger than CInt.max.
+        // This is very unusual but it can happen. To work around it, we just pretend that there were
+        // multiple writes.
+        let maxWriteSize = Int(CInt.max)
+        while let slice = data.readSlice(length: maxWriteSize) {
+            bufferedWrites.append((data: slice, promise: promise))
+        }
+
+        assert(data.readableBytes <= maxWriteSize)
         bufferedWrites.append((data: data, promise: promise))
```
dnadoba, line 707: This will append an empty slice if `data.readableBytes % maxWriteSize == 0`. Should we guard against empty buffer appends?

Lukasa: Oh shoot I think we're mishandling promises here too. Another patch inbound.

Lukasa: To be clear, empty writes are fine here: we just need to handle promises properly.

FranzBusch: Oh you are right! Good catch!

PR #385, opened 2022-07-26
Clean up write logic

Lukasa:
Motivation:

Unnecessary empty writes should be avoided, and we should avoid
completing promises too often.

Modifications:

- Avoid sending a split promise on every write when writes are too big.
- Avoid sending a zero-length write when writes are too big.

Result:

Slightly better write behaviour.

Note that this is fundamentally un-testable: there's no place to put a
test hook here, and we coalesce writes on the far side, so we can't do
much here.

### DIFF
--- a/Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
+++ b/Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
@@ -429,9 +429,6 @@ class NIOSSLIntegrationTest: XCTestCase {
     static var cert: NIOSSLCertificate!
     static var key: NIOSSLPrivateKey!
     static var encryptedKeyPath: String!
-
-    static let giantWriteSize = Int(CInt.max) + 1
-    static let giantWrite = ByteBuffer(repeating: 0, count: giantWriteSize)
     
     override class func setUp() {
         super.setUp()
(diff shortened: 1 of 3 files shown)


Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
```diff
@@ -429,6 +429,9 @@ class NIOSSLIntegrationTest: XCTestCase {
     static var cert: NIOSSLCertificate!
     static var key: NIOSSLPrivateKey!
     static var encryptedKeyPath: String!
+
+    static let giantWriteSize = Int(CInt.max) + 1
+    static let giantWrite = ByteBuffer(repeating: 0, count: giantWriteSize)
```
dnadoba, line 434: This will continue to eat up ~2GB of memory, even after `testGiantWrite()` has finished. Any particular reason you have made it a `static let`?

dnadoba: And I also just noticed that it will crash on system with less than 2GB of memory, regardless of where we allocate it. Should we skip the test on such systems? I think we usually don't run tests on such constrained systems but just a thought I wanted to write down. I'm also not aware of an easy way to find out about the max memory available. I guess we could first try to manually `malloc` a 2GB chunk and see if it returns `nil` or not and `free` it again before we allocated the `ByteBuffer`.

Lukasa: This shouldn't crash on systems with less than 2GB of memory, it should just be slow. There is no requirement that all your virtual memory fits into RAM.

The move to `static let` was motivated by my attempt to write additional tests, but when I realised I couldn't I didn't move it back. Will move it now.

Lukasa: > This shouldn't crash on systems with less than 2GB of memory, it should just be slow. There is no requirement that all your virtual memory fits into RAM.


Concretely, this assumes that you have swap enabled. If you don't then yes, this absolutely _will_ crash the system, but I think we should file this under "not a problem until someone tells us it is".

dnadoba (approved):

PR #386, opened 2022-07-27
Add sendCANameList to best effort functions.

Lukasa:
Motivation:

TLSConfiguration cannot be made equatable due to its use of closures, so
we have our own best-effort functions. Unfortunately, I forgot to make
sure everything was added to them, so I'm doing so now.

Modifications:

- Add missing fields to bestEffort functions.

Result:

Better bestEffort functions.


glbrntt (approved):
LGTM. 

Is there anything we can do to avoid missing new properties in the future? 

The only thing that came to mind was something like:

```swift
XCTAssertEqual(MemoryLayout<TLSConfiguration>.size, ..., "TLSConfiguration has changed size; do bestEffortEquals and bestEffortHash need updating?")
```

Lukasa:
Yeah that's a good idea, let's add that.

glbrntt (approved):

PR #387, opened 2022-07-27
Invoke `additionalPeerCertificateVerificationCallback` with peer certificate

dnadoba:
### Motivation
We want that users can do additional verification of the peer certificate without completely disabling the default certificate chain verification. `additionalPeerCertificateVerificationCallback` is called after the normal certificate chain verification was successful but still gives the user the possibility to accept or reject a connection asynchronously based on the peer certificate.

### Modifications
- rename `AdditionalCertificateChainVerificationCallback` to `_NIOAdditionalPeerCertificateVerificationCallback` as we do not pass in the full certificate chain but just the peers leaf certificate. 
- move the callback from `NIOSSLContext` to `NIOSSLHandler` so users can capture variables in a closure per channel
- make the API public but prefixed with an underscore to indicate that this is not a stable feature. We would like to eventually provide the full certificate chain instead of just the the leaf certificate once we can request it from BoringSSL.

### Result
Users can now easily do custom peer certificate validation while still getting the system/BoringSSL provided full certificate chain validation.

### DIFF
--- a/Sources/NIOSSL/NIOSSLHandler.swift
+++ b/Sources/NIOSSL/NIOSSLHandler.swift
@@ -57,6 +57,12 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
     }
     
     internal init(connection: SSLConnection, shutdownTimeout: TimeAmount, additionalPeerCertificateVerificationCallback: _NIOAdditionalPeerCertificateVerificationCallback?) {
+        let configuration = connection.parentContext.configuration
+        precondition(
+            additionalPeerCertificateVerificationCallback == nil ||
+            configuration.certificateVerification != .none,
+            "TLSConfiguration.certificateVerification must be either set to .noHostnameVerification or .fullVerification if additionalPeerCertificateVerificationCallback is specified"
+        )
         self.connection = connection
         self.bufferedWrites = MarkedCircularBuffer(initialCapacity: 96)  // 96 brings the total size of the buffer to just shy of one page
         self.shutdownTimeout = shutdownTimeout
@@ -260,7 +266,14 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
             
             if let additionalPeerCertificateVerificationCallback = self.additionalPeerCertificateVerificationCallback {
                 state = .additionalVerification
-                additionalPeerCertificateVerificationCallback(connection.getPeerCertificate(), context.channel)
+                guard let peerCertificate = connection.getPeerCertificate() else {
+                    preconditionFailure("""
+                        Couldn't get peer certificate after chain verification was successful.
+                        This should be impossible as we have a precondition during creation of this handler that requires certificate verification.
+                        Please fill an issue.
+                    """)
+                }
+                additionalPeerCertificateVerificationCallback(peerCertificate, context.channel)
                     .hop(to: context.eventLoop)
                     .whenComplete { result in
                         self.completedAdditionalPeerCertificateVerification(result: result)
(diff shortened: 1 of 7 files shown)


Sources/NIOSSL/SSLCallbacks.swift
```diff
@@ -89,6 +89,17 @@ public typealias NIOSSLVerificationCallback = (NIOSSLVerificationResult, NIOSSLC
...
+/// A custom verification callback that allows additional peer certificate verification logic after the logic of BoringSSL has completed successfully.
+///
+/// It is invoked with two arguments:
+/// 1. The leaf certificate from the peer certificate chain if available
+/// 2. The channel to which the certificate belongs
+///
+/// The handshake will only succeed if the returned promise completes successfully.
+///
+/// - warning: This API is not guaranteed to be stable and is likely to be changed without further notice, hence the underscore prefix.
+public typealias _NIOAdditionalPeerCertificateVerificationCallback = (NIOSSLCertificate?, Channel) -> EventLoopFuture<Void>
```
Lukasa, line 101: I don't think this should ever be called with `nil` as the first argument. The only circumstance in which this may be `nil` is if the TLS configuration was not set to validate the peer certificate, and IMO we should error when we're constructed in that configuration.

Sources/NIOSSL/UniversalBootstrapSupport.swift
```diff
@@ -36,17 +36,30 @@ public struct NIOSSLClientTLSProvider<Bootstrap: NIOClientTCPBootstrapProtocol>:
     let context: NIOSSLContext
     let serverHostname: String?
     let customVerificationCallback: NIOSSLCustomVerificationCallback?
+    let additionalPeerCertificateVerificationCallback: _NIOAdditionalPeerCertificateVerificationCallback?
```
Lukasa, line 39: Can we revert this change? I don't think we need it, so I'm inclined to discard it for now.

Lukasa (changes requested):
Generally looks great, a few quick notes here.

Sources/NIOSSL/UniversalBootstrapSupport.swift
dnadoba: Everything is internal only and we need it for tests or we need to rewrite them quite a bit I guess. Here is where we currently need it:
https://github.com/apple/swift-nio-ssl/blob/22bc6f6e4b5ee7487e5fdf0f23cecd16f5bae7ea/Tests/NIOSSLTests/NIOSSLIntegrationTest.swift#L293-L307

Lukasa: I hadn't realised that we were using the TLS provider API for all our testing. Ok, we can keep this change.

Sources/NIOSSL/SSLCallbacks.swift
Lukasa: I'm inclined to trap during handler creation, actually: this is purely a configuration mistake, and not recoverable.

dnadoba: Done in https://github.com/apple/swift-nio-ssl/pull/387/commits/bbdaafce153b497097e15c57e6ab2cce04fbf54a

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -256,12 +264,19 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
...
             
-            if let additionalCertificateChainVerification = connection.parentContext.additionalCertificateChainVerification {
+            if let additionalPeerCertificateVerificationCallback = self.additionalPeerCertificateVerificationCallback {
                 state = .additionalVerification
-                additionalCertificateChainVerification(context.channel)
+                guard let peerCertificate = connection.getPeerCertificate() else {
+                    preconditionFailure("""
+                        Couldn't get peer certificate after chain verification was successful.
+                        This should be impossible as we have a precondition during creation of this handler that requires certificate verification.
+                        Please fill an issue.
```
Lukasa, line 273:
```suggestion
                        Please file an issue.
```

Lukasa (commented):
One quick change please

Lukasa (approved):
Nice one, thanks.

PR #389, opened 2022-08-01
Make `NIOSSLCertificate._subjectAlternativeNames()` return type non-optional

dnadoba:
### Motivation

`NIOSSLCertificate._subjectAlternativeNames()` can currently return `nil` or an empty `_SubjectAlternativeNames` collection. There is no semantic difference in these two cases.

### Modification
- support a `nil` `nameStack` in `_SubjectAlternativeNames`
- make `NIOSSLCertificate._subjectAlternativeNames()` return type non-optional 

### Result

We now return an empty `_SubjectAlternativeNames` instead of `nil`

### Source Compatibility

This is a non-breaking change even though this is public API because it has an underscore prefix. It is therefore unstable and allowed to change in a minor release.


Lukasa (approved):

Lukasa:
Merging over the API breakage checker.

PR #393, opened 2022-08-10
Store internal error file and line number information in error wrappers

wrinkl3:
# Motivation
When generating errors, we want Error wrapper objects to store the file and the line number that the error originated from. This will greatly simplify debugging of the runtime failures.

# Modification
* Extended `BoringSSLInternalError` to store file path and line number along with the BoringSSL error code.
* Extended `errorMessage` to append `filepath:line` to the message string
* Now calling `CNIOBoringSSL_ERR_get_error_line` instead of `CNIOBoringSSL_ERR_get_error` to extract the necessary info

# Result
Printing a BoringSSL error now shows the file path and the line number


Lukasa:

Lukasa (approved):
Nice one, thanks!

Sources/NIOSSL/SSLErrors.swift
```diff
@@ -18,20 +18,21 @@ import NIOCore
 /// Wraps a single error from BoringSSL.
 public struct BoringSSLInternalError: Equatable, CustomStringConvertible, NIOSendable {
     private enum Backing: Hashable {
-        case boringSSLErrorCode(UInt32)
+        case boringSSLErrorInfo(UInt32, String, UInt)
```
glbrntt, line 21: Should these be ignored from the `Equatable`/`Hashable` conformance @Lukasa?

Lukasa: I'm inclined to say no, though it's an interesting question.

glbrntt: Yup! I lean ever-so-slightly towards not including them as the underlying error is still the same regardless of where it was thrown.

Lukasa (approved):

Sources/NIOSSL/SSLErrors.swift
dnadoba: We could also expose the error code (wrapped in a struct) if we need to compare two errors just base on the error code.

PR #392, opened 2022-08-10
Adopt `Sendable` incrementally

dnadoba:
Incremental `Sendable` adoption.
### Other changes
- use `swift-nio` 2.41.1 as we want to use `NIOPreconcurrencySendable`
- make `NIOSSLCertificate`, `NIOSSLPrivateKey` and `NIOSSLPublicKey` final

PR #395, opened 2022-08-11
Adopt `Sendable` incrementally

dnadoba:
Incremental `Sendable` adoption. Follow up of https://github.com/apple/swift-nio-ssl/pull/392

Properties are not marked with `@preconcurrency` because we otherwise get warnings because of https://github.com/apple/swift/issues/60508
I think that the latest Swift 5.7 toolchain doesn't produces any errors related to `Sendable` requirements in this context and we therefore don't actually need `@preconcurrency` here.

### DIFF
--- a/Sources/NIOSSL/TLSConfiguration.swift
+++ b/Sources/NIOSSL/TLSConfiguration.swift
@@ -288,11 +288,21 @@ public struct TLSConfiguration {
     /// The private key associated with the leaf certificate.
     public var privateKey: NIOSSLPrivateKeySource?
     
+    #if swift(>=5.6)
+    /// PSK Client Callback to get the key based on hint and identity.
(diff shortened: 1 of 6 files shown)


Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -315,7 +315,7 @@ public struct TLSConfiguration {
     /// The amount of time to wait after initiating a shutdown before performing an unclean
     /// shutdown. Defaults to 5 seconds.
     public var shutdownTimeout: TimeAmount
-
+
```
Lukasa, line 318: Can you remove these changes?

Lukasa:
Can you also clean up the test errors on nightly?

Lukasa (approved):

PR #413, opened 2022-12-22
.noHostnameNoPeerVerification

borisreitman:
This option for `TLSConfiguration` is similar to `.noHostname` but will not fail the handshake if the client certificate is not provided. But, if the client certificate _is_ provided, then the `customVerificationCallback` (if provided) will be called.

From an article referenced below, here is a quote:

"Then we tell it to accept requests with no valid certificate. We need this to handle invalid connections as well (for example to display an error message), otherwise, they would just get a cryptic HTTPS error message from the browser (ERR_BAD_SSL_CLIENT_AUTH_CERT to be precise)"

Reference: https://medium.com/@sevcsik/authentication-using-https-client-certificates-3c9d270e8326

### DIFF
--- a/Sources/NIOSSL/TLSConfiguration.swift
+++ b/Sources/NIOSSL/TLSConfiguration.swift
@@ -140,6 +140,9 @@ public enum CertificateVerification: Sendable {
     /// be checked to see if they are valid for the given hostname.
     case noHostnameVerification
 
+    /// Peer certificate is optional
+    case noHostnameNoPeerVerification
+
     /// Certificates will be validated against the trust store and checked
     /// against the hostname of the service we are contacting.
     case fullVerification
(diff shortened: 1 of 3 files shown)


Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -140,6 +140,9 @@ public enum CertificateVerification: Sendable {
     /// be checked to see if they are valid for the given hostname.
     case noHostnameVerification
 
+    /// Peer certificate is optional
+    case noHostnameNoPeerVerification
```
dnadoba, line 144: Adding a new case to a public enum is an API breaking change because a user could switch over it and the compiler would now complain that the new case `.noHostnameNoPeerVerification` is missing.
We will need to think of some other way to make this configurable. Maybe deprecating the current enum and adding a new struct that is backed by an enum.

borisreitman:
@dnadoba Can you check it again, I pushed a new changeset.

Sources/NIOSSL/TLSConfiguration.swift
```diff
@@ -268,6 +268,7 @@ public struct TLSConfiguration {
 
     /// Whether to verify remote certificates.
     public var certificateVerification: CertificateVerification
+    public var certificateRequired: Bool = true
```
Lukasa, line 271: Can you add a doc comment to this explaining what it does and how it interacts with the above?

Lukasa:

PR #424, opened 2023-04-13
Update BoringSSL to abfd5ebc87ddca0fab9fca067c9d7edbc355eae8

Lukasa:
This patch updates BoringSSL, and also makes a number of tweaks required to keep the vendoring script working well.


dnadoba (approved):
LGTM. but I just had a look at the changed .swift files and the scripts.

glbrntt (approved):
Looking forward to `patch-4-even-more-inttypes.patch` 🙃

PR #425, opened 2023-04-13
Drop Swift 5.5

Lukasa:
This patch drops Swift 5.5 support. It removes docker-compose files and deletes any conditional compilation guards that are no longer necessary.


FranzBusch (approved):

yim-lee:
@Lukasa I deleted the 5.5 builder

PR #426, opened 2023-04-24
Skipping testNotValidAfter on 32-bit platform

### LINKED ISSUE #344: SSLCertificateTest.testNotValidAfter fails on Android armv7 and likely other 32-bit platforms
Somebody recently reported problems using this package on Android armv7, termux/termux-packages#8569, so I had him run the tests for this package on there and he reported this test failing. I reproduced and looked into it today: the issue is an overflow when [trying to put too large a UNIX epoch time in seconds into `time_t`](https://github.com/apple/swift-nio-ssl/commit/58b355e5dc68a05c0c69ec14a1f44bdb94791e7c#diff-a9515aaeded6b0c424bee2094c34bca1ed80807fbb9c13264513d51ff4c7850aR540), which is a 32-bit signed integer on Android armv7.

That test has a certificate valid till 2047, which is too long so it fails; you should be able to reproduce on any other 32-bit platform with a similar `time_t`.

bharsaklemukesh975:
Fixes #344

### DIFF
--- a/Tests/NIOSSLTests/SSLCertificateTest.swift
+++ b/Tests/NIOSSLTests/SSLCertificateTest.swift
@@ -480,12 +480,14 @@ class SSLCertificateTest: XCTestCase {
     }
 
     func testNotValidAfter() throws {
+        #if arch(arm64) || arch(x86_64)
         let cert = try NIOSSLCertificate(bytes: .init(samplePemCert.utf8), format: .pem)
         let notValidBeforeSeconds = cert.notValidAfter
 
         let expectedDate = self.dateFromComponents(year: 2047, month: 10, day: 9, hour: 21, minute: 01, second: 02)
         let expectedSeconds = time_t(expectedDate.timeIntervalSince1970)
         XCTAssertEqual(notValidBeforeSeconds, expectedSeconds)
+        #endif
     }
 
     func testNotBeforeAfterGeneratedCert() throws {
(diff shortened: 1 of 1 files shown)


Tests/NIOSSLTests/SSLCertificateTest.swift
```diff
@@ -480,12 +480,14 @@ class SSLCertificateTest: XCTestCase {
     }
 
     func testNotValidAfter() throws {
+        #if arch(arm64) || arch(x86_64)
```
Lukasa, line 483: Thanks for this! It'd probably be better to express this constraint as "the system must have a 64-bit `time_t`", just for future-proofing. That would mean something like:

```swift
guard MemoryLayout<time_t>.size >= 8 else {
    throw XCTSkip()
}
```

Tests/NIOSSLTests/SSLCertificateTest.swift
```diff
@@ -480,6 +480,9 @@ class SSLCertificateTest: XCTestCase {
     }
 
     func testNotValidAfter() throws {
+        guard MemoryLayout<time_t>.size >= 8 else {
+            throw XCTSkip()
+        }
```
dnadoba, line 485: optional nit: this is now the behaviour we want but we can make the spelling a bit more idiomatic by using `XCTSkipUnless` and include a reason:
```suggestion
        try XCTSkipUnless(MemoryLayout<time_t>.size >= 8, "size of time_t must be 64bit or greater")
```

dnadoba (approved):
@bharsaklemukesh975 Thank you for tackling this issue! LGTM.

dnadoba:

PR #428, opened 2023-05-15
`NIOSSLHandler`: behave sensibly on `close(mode: .output)`

felixschlegel:
### Motivation

If NIOSSLHandler receives `close(mode: .output) or close(mode: .input)`, it ignores them, optionally failing their promises. This is weird behaviour. It should do better.

### Modifications

* flush outstanding messages on `NIOSSLHandler.close(mode: .output)`
* close SSL Connection on `NIOSSLHandler.close(mode: .output)`
* support closing the input of `NIOSSLHandler` when used as a server with `ChannelOptions.Types.AllowRemoteHalfClosureOption` enabled
* trigger TCP close when NIOSSLHandler.state == .inputClose
* write integration tests to verify the behaviour

> **Warning**: This PR relies on `EmbeddedChannel` supporting `ChannelOptions.Types.AllowRemoteHalfClosureOption`, which is currently not possible and [proposed as a Patch to `swift-nio`](https://github.com/apple/swift-nio/pull/2429)

> **Note**: I have removed a lot of whitespace, so consider selecting `Hide Whitespace` in the GitHub `Files Changed` tab

### DIFF
--- a/Sources/NIOSSL/NIOSSLHandler.swift
+++ b/Sources/NIOSSL/NIOSSLHandler.swift
@@ -41,10 +41,11 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
         case idle
         case handshaking
         case additionalVerification
-        case active(CloseMode?) // Despite being active, our connection can have a closed output
+        case active
         case unwrapping(Scheduled<Void>)
         case closing(Scheduled<Void>)
         case unwrapped
+        case outputClosed // TODO: make outputClosed an associated value of .active?
         case closed
     }
 
@@ -168,7 +169,7 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
         switch state {
         case .handshaking:
             doHandshakeStep(context: context)
-        case .active:
+        case .active, .outputClosed:
             doDecodeData(context: context)
             doUnbufferWrites(context: context)
         case .closing:
@@ -193,7 +194,7 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
     }
 
     public func write(context: ChannelHandlerContext, data: NIOAny, promise: EventLoopPromise<Void>?) {
-        if case let .active(mode) = self.state, mode == .output {
+        if case .outputClosed = self.state {
             promise?.fail(ChannelError.outputClosed)
             return
         }
@@ -205,7 +206,7 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
         case .idle, .handshaking, .additionalVerification:
             // we should not flush immediately as we have not completed the handshake and instead buffer the flush
             self.bufferFlush()
-        case .active, .unwrapping, .closing, .unwrapped, .closed:
+        case .active, .unwrapping, .closing, .unwrapped, .outputClosed, .closed:
             self.bufferFlush()
             self.doUnbufferWrites(context: context)
         }
@@ -233,8 +234,8 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
             } else if let promise = promise {
                 self.closePromise = promise
             }
-        case .idle, .closed, .unwrapping, .unwrapped:
-            // For idle, closed, unwrapping, and unwrapped connections we immediately pass this on to the next
+        case .idle, .outputClosed, .closed, .unwrapping, .unwrapped:
+            // For idle, outputClosed, closed, unwrapping, and unwrapped connections we immediately pass this on to the next
             // channel handler.
             context.close(mode: .output, promise: promise)
         case .active, .handshaking, .additionalVerification:
@@ -250,7 +251,7 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
             self.flush(context: context)
 
             self.closeOutputPromise?.futureResult.whenSuccess {
-                self.state = .active(.output)
+                self.state = .outputClosed
             }
         }
     }
@@ -282,7 +283,7 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
             // For idle, closed, and unwrapped connections we immediately pass this on to the next
             // channel handler.
             context.close(promise: promise)
-        case .active, .handshaking, .additionalVerification:
+        case .active, .outputClosed, .handshaking, .additionalVerification:
             // We need to begin processing shutdown now. We can't fire the promise for a
             // while though.
             self.state = .closing(self.scheduleTimedOutShutdown(context: context))
@@ -333,7 +334,7 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
                 return
             }
 
-            state = .active(nil)
+            state = .active
             completeHandshake(context: context)
         case .failed(let err):
             writeDataToNetwork(context: context, promise: nil)
@@ -373,7 +374,7 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
         context.eventLoop.preconditionInEventLoop()
 
         switch self.state {
-        case .idle, .handshaking, .active:
+        case .idle, .handshaking, .active, .outputClosed:
             preconditionFailure("invalid state \(self.state)")
         case .additionalVerification:
             switch result {
@@ -382,7 +383,7 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
                 context.fireErrorCaught(error)
                 channelClose(context: context, reason: error)
             case .success:
-                state = .active(nil)
+                state = .active
                 completeHandshake(context: context)
             }
         case .unwrapping, .closing, .unwrapped, .closed:
@@ -454,7 +455,7 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
     private func scheduleTimedOutShutdown(context: ChannelHandlerContext) -> Scheduled<Void> {
         return context.eventLoop.scheduleTask(in: self.shutdownTimeout) {
             switch self.state {
-            case .idle, .handshaking, .additionalVerification, .active:
+            case .idle, .handshaking, .additionalVerification, .active, .outputClosed:
                 preconditionFailure("Cannot schedule timed out shutdown on non-shutting down handler")
 
             case .closed, .unwrapped:
@@ -512,7 +513,7 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
                     // is the result of a wonky I/O pattern and just ignore it.
                     self.plaintextReadBuffer = receiveBuffer
                     break readLoop
-                case .active:
+                case .active, .outputClosed:
                     self.state = .closing(self.scheduleTimedOutShutdown(context: context))
                 case .unwrapping, .closing:
                     break
@@ -739,7 +740,7 @@ extension NIOSSLHandler {
             self.shutdownPromise = promise
             self.channelUnwrap(context: storedContext)
 
-        case .handshaking, .active, .additionalVerification:
+        case .handshaking, .active, .outputClosed, .additionalVerification:
             // Time to try to strip TLS.
             guard let storedContext = self.storedContext else {
                 promise?.fail(NIOTLSUnwrappingError.invalidInternalState)
--- a/Sources/NIOSSL/NIOSSLHandler.swift
+++ b/Sources/NIOSSL/NIOSSLHandler.swift
@@ -200,6 +200,52 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
         self.writeDataToNetwork(context: context, promise: nil)
     }
 
+    // TODO: how do I test this?
+    public func userInboundEventTriggered(context: ChannelHandlerContext, event: Any) {
+        switch event {
+        case ChannelEvent.inputClosed:
+            userInboundInputClosedTriggered(context: context)
+        default:
+            context.fireUserInboundEventTriggered(event)
+            // Trigger closure of TCP input as this event has not occured although we are in state .inputClosed already.
+            if case .inputClosed = state {
+                context.fireUserInboundEventTriggered(ChannelEvent.inputClosed)
+            }
+        }
+    }
+
+    private func userInboundInputClosedTriggered(context: ChannelHandlerContext) {
+        let channelError: NIOSSLError
+        switch self.state {
+        case .closed, .idle:
+            return
+        case .inputClosed:
+            context.fireUserInboundEventTriggered(ChannelEvent.inputClosed)
+            return
+        case .handshaking:
+            // In this case the channel is going through the doHandshake steps and
+            // a channelInactive is fired taking down the connection.
+            // This case propogates a .handshakeFailed instead of an .uncleanShutdown.
+            // We use a synthetic error here as the error stack will be empty, and we should try to
+            // provide some diagnostic help.
+            channelError = NIOSSLError.handshakeFailed(.sslError([.eofDuringHandshake]))
+        case .additionalVerification:
+            // In this case the channel is going through the doHandshake steps and
+            // a channelInactive is fired taking down the connection.
+            // This case propogates a .handshakeFailed instead of an .uncleanShutdown.
+            // We use a synthetic error here as the error stack will be empty, and we should try to
+            // provide some diagnostic help.
+            channelError = NIOSSLError.handshakeFailed(.sslError([.eofDuringAdditionalCertficiateChainValidation]))
+        default:
+            // This is a ragged EOF: we weren't sent a CLOSE_NOTIFY. We want to send a user
+            // event to notify about this before we propagate channelInactive. We also want to fail all
+            // these writes.
+            channelError = NIOSSLError.uncleanShutdown
+        }
+        context.fireErrorCaught(channelError)
+
+    }
+
     public func write(context: ChannelHandlerContext, data: NIOAny, promise: EventLoopPromise<Void>?) {
         bufferWrite(data: unwrapOutboundIn(data), promise: promise)
     }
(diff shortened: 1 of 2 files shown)


Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -205,14 +210,52 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
...
+            // of the TLS shutdown request to this new one.
+            if let promise = promise, let closePromise = self.closePromise {
+                closePromise.futureResult.cascade(to: promise)
+            } else if let promise = promise {
+                self.closePromise = promise
+            }
+        case .idle, .closed, .unwrapping, .unwrapped:
+            // For idle, closed, unwrapping, and unwrapped connections we immediately pass this on to the next
+            // channel handler.
+            context.close(mode: .output, promise: promise)
```
felixschlegel, line 239: Question: what do you think of this behaviour? Passing on `context.close(mode: .output)` can result in errors if the next `ChannelHandler` does not support `CloseMode.output`?

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -102,14 +107,18 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
     }
 
     public func channelActive(context: ChannelHandlerContext) {
+        context.channel.getOption(ChannelOptions.allowRemoteHalfClosure).whenSuccess { halfClosureAllowed in
```
Lukasa, line 110: I don't think caching this value is semantically quite right: users can change this value during the lifetime of the `Channel`. We should look it up when we need it.

Similarly, we should try to use the sync options where we can.

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -190,7 +199,53 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
         self.doFlushReadData(context: context, receiveBuffer: receiveBuffer, readOnEmptyBuffer: true)
         self.writeDataToNetwork(context: context, promise: nil)
     }
-    
+
+    // TODO: how do I test this?
```
Lukasa, line 203: With `EmbeddedChannel`, you can fire the input closed event into the pipeline manually.

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -190,7 +199,53 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
...
+    public func userInboundEventTriggered(context: ChannelHandlerContext, event: Any) {
+        switch event {
+        case ChannelEvent.inputClosed:
+            userInboundInputClosedTriggered(context: context)
+        default:
+            context.fireUserInboundEventTriggered(event)
+            // Trigger closure of TCP input as this event has not occured although we are in state .inputClosed already.
```
Lukasa, line 210: This code seems to be in the wrong place.

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -190,7 +199,53 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
...
+                context.fireUserInboundEventTriggered(ChannelEvent.inputClosed)
+            }
+        }
+    }
+
+    private func userInboundInputClosedTriggered(context: ChannelHandlerContext) {
+        let channelError: NIOSSLError
+        switch self.state {
+        case .closed, .idle:
+            return
```
Lukasa, line 221: In both of these cases we should probably still send the user event onwards.

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -190,7 +199,53 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
...
+        case .inputClosed:
+            context.fireUserInboundEventTriggered(ChannelEvent.inputClosed)
```
Lukasa, line 223: Ideally this isn't right. We should synthesise this event when we _enter_ the `.inputClosed` state, and then suppress it here.

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -200,20 +255,70 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
         case .idle, .handshaking, .additionalVerification:
             // we should not flush immediately as we have not completed the handshake and instead buffer the flush
             self.bufferFlush()
-        case .active, .unwrapping, .closing, .unwrapped, .closed:
+        case .active, .unwrapping, .closingOutput, .closing, .unwrapped, .inputClosed, .outputClosed, .closed: // TODO: .closingOutput + .inputClosed correct here?
```
Lukasa, line 258: All of these are correct.

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -200,20 +255,70 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
...
+
     public func close(context: ChannelHandlerContext, mode: CloseMode, promise: EventLoopPromise<Void>?) {
-        guard mode == .all else {
-            // TODO: Support also other modes ?
+        switch mode {
+        case .output:
+            self.closeOutput(context: context, promise: promise)
+        case .all:
+            self.closeAll(context: context, promise: promise)
+        default:
```
Lukasa, line 270: May as well write `case .input` as it's the only case left and take advantage of the exhaustive switch.

Sources/NIOSSL/NIOSSLHandler.swift
Lukasa: Passing this on is fine.

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -200,20 +255,70 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
...
+        case .idle, .outputClosed, .closed, .unwrapping, .unwrapped:
+            // For idle, outputClosed, closed, unwrapping, and unwrapped connections we immediately pass this on to the next
+            // channel handler.
+            context.close(mode: .output, promise: promise)
+        case .active, .handshaking, .additionalVerification, .inputClosed:
+            // It may occur that we are still in the process of handshaking or additional verification. We'll let that happen.
+            // We flush all outstanding writes once the handshake step is complete and set our state to .outputClosed aftwerwards.
+            // This prevents any further writes to this channel.
+            if let promise = promise, let closeOutputPromise = self.closeOutputPromise {
+                closeOutputPromise.futureResult.cascade(to: promise)
```
Lukasa, line 303: Can we possibly have a `closeOutputPromise` if we're not in state `closingOutput`?

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -200,20 +255,70 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
...
+            } else if let promise = promise {
+                self.closeOutputPromise = promise
+                self.state = .closingOutput(self.scheduleTimedOutShutdown(context: context))
+                self.doShutdownStep(context: context)
```
Lukasa, line 307: It seems wrong to only do the shutdown if there's a promise.

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -200,20 +255,70 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
...
+            }
+
+            self.flush(context: context)
```
Lukasa, line 310: I think we have to flush _before_ we start the shutdown.

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -200,20 +255,70 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
...
+
+            self.closeOutputPromise?.futureResult.whenSuccess {
```
Lukasa, line 312: This is incorrect: if we don't have a promise, this never fires.

This makes me thing that maybe `outputClosing` isn't necessary at all. It's a sort of immediate transition: we send our CLOSE_NOTIFY and the process is conceptually "done". So maybe this is an atomic transition.

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -391,6 +503,8 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
             context.fireUserInboundEventTriggered(TLSUserEvent.shutdownCompleted)
 
             switch targetCompleteState {
+            case .outputClosed:
+                break
```
Lukasa, line 507: I think this is a relevant edge: if it happens, we end up immediately in full closed.

Sources/NIOSSL/NIOSSLHandler.swift
felixschlegel: Ok, I have added two new commits, one removing the `.outputClosing` state and one removing the `closeOutputPromise` — let me know what you think!

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -469,15 +553,24 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
                     // is the result of a wonky I/O pattern and just ignore it.
                     self.plaintextReadBuffer = receiveBuffer
                     break readLoop
-                case .active:
-                    self.state = .closing(self.scheduleTimedOutShutdown(context: context))
-                case .unwrapping, .closing:
+                case .active, .outputClosed:
+                    if self.getAllowRemoteHalfClosureFromChannel(context: context) == false {
```
Lukasa, line 557: Ideally we'd cache this across the call to `doFlushReadData`.

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -491,6 +584,16 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
         }
     }
 
+    /// Checks if the `allowRemoteHalfClosure` channel option is set.
+    private func getAllowRemoteHalfClosureFromChannel(context: ChannelHandlerContext) -> Bool {
+        var halfClosureAllowed = false
+        if let syncOptions = context.channel.syncOptions {
+            let result = try? syncOptions.getOption(ChannelOptions.allowRemoteHalfClosure)
+            result.map { halfClosureAllowed = $0 }
```
Lukasa, line 592: Using `map` for the side-effect is a bit weird here, it's probably better to use `if let`.

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -722,7 +825,33 @@ extension NIOSSLHandler {
 extension NIOSSLHandler {
     private typealias BufferedWrite = (data: ByteBuffer, promise: EventLoopPromise<Void>?)
 
+    /// Completes the given promise when all outstanding writes have been flushed.
+    private func bufferedWritesCascadeFlushPromise(
+        to promise: EventLoopPromise<Void>?,
```
Lukasa, line 830: I don't think this use of the promise is correct. It should be attached to the `close(mode: .output)` operation. We should _fail_ it if we don't actually issue that call, but we shouldn't succeed it ourselves.

Sources/NIOSSL/NIOSSLHandler.swift
felixschlegel: Good point!

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -190,7 +192,48 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
...
+            // We use a synthetic error here as the error stack will be empty, and we should try to
+            // provide some diagnostic help.
+            channelError = NIOSSLError.handshakeFailed(.sslError([.eofDuringAdditionalCertficiateChainValidation]))
+        default:
+            // This is a ragged EOF: we weren't sent a CLOSE_NOTIFY. We want to send a user
+            // event to notify about this before we propagate channelInactive. We also want to fail all
+            // these writes.
+            channelError = NIOSSLError.uncleanShutdown
+        }
+        context.fireErrorCaught(channelError)
```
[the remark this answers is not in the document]
Lukasa: Can we reverse the order of these two events? The error is more prominent, I think, especially when we hit a ragged EOF.

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -200,19 +244,56 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
...
+            // We flush all outstanding writes once the handshake step is complete and set our state to .outputClosed aftwerwards.
+            // This prevents any further writes to this channel.
+            self.state = .outputClosed
+            self.closeOutputPromise = promise
+
+            let flushCompletePromise: EventLoopPromise<Void> = context.eventLoop.makePromise()
+            flushCompletePromise.futureResult.whenComplete { _ in
+                self.doShutdownStep(context: context)
+            }
+            bufferedWritesCascadeFlushPromise(to: flushCompletePromise, context: context)
```
Lukasa, line 290: So this promise dance is probably unnecessary. We can send `closeOutput` once we've called `flush`, even if the write promises haven't completed. Handlers are required to queue `close(mode: .output)` behind writes.

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -342,7 +423,7 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
                 state = .active
                 completeHandshake(context: context)
             }
-        case .unwrapping, .closing, .unwrapped, .closed:
+        case .unwrapping, .closing, .unwrapped, .closed, .inputClosed, .outputClosed:
```
Lukasa, line 426: I think this is wrong, `.inputClosed` and `.outputClosed` should match `active` here I think.

felixschlegel:
Hey @Lukasa ,

I have implemented your requested changes and added the buffering of `closeOutput` when we are still in `state = .handshaking / .additionalVerification`.

I still have a couple of open questions for this PR:

* should we also support `scheduledShutdown` for `close(mode: .output)` (see the normal close mode for reference)
* what if the downstream `ChannelHandler` does not support `close(mode: .output)`? At the moment we just invoke `context.close(mode: .output)` hoping for the next `ChannelHandler` not to fail

Best regards,
-Felix

Lukasa:

Lukasa:
> should we also support scheduledShutdown for close(mode: .output) (see the normal close mode for reference)

No, scheduled shutdown is waiting for us to get a `CLOSE_NOTIFY` from the remote peer. In this mode, we don't expect one.

> what if the downstream ChannelHandler does not support close(mode: .output)?

Then the partial close fails and the user should upgrade to full close.

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -732,27 +865,29 @@ extension NIOSSLHandler {
...
-        bufferedWrites.mark()
+        bufferedActions.mark()
     }
 
-    private func discardBufferedWrites(reason: Error) {
-        while self.bufferedWrites.count > 0 {
-            let bufferedWrite = self.bufferedWrites.removeFirst()
-            bufferedWrite.promise?.fail(reason)
+    private func discardBufferedActions(reason: Error) {
+        while self.bufferedActions.count > 0 {
```
Lukasa, line 880: This can be more concisely expressed as `while let bufferedAction = self.bufferedActions.popFirst()`.

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -778,13 +920,28 @@ extension NIOSSLHandler {
...
+            // We detected a .closeOutput action in our action buffer. This means we
+            // close the output after we have written all pending writes.
+            if invokeCloseOutput {
+                self.state = .outputClosed
+                self.doShutdownStep(context: context)
+            }
         } catch {
             // We encountered an error, it's cleanup time. Close ourselves down.
-            channelClose(context: context, reason: error)
+            if case .outputClosed = self.state {} else {
```
Lukasa, line 932: This is a really weird construction, prefer using a `switch`.

felixschlegel:

yim-lee:

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -391,6 +484,10 @@ public class NIOSSLHandler : ChannelInboundHandler, ChannelOutboundHandler, Remo
             context.fireUserInboundEventTriggered(TLSUserEvent.shutdownCompleted)
 
             switch targetCompleteState {
+            case .outputClosed:
+                // No full channel close here. We would expect users to invoke a full close regardless of
+                // previously completed half closures.
```
Lukasa, line 489: This comment is out of date now, right?

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -778,13 +933,33 @@ extension NIOSSLHandler {
...
+                self.state = .outputClosed
+                self.doShutdownStep(context: context)
+                self.discardBufferedActions(reason: ChannelError.outputClosed)
+            }
         } catch {
             // We encountered an error, it's cleanup time. Close ourselves down.
-            channelClose(context: context, reason: error)
+            switch self.state {
+            case .outputClosed:
+                break
```
Lukasa, line 948: Why do we suppress error handling in this state?

Sources/NIOSSL/NIOSSLHandler.swift
felixschlegel: Thanks for pointing that out! I have updated the comment so that it reflects the current state of our half-closure behaviour 😄

Sources/NIOSSL/NIOSSLHandler.swift
felixschlegel: Fixed 🔨

Lukasa (approved):

PR #434, opened 2023-06-08
Use EVP_PKEY_assign_RSA instead of EVP_PKEY_assign

davidben:
EVP_PKEY_assign_RSA, at least in C, is more type-safe. It also is compatible with an upcoming BoringSSL change to make the RSA struct opaque. For some Swift reasons I don't fully understand (but relating to the OpaquePointer mess), when RSA becomes opaque, EVP_PKEY_assign no longer works.

I assume it could be made to work with some appropriate cast, but since EVP_PKEY_assign_RSA already exists (and will, in the future, be more binary-size-friendly), just use that.


Lukasa (approved):
Thanks @davidben! :sparkles:

Lukasa:

PR #436, opened 2023-06-21
Add if #available for xrOS

brennobemoura:
Xcode 14 Beta 2 is warning about using SecTrustEvaluateAsync method

> **Warning**: SecTrustEvaluateAsync' was deprecated in xrOS 1.0: renamed to 'SecTrustEvaluateAsyncWithError(_:_:_:)'

### DIFF
--- a/Sources/NIOSSL/SecurityFrameworkCertificateVerification.swift
+++ b/Sources/NIOSSL/SecurityFrameworkCertificateVerification.swift
@@ -66,7 +66,7 @@ extension SSLConnection {
             callbackQueue.async {
                 let result: OSStatus
 
-                if #available(iOS 13, macOS 10.15, tvOS 13, watchOS 6, *) {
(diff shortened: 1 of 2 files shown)


FranzBusch:

FranzBusch:
I confirmed this warning locally. Just running the CI to see if we can land this without breaking older Swift versions

FranzBusch:

brennobemoura:
@FranzBusch I've updated this. I understand now that this change as effect only in the compilation warning, not the right behavior of the code when running in xrOS. So, my question is, do we need this PR?

FranzBusch:
This warning is interesting and looks like a compiler bug to me. We are actually taking the correct branch at runtime but we are getting the warning in the wrong branch. I filed an issue for the compiler: https://github.com/apple/swift/issues/66854

PR #437, opened 2023-06-22
UnwrappingTests use own class member cert values

rnro:
Motivation:

Running the tests with `--parallel --num-workers=1` revealed a failure on tearDown

Modifications:

UnwrappingTests use own class member cert values rather than those from NIOSSLIntegrationTest.

Result:

Tests pass


glbrntt (approved):

glbrntt:

PR #441, opened 2023-08-14
Fix expired test certs

### LINKED ISSUE #440: Certificate tests started failing today
I haven't bothered dumping them to check the dates, but the exact same `TLSConfigurationTest.testFullVerificationWithCANamesFrom{Certificate,File}` tests that passed yesterday [started failing on my daily Android CI today](https://github.com/finagolfin/swift-android-sdk/actions/runs/5830798355/job/15815405738#step:3:9344). I'm guessing the certs added for these tests in #307 expired today.

czechboy0:
## Summary

Fixes #440 by generating new certs for tests, and added instructions on how to do it in a year again.

## Test Plan

This PR is fixing tests.


dnadoba (approved):
Thanks! Hopefully by next year we can generate the certs with `swift-certificates` ;)

dnadoba:

finagolfin:
They've [expired again](https://github.com/finagolfin/swift-android-sdk/actions/runs/10366884580/job/28697157858#step:35:10390), @dnadoba, do your magic. 😉

dnadoba:
This would add a dependency for `swift-asn1`, `swift-certificates` and `swift-crypto`. Long term we end up with these dependencies anyway and this should be fine.
Even if not, IIRC we have done an investigation a while ago and test only dependencies are actually fine and don't restrict dependency resolution of upstream packages. @czechboy0 do you remember the details?

czechboy0:
I think that's right, I'll let @simonjbeaumont confirm though, as he's taken a closer look at this.

simonjbeaumont:
Yeah. Target-based dependency resolution should allow for us to generate the test certs. 

@Lukasa are you happy for us to have test-only package dependencies here?

If so, I'm happy to pick this up when I'm back from vacation.

Lukasa:
Yes, I'm fine with that.

PR #443, opened 2023-09-12
Add support for shrinking massive ByteBufferBIOs

Lukasa:
Motivation:

ByteBufferBIOs will grow their outbound buffer until flush is called. They will retain their capacity at however large it got to avoid reallocating, as they assume that your write pattern will be fairly consistent.

In some cases this is a bad strategy, because it overcommits data. To allow users to reclaim this data, we can add a configuration option that will shrink the playload on the next write.

Modifications:

- Added support for ByteBufferBIO to shrink the payload.
- Plumbed up support for handler configuration.
- Tests that shrinkage actually works.

Result:

Users can reclaim memory.

### DIFF
--- a/Sources/NIOSSL/NIOSSLHandler+Configuration.swift
+++ b/Sources/NIOSSL/NIOSSLHandler+Configuration.swift
@@ -1,0 +1,38 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2023 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+extension NIOSSLHandler {
+    /// Configuration for a specific instance of ``NIOSSLHandler``, either client or server.
+    ///
+    /// This type is distinct from ``TLSConfiguration`` because it does not contain settings that
+    /// apply to TLS itself. Instead, this configuration manages how the ``NIOSSLHandler`` itself
+    /// operates.
+    public struct Configuration: Hashable, Sendable {
+        /// The maximum number of bytes we'll preserve in the outbound buffer that ``NIOSSLHandler``
+        /// holds.
+        ///
+        /// This buffer is not typically deallocated, as it is re-used throughout the lifetime of
+        /// the program. In cases where there are extremely large peak writes that are outliers in
+        /// the code, the buffer may remain excessively large.
+        ///
+        /// Set this value to a lower value to avoid preserving too much memory. This will cause
+        /// ``NIOSSLHandler`` to reallocate memory more often, which can inhibit performance, so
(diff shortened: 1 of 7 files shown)


Sources/NIOSSL/NIOSSLHandler+Configuration.swift
```diff
@@ -0,0 +1,38 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+extension NIOSSLHandler {
+    /// Configuration for a specific instance of ``NIOSSLHandler``, either client or server.
+    ///
+    /// This type is distinct from ``TLSConfiguration`` because it does not contain settings that
+    /// apply to TLS itself. Instead, this configuration manages how the ``NIOSSLHandler`` itself
+    /// operates.
+    public struct Configuration: Hashable, Sendable {
```
glbrntt, line 20: I get a bit nervous adding `Hashable` to configuration objects in case we want to add non-`Hashable` configuration later (callbacks etc.)

Sources/NIOSSL/NIOSSLHandler+Configuration.swift
```diff
@@ -0,0 +1,38 @@
...
+        ///
+        /// This buffer is not typically deallocated, as it is re-used throughout the lifetime of
+        /// the program. In cases where there are extremely large peak writes that are outliers in
+        /// the code, the buffer may remain excessively large.
+        ///
+        /// Set this value to a lower value to avoid preserving too much memory. This will cause
+        /// ``NIOSSLHandler`` to reallocate memory more often, which can inhibit performance, so
+        /// avoid lowering this value unless you're running into trouble with memory pressure and
+        /// are confident that ``NIOSSLHandler`` is at fault.
+        public var maximumPreservedOutboundBufferCapacity: Int
```
glbrntt, line 32: Do we mind if this is negative? Looks like the effect will be that we'll just clear the buffer each time.

glbrntt (approved):

Sources/NIOSSL/NIOSSLHandler+Configuration.swift
Lukasa: Yeah, I think that's fine. It's a silly thing to do but nothing bad happens.

Sources/NIOSSL/NIOSSLHandler+Configuration.swift
Lukasa: Ah, I feel the other way around: having them be `Hashable` up front helps you consider what a better strategy would be than callbacks in config objects.

glbrntt: Fair!

FranzBusch: I am kinda leaning towards George lately. I have been bitten by making `Configuration` objects `Hashable` lately. So I am also leaning towards not making them `Hashable`. I am curious when people actually want to check them for equality or store them in a dict.

Lukasa: In this particular case I guess I get it, but `TLSConfiguration` is the standard counter-example. It's used to build an `SSLContext`, which is an expensive object, and so it is frequently used as a cache key.

dnadoba (approved):

PR #448, opened 2023-11-01
fclose fix for linux-glibc 2.38

Saljooq:
This is a fix for linux-glibc that would avoid build failure because of wrong type of fclose


Saljooq:
@Lukasa Can you please have a look?

Lukasa:

Lukasa (approved):
Thanks!

PR #449, opened 2023-11-25
Android: update Bionic declarations for nullability annotations added in NDK 26

finagolfin:
Also, force unwrap an `opendir()` call that could be null.

This is needed because [Bionic recently added a bunch of these annotations](https://android.googlesource.com/platform/bionic/+/7d2df8be98f10150b2c30e18186c9ecbb724549a%5E%21/#F0). I made sure this pull doesn't break anything by testing it on linux x86_64, and the force unwrap with the previous NDK 25c too. I used this patch with others to build the Swift toolchain and this package for my Android CI, finagolfin/swift-android-sdk#122.


Lukasa:

Lukasa (approved):

PR #451, opened 2023-12-12
Define _DARWIN_C_SOURCE on Darwin to fix the build in Xcode.

al45tair:
For some reason, we need this in order to build in Xcode, but using SwiftPM directly from the command line works.  Anyway, defining it should be harmless for non-Darwin platforms.


FranzBusch (approved):
Thanks!

PR #452, opened 2023-12-18
Pass empty psk hint to client callback when none is provided by the server

martialln:
Hello,

As I was trying to implement some HTTP calls using `async-http-client` and TLS-PSK authentication, I realised that my server was not sending any PSK hint.

The actual implementation of the clientPSKCallback makes the server psk hint mandatory.

Reading [RFC4279](https://datatracker.ietf.org/doc/html/rfc4279#section-5.2) it seems that on the opposite the server psk hint should be provided only on application that mandate it.

> [5.2](https://datatracker.ietf.org/doc/html/rfc4279#section-5.2).  Identity Hint
>    In the absence of an application profile specification specifying
>    otherwise, servers SHOULD NOT provide an identity hint and clients
>    MUST ignore the identity hint field.  Applications that do use this
>    field MUST specify its contents, how the value is chosen by the TLS
>    server, and what the TLS client is expected to do with the value.

I made this PR to provide an empty string to the callback in the case the server doesn't provide one.
Providing an optional might be a better solution in term of semantic but it would break the interface


Lukasa (commented):
Thanks for this patch, it's a good change! Can I suggest that it might be better for us to deprecate and replace this API with one that handles optional PSK hints better?

This will fundamentally involve deprecating the existing config fields and setting new ones with new types. We could then have the two computed properties forward to an internal stored property that records which one the client used, and prefer to only call that one.

How does that sound?

martialln:
@Lukasa thanks for your feedback.

I agree with you that deprecating the current API and proposing a new one is better on the long term but I was not confortable proposing it directly :)

I will rework the PR in that direction.

Tests/NIOSSLTests/TLSConfigurationTest.swift
```diff
@@ -1258,7 +1274,7 @@ class TLSConfigurationTest: XCTestCase {
         try assertHandshakeSucceeded(withClientConfig: clientConfig, andServerConfig: serverConfig)
     }
     
-    func testTLSPSKFailure() throws {
+    func testTLSPSKFailureDeprecated() throws {
```
Lukasa, line 1277: These tests should be annotated with `@available(*, deprecated)` to suppress warnings.

Lukasa (commented):
Looking generally good, a few notes.

martialln:
Thank you for the feedbacks @Lukasa I have applied your suggestions.

Tests/NIOSSLTests/TLSConfigurationTest.swift
```diff
@@ -1302,6 +1322,182 @@ class TLSConfigurationTest: XCTestCase {
...
+        try assertHandshakeError(withClientConfig: clientConfig, andServerConfig: serverConfig, errorTextContainsAnyOf: ["SSLV3_ALERT_BAD_RECORD_MAC"])
+    }
+    
+    func testTLSPSKNoServerHint() throws {
+        let expectation = expectation(description: "pskClientProvider is called")
+        // This test ensures that different PSKs used on the client and server fail when passed in.
+        let pskClientProvider: NIOPSKClientIdentityProvider = { (context: PSKClientContext) -> PSKClientIdentityResponse in
+            expectation.fulfill()
+            // Ensure server hint is nil
+            XCTAssertEqual(context.hint, nil)
```
Lukasa, line 1470: We need to add equivalent checks on this in the other tests. We should probably also validate the identity and hint in the server callbacks in all tests.

Lukasa (commented):
Looking very good, we're close now.

Lukasa (approved):
Ok cool, I think this looks good.

Lukasa:

Lukasa:
Thank you for your contribution!

martialln:
Thank you for your patience and your reviews

PR #453, opened 2023-12-19
fix https fail on Android

wangx0404:
fix https handshake fail on Android platform.
An example error:

> handshakeFailed(NIOSSL.BoringSSLError.sslError([Error: 268435581 error:1000007d:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED at xxx/.build/checkouts/swift-nio-ssl/Sources/CNIOBoringSSL/ssl/handshake.cc:393]))


Sources/NIOSSL/AndroidCABundle.swift
```diff
@@ -0,0 +1,34 @@
...
+internal let rootCADirectoryPath: String? = locateRootCADirectory()
+
+/// This is a list of root CA directory search paths.
+///
+/// This list contains paths as validated against several distributions. If you are aware of a CA bundle on a specific distribution
+/// that is not present here, please open a pull request that adds the appropriate search path.
+/// Some distributions do not ship CA directories: as such, it is not a problem if a distribution that is present in rootCAFileSearchPaths
+/// is not present in this list.
+private let rootCADirectorySearchPaths = [
+    "/system/etc/security/cacerts",  // Android
```
Lukasa, line 28: @finagolfin does this path look appropriate to you? It's hard for me to validate this logic from here.

finagolfin: I have not done anything with this before, but [it looks like the path is correct](https://stackoverflow.com/questions/75800758/cannot-install-a-system-level-certificate-in-android-14-using-the-same-method-t).

wangx0404:
from Android14, a new path is added for updating CA certificates. so I just updated the commit to read this new path first or else the older one.
refer:https://android.googlesource.com/platform/frameworks/base/+/8b192b19f264a8829eac2cfaf0b73f6fc188d933%5E%21/#F0

Lukasa (approved):

Lukasa:

PR #454, opened 2023-12-19
Add cxx interop build pipeline

gjcairo:
This PR adds a new target to `docker-compose` that builds a new project that depends on this one, with C++ interoperability enabled, with the goal of adding a new CI pipeline that checks the build succeeds.

### Motivation:

To make sure that C++ interoperability can be used on projects that depend on `swift-nio-ssl`.

### Modifications:

- Added a new target to docker-compose that creates a new package that depends on `swift-nio-ssl`, and builds it with C++ interoperability enabled.
- Will request a new CI pipeline that runs this new target.

### Result:

We will ensure this projects builds properly with C++ interoperability enabled.


scripts/cxx-interop-compatibility.sh
```diff
@@ -0,0 +1,20 @@
...
+## Licensed under Apache License v2.0
+##
+## See LICENSE.txt for license information
+## See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+##
+## SPDX-License-Identifier: Apache-2.0
+##
+##===----------------------------------------------------------------------===##
+
+mkdir -p interop
```
glbrntt, line 16: Would be good if we can use temporary directory instead to avoid collisions and not clutter the checkout if this is run locally, for example.

glbrntt (approved):

scripts/cxx-interop-compatibility.sh
```diff
@@ -0,0 +1,68 @@
...
+                .product(name: "NIOSSL", package: "swift-nio-ssl"),
+                .product(name: "NIOTLSServer", package: "swift-nio-ssl"),
+                .product(name: "NIOSSLHTTP1Client", package: "swift-nio-ssl")
+            ],
+            swiftSettings: [.interoperabilityMode(.Cxx)]
+        ),
+        .testTarget(
+            name: "interopTests",
+            dependencies: ["interop"],
+            swiftSettings: [.interoperabilityMode(.Cxx)]
```
Lukasa, line 56: I don't think we need the test target.

Lukasa (changes requested):
One quick nit and then we're good.

yim-lee:

gjcairo:
@yim-lee can thanks for dealing with the pipelines!

scripts/cxx-interop-compatibility.sh
```diff
@@ -0,0 +1,63 @@
...
+    ],
+    targets: [
+        .target(
+            name: "interop",
+            // Depend on all products of swift-nio-ssl to make sure they're all
+            // compatible with cxx interop.
+            dependencies: [
+                .product(name: "NIOSSL", package: "swift-nio-ssl"),
+                .product(name: "NIOTLSServer", package: "swift-nio-ssl"),
+                .product(name: "NIOSSLHTTP1Client", package: "swift-nio-ssl")
```
glbrntt, line 49: I don't think it makes sense to include these as they're executable targets

Lukasa (approved):

PR #456, opened 2024-01-15
Implement SSL_CTX_set_cert_cb

AndrewBarba:
This is a first pass at reviving https://github.com/apple/swift-nio-ssl/pull/311 in order to fix https://github.com/apple/swift-nio-ssl/issues/310

Credit to @TechnikEmpire, I did my best to address the comments in the original PR from @Lukasa 

If the approach looks okay I'll move to testing


Sources/NIOSSL/SSLContext.swift
```diff
@@ -779,6 +791,90 @@ extension NIOSSLContext {
...
+                // here and handle any such errors by signaling the connection to fail.
+                if nativeContextValue != userChosenContext.sslContext {
+                    return SSL_TLSEXT_ERR_NOACK
+                }
+
+                return SSL_TLSEXT_ERR_OK
+            }
+
+            // Make sure we get the server name extracted.
+            guard let cServerName = CNIOBoringSSL_SSL_get_servername(ssl, CNIOBoringSSL.TLSEXT_NAMETYPE_host_name) else {
```
Lukasa: Nit: we shouldn't need the explicit module name:

```suggestion
            guard let cServerName = CNIOBoringSSL_SSL_get_servername(ssl, TLSEXT_NAMETYPE_host_name) else {
```

Sources/NIOSSL/SSLContext.swift
```diff
@@ -779,6 +791,90 @@ extension NIOSSLContext {
...
+                return SSL_TLSEXT_ERR_NOACK
+            }
+
+            let serverName = String(cString: cServerName)
+
+            // Can do nothing about empty indicators.
```
Lukasa, line 848: This isn't true: we should be able to initialize the server name property to `nil`. This will necessitate tweaking the type, but that's fine.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -779,6 +791,90 @@ extension NIOSSLContext {
...
+                return SSL_TLSEXT_ERR_NOACK
+            }
+
+            // Construct extension values
+            let values = NIOSSLClientExtensionValues(serverName: serverName)
+
+            // Issue call to get a new ssl context
+            let futureSSLContext: EventLoopFuture<NIOSSLContext>
+            do {
+                futureSSLContext = try parentSwiftContext.sslContextCallback!(values, parentSwiftContext)
```
Lukasa, line 859: We need a comment above the force-unwrap.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -779,6 +791,90 @@ extension NIOSSLContext {
...
+            } catch {
+                return SSL_TLSEXT_ERR_NOACK
```
Lukasa, line 861: This is a complex ask, but it would be my preference that we didn't throw the user's error away. Ideally we'd persist the error and propagate it down the pipeline, which will require us to store it somewhere on the `SSLContext` object.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -779,6 +791,90 @@ extension NIOSSLContext {
...
+            }
+
+            // Save the future as pending on the parent context
+            parentSwiftContext.futureSSLContext = futureSSLContext
+
+            // Resume the handshake when the future completes
+            futureSSLContext.whenComplete { _ in
+                CNIOBoringSSL_SSL_do_handshake(ssl)
```
Lukasa, line 869: This will cause us to lose track of handshake progress if we call it here. The preference would be to do the same as we do for the custom verify callback, which you can see [here](https://github.com/apple/swift-nio-ssl/blob/7c381eb6083542b124a6c18fae742f55001dc2b5/Sources/NIOSSL/SSLCallbacks.swift#L242-L255).

Sources/NIOSSL/SSLContext.swift
```diff
@@ -779,6 +791,90 @@ extension NIOSSLContext {
...
+            }
+
+            // We must return a negative value to suspend the handshake
+            return -1
```
Lukasa, line 873: You'll also need to add `wantX509Lookup` to the list of reason codes that cause us to assume a handshake is incomplete in `SSLConnection.doHandshake`.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -779,6 +791,90 @@ extension NIOSSLContext {
...
+            guard let ssl = ssl else {
+                return SSL_TLSEXT_ERR_NOACK
+            }
+
+            let parentSwiftContext = NIOSSLContextHelpers.getContextFromRawContextAppData(ssl: ssl)
+
+            // Check if we have a pending context previously set by this callback
+            if let futureSSLContext = parentSwiftContext.futureSSLContext {
+                // If we do and we are executing this callback then the future has already resolved
+                // We can safely call wait() which will resolve immediately
```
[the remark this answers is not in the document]
AndrewBarba: I quickly found out about the deadlock after writing a test. Pushed up a new approach.

And this callback is called twice, the 2nd time after resuming the handshake. I read this in the docs for SSL_CTX_set_cert_cb: https://www.openssl.org/docs/man3.2/man3/SSL_CTX_set_cert_cb.html

Sources/NIOSSL/SSLContext.swift
AndrewBarba: Is it fine for `serverHostname` to just be an empty string and I can remove this check? Or should I check the empty string and pass `nil` when constructing the extension values?

Sources/NIOSSL/SSLContext.swift
AndrewBarba: My new approach uses a saved result on the `SSLContext` so I decided to save the error as a failed result. This feels right, as the synchronous error here is basically the same as the future resolving to an error later. We would want to handle those types of errors in same way. Let me know if that makes sense, and where we can propagate further.

Sources/NIOSSL/SSLContext.swift
AndrewBarba: I think this is resolved in my updated approach

Tests/NIOSSLTests/SSLContextTests.swift
```diff
@@ -0,0 +1,110 @@
...
+                                                 postHandlers: [handshakeWatcher],
+                                                 group: group,
+                                                 connectingTo: serverChannel.localAddress!,
+                                                 serverHostname: sniField)
+        defer {
+            _ = try? clientChannel.close().wait()
+        }
+
+        try handshakeResultPromise.futureResult.wait()
+        XCTAssertEqual(true, true)
```
AndrewBarba, line 104: We're never getting here, still debugging. The new cert_cb is correctly going through the two phases and returning `1` in the 2nd case, but the handshake is not completing.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -779,6 +788,96 @@ extension NIOSSLContext {
...
+            futureSSLContext.whenComplete { result in
+                // Save the result to the parent context to be inspected after resuming
+                parentSwiftContext.sslContextCallbackResult = result
+
+                // Schedule the handshake to resume. This must be scheduled on next tick to avoid resuming
+                // before suspending below. Once the handshake is resumed the entire SSL_CTX_set_cert_cb
+                // callback will be executed again, and this time the ssl context result will be saved to
+                // the parent context allowing us to set the new context.
+                futureSSLContext.eventLoop.execute {
+                    CNIOBoringSSL_SSL_do_handshake(ssl)
```
AndrewBarba, line 871: I think this call here is a problem. This properly resumes and I see the callback running a 2nd time but nothing on the Swift side knows about it. I think this line need to somehow signal to `SSLConnection` to `doHandshake()`

AndrewBarba:
@Lukasa Thanks so much for the comments earlier. I've since changed quite a bit now that I have a better understanding of whats going on. I was able to create a test that at least ensures the handshake is completing successfully, where as before I was never signaling to the connection that our async lookup was complete. Im sure there's still a lot more testing and error handling we can do but I wanted to pause here and wait for more feedback.

Sources/NIOSSL/SSLContext.swift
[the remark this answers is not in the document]
AndrewBarba: I also wanted to follow up that in the current state its correctly calling the callback twice after `doHandshake()` and it seems to be working as expected (as far as the behavior of this callback)

Sources/NIOSSL/SSLContext.swift
```diff
@@ -211,6 +203,72 @@ private func clientPSKCallback(ssl: OpaquePointer?,
...
+    if let sslContextCallbackResult = parentSwiftContext.sslContextCallbackResult {
+        let userChosenContext: NIOSSLContext
+        switch sslContextCallbackResult {
+        case .success(let value):
+            userChosenContext = value
+        case .failure:
+            return SSL_TLSEXT_ERR_NOACK
+        }
+
+        // The user may or may not retain a reference to the NIOSSLContext
```
AndrewBarba, line 220: Is this necessary now that we are using an EventLoopFuture? We are holding on to the future and the future I think holds a strong reference to the context right?

Sources/NIOSSL/SSLConnection.swift
```diff
@@ -31,6 +31,7 @@ internal let sslConnectionExDataIndex = CNIOBoringSSL_SSL_get_ex_new_index(0, ni
 /// handle the "must wait for more data" case by calling it out directly.
 enum AsyncOperationResult<T> {
     case incomplete
+    case lookup
```
Lukasa, line 34: I don't think this is needed: the existing `incomplete` flows should work fine, shouldn't they?

AndrewBarba: I don't think they do. There are multiple cases where `incomplete` branch will be called, and then you end up in a state where you have to track if you added a complete handler to the future already. Seemed much cleaner to have an explicit branch for lookups. The lookup case felt important enough to me to not hide those details

Lukasa: We can attach the completion handler inside the callback itself though right?

AndrewBarba: If you attach complete handler inside the callback then there's no way to signal back out to `SSLConnection` that it should continue the handshake. You could resume the handshake inside the complete handler (which I originally had) but then `SSLConnection` doesn't know about it and never fires the complete events.

Lukasa: You can follow the pattern we use for the custom certificate verification code: https://github.com/apple/swift-nio-ssl/blob/7c381eb6083542b124a6c18fae742f55001dc2b5/Sources/NIOSSL/SSLCallbacks.swift#L240-L255.

Lukasa: I think the cert_cb registration still gets the `SSLConnection` as something it can pull out, so you can probably still do this from where it is.

AndrewBarba: I made a mistake in my last couple comments, its not `SSLConnection` that we need access to in order to complete the handshake, it's `SSLHandler` that needs to call `doHandshakeStep` which in turns takes result from `SSLConnection.doHandshake` and fires the necessary events.

AndrewBarba:
@Lukasa Sorry for the delay, the PR is cleaned up to remove that lookup state and keep everything inside the callback and new context manager

Sources/NIOSSL/SSLCallbacks.swift
```diff
@@ -173,6 +173,77 @@ public struct PSKClientIdentityResponse: Sendable {
...
+public typealias NIOSSLContextCallback = @Sendable (NIOSSLClientExtensionValues, NIOSSLContext) -> EventLoopFuture<NIOSSLContext>
+
+/// A struct that provides helpers for working with a NIOSSLContextCallback.
+internal class CustomContextManager {
+    internal enum State {
+        case loading(_ future: EventLoopFuture<NIOSSLContext>)
+        case complete(_ result: Result<NIOSSLContext, Error>)
+    }
+
+    internal private(set) var callback: NIOSSLContextCallback
```
Lukasa, line 203: This is never mutated.

Sources/NIOSSL/SSLCallbacks.swift
```diff
@@ -173,6 +173,77 @@ public struct PSKClientIdentityResponse: Sendable {
...
+/// server context being used for the handshake.
+///
+/// Within this callback, the user can create and return a new server `NIOSSLContext` for the given host,
+/// return the same context provided if it is sufficient to complete the handshake, or return `throw` in the event
+/// of an exception.
+///
+public typealias NIOSSLContextCallback = @Sendable (NIOSSLClientExtensionValues, NIOSSLContext) -> EventLoopFuture<NIOSSLContext>
+
+/// A struct that provides helpers for working with a NIOSSLContextCallback.
+internal class CustomContextManager {
```
Lukasa, line 197: This seems like it should be able to be a `struct`, if we hoist most of the code currently written inside the `SSLContext.swift` file into a `mutating` method here.

Sources/NIOSSL/SSLCallbacks.swift
```diff
@@ -173,6 +173,77 @@ public struct PSKClientIdentityResponse: Sendable {
...
+///
+/// When set, this callback will be invoked once per TLS hello. The provided `NIOSSLClientExtensionValues` will contain the
+/// host name indicated in the TLS client hello, while the provided `NIOSSLContext` will be the current
+/// server context being used for the handshake.
+///
+/// Within this callback, the user can create and return a new server `NIOSSLContext` for the given host,
+/// return the same context provided if it is sufficient to complete the handshake, or return `throw` in the event
+/// of an exception.
+///
+public typealias NIOSSLContextCallback = @Sendable (NIOSSLClientExtensionValues, NIOSSLContext) -> EventLoopFuture<NIOSSLContext>
```
Lukasa, line 194: We should consider providing an `EventLoopPromise` here instead of returning an `EventLoopFuture`. The advantage of this pattern is that users don't have to close over an appropriate `EventLoop` type to create the future, making it easier to write code that doesn't thread-hop.

Sources/NIOSSL/SSLCallbacks.swift
```diff
@@ -173,6 +173,77 @@ public struct PSKClientIdentityResponse: Sendable {
...
+
+        // Ensure we have a parent handler to resume the connection
+        guard let parentHandler = connection.parentHandler else {
+            preconditionFailure("""
+                SSL_CTX_set_cert_cb was executed without a parent handler assigned to the connection.
+                This should not be possible, please file an issue.
+            """)
+        }
+
+        future.hop(to: eventLoop).whenComplete { result in
```
Lukasa, line 235: If we create the future, we can elide the `hop(to:)` here.

Sources/NIOSSL/SSLContext.swift
Lukasa: That's true, but reveals a concern: we can end up with a reference cycle that needs to be explicitly broken. So I think we want to keep the code we have here, and drop the stored future. It's not that useful to us anyway, because we only want the ability to attach a callback to it once, allowing us to safely drop it once we've done that.

This also takes us closer to the `CustomVerifyManager` pattern.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -211,6 +203,78 @@ private func clientPSKCallback(ssl: OpaquePointer?,
...
+            return SSL_TLSEXT_ERR_NOACK
+        }
+        
+        // We must return 1 to signal a successful load of the new context
+        return 1
+    case .loading:
+        preconditionFailure("""
+            SSL_CTX_set_cert_cb was executed in a pending state.
+            This should not be possible, please file an issue.
+        """)
```
Lukasa, line 274: I think this is gonna be too aggressive. We may, for other reasons, end up calling `doHandshake` again (e.g. because more data arrives). We should tolerate being called again, and just returning negative values.

AndrewBarba:
@Lukasa Just wanted to checkin and see if you had any other feedback. Sorry if this isn't heading in the right direction, don't want to take up too much of your time on it.

Lukasa:
Sorry for the delay @AndrewBarba, it's not a reflection on your work: I was just on vacation! 🏝️

Sources/NIOSSL/SSLCallbacks.swift
```diff
@@ -173,6 +173,84 @@ public struct PSKClientIdentityResponse: Sendable {
     }
 }
 
+public struct NIOSSLClientExtensionValues: Hashable {
```
Lukasa, line 176: Can we add some doc comments to this?

Sources/NIOSSL/SSLCallbacks.swift
```diff
@@ -173,6 +173,84 @@ public struct PSKClientIdentityResponse: Sendable {
...
+        // Ensure we have an event loop
+        guard let eventLoop = connection.eventLoop else {
+            preconditionFailure("""
+                SSL_CTX_set_cert_cb was executed without an event loop assigned to the connection.
+                This should not be possible, please file an issue.
+            """)
+        }
+
+        // Ensure we have a parent handler to resume the connection
+        guard let parentHandler = connection.parentHandler else {
```
Lukasa, line 235: We don't need to crash in this context: if we can't resume the connection because we have no parent handler, there are lots of good reasons for it.

Sources/NIOSSL/SSLCallbacks.swift
```diff
@@ -173,6 +173,84 @@ public struct PSKClientIdentityResponse: Sendable {
...
+    internal private(set) var state: State
+
+    init(callback: @escaping NIOSSLContextCallback) {
+        self.callback = callback
+        self.state = .notStarted
+    }
+}
+
+extension CustomContextManager {
+    internal enum State {
```
Lukasa, line 209: I think it's generally nicer if this is private to `CustomContextManager`, and `loadContext` becomes responsible for correctly managing the state. It makes it much easier to confirm that the API to this type is sound.

I do recommend having a little computed property that tells the outer code whether it needs to construct a `NIOSSLClientExtensionValues`, but that should be as much of the internal state as this method shows. In fact, we could cache the values, and have that computed property be `var extensionValues: NIOSSLClientExtensionValues?`. If there are no cached values we'll build them, otherwise we just use the cached one.

Sources/NIOSSL/SSLCallbacks.swift
```diff
@@ -173,6 +173,84 @@ public struct PSKClientIdentityResponse: Sendable {
...
+///
+/// When set, this callback will be invoked once per TLS hello. The provided `NIOSSLClientExtensionValues` will contain the
+/// host name indicated in the TLS client hello, while the provided `NIOSSLContext` will be the current
+/// server context being used for the handshake.
+///
+/// Within this callback, the user can create and return a new server `NIOSSLContext` for the given host,
+/// return the same context provided if it is sufficient to complete the handshake, or return `throw` in the event
+/// of an exception.
+///
+public typealias NIOSSLContextCallback = @Sendable (NIOSSLClientExtensionValues, NIOSSLContext) -> EventLoopPromise<NIOSSLContext>
```
Lukasa, line 194: This function should accept a promise as an argument, not return one.

Sources/NIOSSL/SSLCallbacks.swift
```diff
@@ -173,6 +173,84 @@ public struct PSKClientIdentityResponse: Sendable {
...
+        guard let parentHandler = connection.parentHandler else {
+            preconditionFailure("""
+                SSL_CTX_set_cert_cb was executed without a parent handler assigned to the connection.
+                This should not be possible, please file an issue.
+            """)
+        }
+
+        promise.futureResult.whenComplete { result in
+            // Update our state with complete result
+            connection.parentContext.customContextManager?.state = .complete(result)
```
Lukasa, line 244: We can't safely do this here: we may hit an exclusivity violation, as `whenComplete` can in some cases execute immediately. This has to be moved into the `eventLoop.execute` block.

Sources/NIOSSL/SSLCallbacks.swift
AndrewBarba: @Lukasa Do you mind expanding on this? Since the initial PR this callback was meant to be a user-land callback where the user asynchronously provides us a new context. How could the argument be a future promise?

AndrewBarba: Is the idea that we create the promise and then in user code they can fulfill it?

AndrewBarba:
No problem @Lukasa hope you enjoyed! I addressed your comments, ready for another round of review.

Sources/NIOSSL/SSLCallbacks.swift
```diff
@@ -173,6 +173,108 @@ public struct PSKClientIdentityResponse: Sendable {
...
+            let promise = eventLoop.makePromise(of: NIOSSLContext.self)
+            self.callback(values, promise)
+
+            // After invoking the user callback we can update our state to pending
+            self.state = .pendingResult
+
+            // Ensure we execute any completion on the next event loop tick
+            // This ensures that we suspend before calling resume
+            eventLoop.execute {
+                promise.futureResult.whenComplete { result in
```
Lukasa, line 267: These two calls are inverted: we need to `execute` inside the `whenComplete`, not the other way around.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -211,6 +203,58 @@ private func clientPSKCallback(ssl: OpaquePointer?,
...
+    // Begin loading a new context
+    let result = contextManager.loadContext(ssl: ssl)
+
+    switch result {
+    case .none:
+        // If we dont have a result yet then we must suspend
+        return -1
+    case .failure:
+        // If loading a context failed then we must signal as such
+        return SSL_TLSEXT_ERR_NOACK
```
Lukasa, line 228: This looks wrong. The `SSL_TLSEXT_ERR_*` constants are all positive integers. Returning them here will cause this to look like _success_.

I think we need a failure case test here, as that should have caught this behaviour. Based on code reading, it looks like we need to return 0 to fail here.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -211,6 +203,58 @@ private func clientPSKCallback(ssl: OpaquePointer?,
...
+        // 1 and thus they can re-use the same context for multiple connections.
+        let nativeContextValue = withExtendedLifetime(userChosenContext) {
+            return CNIOBoringSSL_SSL_set_SSL_CTX(ssl, userChosenContext.sslContext)
+        }
+
+        // CNIOBoringSSL_SSL_set_SSL_CTX should return a pointer to the very same context it was
+        // told to assign on success. On error, this will not happen, so we check for equality
+        // here and handle any such errors by signaling the connection to fail.
+        if nativeContextValue != userChosenContext.sslContext {
+            return SSL_TLSEXT_ERR_NOACK
```
Lukasa, line 250: Same note here, this should probably return 0.

Sources/NIOSSL/SSLCallbacks.swift
```diff
@@ -173,6 +173,108 @@ public struct PSKClientIdentityResponse: Sendable {
...
+///
+/// When set, this callback will be invoked once per TLS hello. The provided `NIOSSLClientExtensionValues` will contain the
+/// host name indicated in the TLS client hello, while the provided `NIOSSLContext` will be the current
+/// server context being used for the handshake.
+///
+/// Within this callback, the user can create and return a new server `NIOSSLContext` for the given host,
+/// return the same context provided if it is sufficient to complete the handshake, or return `throw` in the event
+/// of an exception.
+///
+public typealias NIOSSLContextCallback = @Sendable (NIOSSLClientExtensionValues, EventLoopPromise<NIOSSLContext>) -> Void
```
Lukasa, line 208: I increasingly agree with you @AndrewBarba that this is maybe a bit too big. You proposed returning a new TLSConfiguration object instead and applying the delta. That's a nice idea, but I wonder if we should instead define a third object (maybe a "post extension config change" object) with a trimmed-down set of fields. That would let us extend it in the future, and avoids some config that would be very hard for us to change. We could start out only with the cert chain.

What do you think?

AndrewBarba: @Lukasa Big +1 from me on this idea. I really like a dedicated struct with just the deltas we want to support. Going to give this a shot later this week

Lukasa: Fantastic, please let me know when that's done and I'll do another review pass. I'm trying to minimise the number of incremental reviews I do this week, because it's WWDC week and I'm slammed.

AndrewBarba:
@Lukasa Hope you had a great WWDC week, I just pushed up first go at a new extension change object. Let me know what you think

Sources/NIOSSL/SSLCallbacks.swift
```diff
@@ -173,6 +173,117 @@ public struct PSKClientIdentityResponse: Sendable {
...
+    ///
+    /// - parameter serverHostname: The hostname value from the SNI extension.
+    public init(serverHostname: String?) {
+        self.serverHostname = serverHostname
+    }
+}
+
+/// A structure representing changes to the SSL/TLS configuration that can be applied 
+/// after the client hello message extensions have been processed.
+public struct NIOSSLContextPostExtensionConfigurationChange {
```
Lukasa, line 200: I think this type looks totally reasonable. Instead of calling it a "change", as it's not a diff, let's call it a "configuration override".

Don't forget to add a `public init()` or users won't be able to construct it. Let's avoid having the init take parameters, just have it default to all-nil.

AndrewBarba: @Lukasa Just pushed this up, but I added the params to constructor (default nil) cause I think its a bit more ergonomic from user perspective to pass in those values if you have them instead of making a new `var` and mutating before returning.

Lukasa: While I agree it's more ergonomic, as we add more fields to the structure we have to add new initializers to make them set that way (for API stability reasons we can't add new arguments with default values). That has the effect of making the type explode with values. In this case, as we know we're going to want to add more, we should just start by putting everything in that footing and get people writing the code we expect.

AndrewBarba: Ah thats helpful context, just pushed up empty init. Do you think there's any value in letting user return `nil` in the callback? Obviously can just return an empty `.init()` as well if no changes are needed.

AndrewBarba: Nice just pushed

AndrewBarba:
@Lukasa Friendly nudge to keep this moving. Let me know what I can tackle next

Sources/NIOSSL/SSLContext.swift
```diff
@@ -213,6 +205,49 @@ private func clientPSKCallback(ssl: OpaquePointer?,
...
+    // Begin loading a new context
+    let result = contextManager.loadContext(ssl: ssl)
+
+    switch result {
+    case .none:
+        // If we dont have a result yet then we must suspend
+        return -1
+    case .failure:
+        // If loading a context failed then we must signal as such
+        return 0
```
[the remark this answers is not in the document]
AndrewBarba: Took a stab at this here: https://github.com/apple/swift-nio-ssl/pull/456/commits/694f5c28c602fc51692c466bbf6c651082a31376

Working on a new test as well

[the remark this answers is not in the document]
AndrewBarba: @Lukasa Error test case has been added

Tests/NIOSSLTests/SSLContextTests.swift
```diff
@@ -0,0 +1,199 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2021 Apple Inc. and the SwiftNIO project authors
```
Lukasa, line 5:
```suggestion
// Copyright (c) 2017-2024 Apple Inc. and the SwiftNIO project authors
```

Tests/NIOSSLTests/SSLContextTests.swift
```diff
@@ -0,0 +1,199 @@
...
+//===----------------------------------------------------------------------===//
+
+import XCTest
+import NIOCore
+import NIOPosix
+import NIOTLS
+import NIOEmbedded
+@testable import NIOSSL
+
+class SSLContextTest: XCTestCase {
```
Lukasa, line 22:
```suggestion
final class SSLContextTest: XCTestCase {
```

Lukasa:
So you can use the docker-compose files in the repo to generate a simulacrum of the CI (for now, until we cut over to GHA). The steps are:

```
docker-compose -f docker/docker-compose.yaml -f docker/docker-compose.2204.510.yaml build
docker-compose -f docker/docker-compose.yaml -f docker/docker-compose.2204.510.yaml run test
```

That latter command reveals the issue:

```
/swift-nio-ssl/Tests/NIOSSLTests/TLSConfigurationTest.swift:824:98: error: 'file' is deprecated: Use 'NIOSSLPrivateKeySource.privateKey(NIOSSLPrivateKey)' to set private key
        var config = TLSConfiguration.makeServerConfiguration(certificateChain: [], privateKey: .file("fake.file"))
                                                                                                 ^
/swift-nio-ssl/Tests/NIOSSLTests/TLSConfigurationTest.swift:824:98: error: 'file' is deprecated: Use 'NIOSSLPrivateKeySource.privateKey(NIOSSLPrivateKey)' to set private key
        var config = TLSConfiguration.makeServerConfiguration(certificateChain: [], privateKey: .file("fake.file"))
```

AndrewBarba:
Thanks @Lukasa I think this is ready for another review.

Lukasa (approved):
Great, really nice job here @AndrewBarba! Thanks for persevering with the reviews, I think it left this patch in a great state. We're good to go.

PR #459, opened 2024-04-08
Add empty privacy manifest

glbrntt:
Motivation:

NIOSSL distributes a copy of BoringSSL, which means it needs a privacy manifest.

Modifications:

- Add a privacy manifest

Result:

NIOSSL shouldn't be rejected from app submission by not declaring a privacy manifest.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -46,6 +46,15 @@ func generateDependencies() -> [Package.Dependency] {
     }
 }
 
+// This doesn't work when cross-compiling: the privacy manifest will be included in the Bundle and
+// Foundation will be linked. This is, however, strictly better than unconditionally adding the
+// resource.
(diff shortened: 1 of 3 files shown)


Package.swift
```diff
@@ -46,6 +46,15 @@ func generateDependencies() -> [Package.Dependency] {
     }
 }
 
+// This doesn't work when cross-compiling: the privacy manifest will be included in the Bundle and
+// Foundation will be linked. This is, however, strictly better than unconditionally adding the
+// resource.
+#if os(Linux) || os(Android)
```
Lukasa, line 52: Let's invert the conditional use `#canImport(Darwin)` as the check.

Lukasa (approved):

PR #460, opened 2024-05-01
Handle error when private key file type is unknown

clintonpi:
Motivation:

When the private key file for a TLS configuration has an unknown type, the program crashes. Hence, users are unable to catch the error.

Modifications:

A new error was added to `NIOSSLExtraError` and the program throws that error instead of crashing.

Result:

The program now throws an error instead of crashing.


glbrntt (approved):

Lukasa (approved):
Really nice change @clintonpi, thanks!

PR #464, opened 2024-05-14
Use case-insensitive comparison when validating cert domain name

### LINKED ISSUE #463: Case-insensitive identity verification
[RFC 5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.6) states that *while uppercase and lowercase letters are allowed in domain names, no significance is attached to the case*, meaning the case should be ignored when comparing domain names.

Most certificates don't have any uppercase letter in their domain name. But some do.

SwiftNIO SSL converts the server hostname to lowercase but doesn't convert the certificate's CN or SAN. Therefore, the domain name comparison always fails when the certificate has uppercase letters in its domain name.

### Steps to reproduce

* Copy [`cert.pem`](https://github.com/apple/swift-nio-ssl/files/15309969/cert.pem.txt) and [`key.pem`](https://github.com/apple/swift-nio-ssl/files/15309982/key.pem.txt) to the current directory.

* Run the server.
  ```sh
  swift run NIOTLSServer
  ```

* Try to connect to the server.
  ```sh
  swift run NIOSSLHTTP1Client https://localhost:4433 cert.pem key.pem cert.pem
  ```

### Expected result

A parsing error `invalid constant string` is thrown (the server is not a valid HTTP server).

### Actual  result

 `NIOSSLExtraError.failedToValidateHostname: Couldn't find localhost in certificate from peer` is thrown.

baarde:
This PR fixes #463.

### DIFF
--- a/Sources/NIOSSL/IdentityVerification.swift
+++ b/Sources/NIOSSL/IdentityVerification.swift
@@ -71,6 +71,11 @@ extension Collection {
     }
 }
 
+extension Sequence<UInt8> {
+    fileprivate func caseInsensitiveElementsEqual(_ other: some Sequence<UInt8>) -> Bool {
+        self.elementsEqual(other) { $0.lowercased() == $1.lowercased() }
+    }
+}
 
 extension UInt8 {
     /// Whether this character is a valid DNS character, which is the ASCII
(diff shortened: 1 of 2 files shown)


Lukasa:

Lukasa (approved):
Again, a really lovely patch, thanks!

PR #468, opened 2024-07-12
Use Bionic module and new Android overlay in Swift 6

finagolfin:
Swift 6 has switched to this new Android overlay, swiftlang/swift#74758, which I've been testing on this repo without a problem using my daily Android CI for weeks now, finagolfin/swift-android-sdk#151.

### DIFF
--- a/Sources/NIOSSL/SSLContext.swift
+++ b/Sources/NIOSSL/SSLContext.swift
@@ -20,8 +20,10 @@ import NIOCore
 import Darwin.C
 #elseif canImport(Musl)
 import Musl
-#elseif os(Linux) || os(FreeBSD) || os(Android)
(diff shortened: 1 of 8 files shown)


Sources/NIOSSL/SSLContext.swift
```diff
@@ -44,7 +46,7 @@ internal enum FileSystemObject {
         }
 
 #if os(Android) && arch(arm)
-        return (statObj.st_mode & UInt32(Glibc.S_IFDIR)) != 0 ? .directory : .file
```
finagolfin, line 47: This `Glibc` scoping on Android armv7 was added for whatever reason in #45, but my testing now shows it isn't needed, so I removed it here and below.

Lukasa (approved):

PR #471, opened 2024-07-22
Fix incorrect filename

kohenkatz:
PR #459 added this privacy manifest, but refers to it by the wrong name here.

Motivation:

When building a project that references this library (via Tuist, but I assume it would be true for other build tools too), I get the following warning:

```
No files found at: /Users/moshe/projects/myproject/Tuist/.build/checkouts/swift-nio-ssl/Sources/NIOSSL/ProcessInfo.xcprivacy
```

Modifications:

Fixed the filename in `Package.swift`.

Result:

No more warnings.


glbrntt:

glbrntt (approved):
Thank you!

Lukasa (approved):

PR #474, opened 2024-07-26
Mark .file case deprecated for NIOSSLPrivateKeySource

ayush1794:
swift-nio-ssl expects file to in specific format when loading the PrivateKey from a file baed on the extension of the file. This is prone to errors, since there is no standard agreement on the extension and format of the file. 
An example of a mismatch is `.key` swift-nio-ssl tries to load such file as DER, though these can be in PEM format too.

We could based on the binary contents of the file provided infer the format of the file, but it doesn't seem like a logic this library should own. The clients should be able to load the file themselves. 

In this PR  `.file` case is marked deprecated and relevant deprecation flags and documentation is added to point to the preferred option.

### DIFF
--- a/Tests/NIOSSLTests/SSLPrivateKeyTests.swift
+++ b/Tests/NIOSSLTests/SSLPrivateKeyTests.swift
@@ -237,6 +237,18 @@ class SSLPrivateKeyTest: XCTestCase {
         }
     }
 
+    @available(*, deprecated, message: "`.file` NIOSSLPrivateKeySource option deprecated")
+    func testMissingPassword() {
+        let configuration = TLSConfiguration.makeServerConfiguration(
+            certificateChain: [],
+            privateKey: .file(SSLPrivateKeyTest.passwordPemKeyFilePath)
+        )
+
+        XCTAssertThrowsError(try NIOSSLContext(configuration: configuration)) { error in
+            XCTAssertEqual(.failedToLoadPrivateKey, error as? NIOSSLError)
+        }
+    }
+
     func testECKeysWorkProperly() throws {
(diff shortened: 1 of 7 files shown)


glbrntt:

glbrntt (commented):
Are there no tests using `.file`? If there are we should deprecate the tests (and their test classes) to suppress the deprecation warning.

Tests/NIOSSLTests/SSLPrivateKeyTests.swift
```diff
@@ -237,17 +237,6 @@ class SSLPrivateKeyTest: XCTestCase {
         }
     }
 
-    func testMissingPassword() {
```
glbrntt, line 240: We can still test this path (the functionality isn't going away), but we'll need to make the tests as deprecated too (they'll still be run).

Tests/NIOSSLTests/TLSConfigurationTest.swift
```diff
@@ -1288,19 +1288,6 @@ class TLSConfigurationTest: XCTestCase {
         serverConfig.pskHint = "pskHint"
         try assertHandshakeError(withClientConfig: clientConfig, andServerConfig: serverConfig, errorTextContainsAnyOf: ["SSLV3_ALERT_BAD_RECORD_MAC"])
     }
-
-    func testUnknownPrivateKeyFileType() throws {
```
glbrntt, line 1292: This test is still valid, we'll just need to mark it as deprecated

glbrntt (changes requested):

glbrntt (approved):
LGTM, thanks @ayush1794

PR #485, opened 2024-10-17
Add support for supplying supported curves to `TLSConfiguration`

mads-frameo:
This PR adds the support to specify what ECs the handler supports when doing the handshake.

```swift
var config = TLSConfiguration.makeClientConfiguration()
config.curves = [.x25519]
```

### DIFF
--- a/Tests/NIOSSLTests/TLSConfigurationTest.swift
+++ b/Tests/NIOSSLTests/TLSConfigurationTest.swift
@@ -832,6 +832,7 @@ class TLSConfigurationTest: XCTestCase {
     func testCompatibleCurves() throws {
         var clientConfig = TLSConfiguration.makeClientConfiguration()
         clientConfig.curves = [.x25519]
+        clientConfig.cipherSuiteValues = [.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384]
         clientConfig.maximumTLSVersion = .tlsv12
         clientConfig.certificateVerification = .noHostnameVerification
         clientConfig.trustRoots = .certificates([TLSConfigurationTest.cert1])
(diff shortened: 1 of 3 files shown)


Lukasa (commented):
Thanks for this patch! You've raised a few questions about the API shape that I'd like us to tackle.

In general, I think the right shape here is to follow what we've done with `NIOTLSCipher` and `SignatureAlgorithm`, which is to use a strongly-typed API that provides us with nice symbolic names.

From there we should probably use the preferred BoringSSL API for this, which is `SSL_CTX_set1_group_ids`. This would also encourage us to use the UInt16 group IDs.

Does that sound reasonable?

mads-frameo:
@Lukasa Definitely! Should we follow the same design as with `NIOTLSCipher` and also provide a string-based API `config.curves = ""`? or just provide with the strongly-typed?

Lukasa:

Tests/NIOSSLTests/TLSConfigurationTest.swift
```diff
@@ -829,6 +829,60 @@ class TLSConfigurationTest: XCTestCase {
...
+            certificateChain: [.certificate(TLSConfigurationTest.cert1)],
+            privateKey: .privateKey(TLSConfigurationTest.key1)
+        )
+        serverConfig.curves = [.x25519, .secp256r1]
+        serverConfig.maximumTLSVersion = .tlsv12
+        serverConfig.certificateVerification = .none
+        try assertHandshakeSucceeded(withClientConfig: clientConfig, andServerConfig: serverConfig)
+    }
+
+    func testNonCompatibleCurves() throws {
```
Lukasa, line 868: I don't think this test currently passes: it looks like the two have compatible curves.

mads-frameo: You are right! fixed it:)

Lukasa (approved):
Fantastic, thanks!

Issue #484, opened 2024-10-13
expose CNIOBoringSSL


xlc:
We would like to use some low level openssl features directly and wanting to use `CNIOBoringSSL`. What is the reason of not exposing CNIOBoringSSL? If it is just some compiler warnings, some compiler flags should address them. Happy to contribute if needed.

Lukasa:
Hi @xlc, thanks for asking. We discussed this on the Swift forums when we [originally proposed cutting over to BoringSSL](https://forums.swift.org/t/rfc-moving-swiftnio-ssl-to-boringssl/18280#packaging-10), but I'll reiterate here:

> BoringSSL does not provide any API or ABI stability guarantees. They are free to, and indeed do, break APIs regularly. If BoringSSL were exposed as as Swift Package Manager package, it would bump its major version number every release. This would make it impossible for the wider community to depend on: different packages would be pinning to different major versions, and dependency resolution would immediately fail.
>
> swift-nio-ssl will present an API that does not contain any part of BoringSSL. If users need to use libssl or libcrypto themselves, they will need to bring their own copy of libssl.

When we update our copy of BoringSSL, we fairly frequently get compile errors that we have to fix. This is despite the fact that we only use a relatively small amount of the API surface. Across the whole of the API surface we'd never be able to revise it.

What features do you want from BoringSSL? We may have good solutions for them in Swift in other packages.

xlc:
Thanks for the explanation. That's a bit unfortunate but make sense.
For the use case here, we are implementing [this protocol](https://github.com/zdave-parity/jam-np/blob/main/simple.md) over QUIC. It have some requirements of the X509 certificate, more specifically, require the peers using a published Ed25519 key. So we will need to be able to generate and verify such certificates.

Moreover, I am using https://github.com/microsoft/msquic as the QUIC implementation, and it requires PKCS12 format if I want to pass the keys in-memory (i.e. not loading from a PEM file).

I discovered CNIOBoringSSl from this issue: https://github.com/apple/swift-certificates/issues/114

Lukasa:
Ok, so let's turn that into a series of feature requests on the higher-level projects. I think there are two features we need:

1. Add support for Ed25519 keys in swift-certificates. This should be very do-able so long as we get the key formats right, and I expect you will have some sample certs laying around somewhere.
2. Add support for serializing PKCS#12 in swift-nio-ssl. In the short term, `PKCS12_create` is sufficient for this.

[...]

PR #486, opened 2024-10-24
Add API to create PKCS#12

gjcairo:
## Motivation
It would be handy to provide an API to create PKCS#12 files from a list of `NIOSSLCertificates` and a `NIOSSLPrivateKey`.
This would be particularly useful when dealing with Network.framework/NIOTransportServices/Security.framework, which use `SecIdentity`s for SSL. Two particular use cases are https://github.com/apple/swift-nio-ssl/issues/484#issuecomment-2410375188 and `grpc-swift-nio-transport`, which would use this API for testing the NIOTS transport implementation.

## Modifications
This PR adds a static method to `NIOSSLPKCS12Bundle` that creates a PKCS#12 file from the given array of certificates + private key, and returns it as an array of bytes.

## Result
PKCS#12 files can be created using NIOSSL.


Sources/NIOSSL/SSLPKCS12Bundle.swift
```diff
@@ -158,6 +158,86 @@ public struct NIOSSLPKCS12Bundle: Hashable {
...
+    ///   - certificates: An array of certificates to include in this PKCS#12. Must contain at least one certificate.
+    ///   - privateKey: The private key associated to the first certificate in `certificates`.
+    ///   - passphrase: The password with which to protect this PKCS#12 file.
+    ///   - name: The name to give this PKCS#12 file.
+    /// - Returns: An array of bytes making up the PKCS#12 file.
+    public static func makePKCS12<Bytes: Collection>(
+        certificates: [NIOSSLCertificate],
+        privateKey: NIOSSLPrivateKey,
+        passphrase: Bytes,
+        name: Bytes
```
Lukasa, line 179: We don't need to provide a name, and I'd argue we shouldn't bother. If we're going to enable this, it should be optional, and needs a different generic type parameter to the passphrase.

Sources/NIOSSL/SSLPKCS12Bundle.swift
```diff
@@ -158,6 +158,86 @@ public struct NIOSSLPKCS12Bundle: Hashable {
...
+    ) throws -> [UInt8] where Bytes.Element == UInt8 {
```
Lukasa, line 180: I think this API does too much. Should we split it out? We can have a memberwise constructor on the PKCS12, and then a serialize function.

Sources/NIOSSL/SSLPKCS12Bundle.swift
```diff
@@ -158,6 +158,86 @@ public struct NIOSSLPKCS12Bundle: Hashable {
...
+        guard let mainCertificate = certificates.first else {
+            preconditionFailure("At least one certificate must be provided")
+        }
+
+        let certificateChainStack = CNIOBoringSSL_sk_X509_new(nil)
+        for additionalCertificate in certificates.dropFirst() {
+            let result = additionalCertificate.withUnsafeMutableX509Pointer { certificate in
+                CNIOBoringSSL_sk_X509_push(certificateChainStack, certificate)
```
Lukasa, line 188: This gets the refcount wrong. We need to increase the refcount of the certificate pointer here, using `X509_up_ref`

Sources/NIOSSL/SSLPKCS12Bundle.swift
```diff
@@ -158,6 +158,86 @@ public struct NIOSSLPKCS12Bundle: Hashable {
...
+            }
+            if result == 0 {
+                fatalError("Failed to add certificate to chain")
+            }
+        }
+
+        let pkcs12 = try passphrase.withSecureCString { passphrase in
```
Lukasa, line 195: There's no matching `PKCS12_free` here.

Sources/NIOSSL/SSLPKCS12Bundle.swift
gjcairo: Do you mean having an `init(certificateChain: [NIOSSLCertificate], privateKey: NIOSSLPrivateKey)` on `NIOSSLPKCS12Bundle`, and then a `serialize(passphrase: Bytes) -> [UInt8]`?

Sources/NIOSSL/SSLPKCS12Bundle.swift
gjcairo: I'll just get rid of it, don't think it serves any purpose really.

Sources/NIOSSL/SSLPKCS12Bundle.swift
gjcairo: Ah, the joys of manual memory management. 
Should we decrease the refcount after we've created the PKCS12?

Sources/NIOSSL/SSLPKCS12Bundle.swift
Lukasa: Yeah.

Sources/NIOSSL/SSLPKCS12Bundle.swift
Lukasa: Yes, we're also missing the free of the `sk_509`. Done correctly, the free call is `sk_X509_pop_free` with the second argument as `X509_free`.

Sources/NIOSSL/SSLPKCS12Bundle.swift
```diff
@@ -158,6 +158,101 @@ public struct NIOSSLPKCS12Bundle: Hashable {
...
+            let result = additionalCertificate.withUnsafeMutableX509Pointer { certificate in
+                CNIOBoringSSL_X509_up_ref(certificate)
+                return CNIOBoringSSL_sk_X509_push(certificateChainStack, certificate)
+            }
+            if result == 0 {
+                fatalError("Failed to add certificate to chain")
+            }
+        }
+
+        defer {
```
Lukasa, line 205: Nit: as a matter of coding style it's a bit nicer to move this immediately after the `sk_X509_new`.

Tests/NIOSSLTests/SSLPKCS12BundleTest.swift
```diff
@@ -527,4 +527,28 @@ class SSLPKCS12BundleTest: XCTestCase {
...
+
+        // Create a PKCS#12...
+        let bundle = NIOSSLPKCS12Bundle(
+            certificateChain: certificates,
+            privateKey: privateKey
+        )
+        let pkcs12 = try bundle.serialize(passphrase: "thisisagreatpassword".utf8)
+
+        // And then decode it into a NIOSSLPKCS12Bundle
+        let decoded = try NIOSSLPKCS12Bundle(buffer: pkcs12, passphrase: "thisisagreatpassword".utf8)
```
Lukasa, line 548: We should probably try some non-matching passphrases.

Lukasa (approved):

PR #497, opened 2024-12-09
Enable MemberImportVisibility check on all targets

rnro:
Enable MemberImportVisibility check on all targets. Use a standard string header and footer to bracket the new block for ease of updating in the future with scripts.

### DIFF
--- a/Sources/NIOSSL/IdentityVerification.swift
+++ b/Sources/NIOSSL/IdentityVerification.swift
@@ -12,11 +12,8 @@
 //
 //===----------------------------------------------------------------------===//
 
-import NIOCore
-
-#if os(Linux) || os(FreeBSD) || os(Android)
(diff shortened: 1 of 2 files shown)


Sources/NIOSSL/IdentityVerification.swift
```diff
@@ -14,6 +14,10 @@
 
 import NIOCore
 
+#if os(Linux) || os(FreeBSD) || os(Android)
+import CNIOLinux
+#endif
```
glbrntt, line 19: You should be able to unconditionally import this -- it if-defs itself so that it's empty on other platforms.

rnro: Thanks, I forgot I did that here

glbrntt (approved):

PR #504, opened 2025-01-30
CI use 6.1 nightlies

rnro:
CI use 6.1 nightlies now that Swift development is happening in the 6.1 branch


Lukasa:
Can you add 6.1 thresholds?

rnro:
Sorry, I added that separately since I believe they're already running by default https://github.com/apple/swift-nio-ssl/pull/505

Lukasa (approved):

Lukasa (approved):

PR #521, opened 2025-04-11
Swap some implementation only imports to internal imports

Lukasa:


Sources/NIOSSL/ByteBufferBIO.swift
```diff
@@ -12,9 +12,14 @@
 //
 //===----------------------------------------------------------------------===//
 
-@_implementationOnly import CNIOBoringSSL
 import NIOCore
 
+#if compiler(>=6.1)
```
glbrntt, line 17: We could do internal imports from 6.0, so why 6.1, did something change?

Lukasa: Yeah, we started getting warnings, which was going to inhibit warnings as errors.

glbrntt (approved):

PR #522, opened 2025-04-11
Remove some redundant suppressed conformances

Lukasa:
These suppressed conformances are not necessary, as they are redundant with the suppressed conformance on the parent class.


Lukasa:
API breakage checker is wrong here, we'll have to merge over it.

glbrntt (approved):

PR #524, opened 2025-04-11
Clean up Sendability errors in pipelines and promises

Lukasa:


Lukasa:
As a fun benefit, this PR also reduces allocations during the handshake.

glbrntt (approved):

PR #533, opened 2025-04-14
Add subjectName, issuerName, and access to the peerCert

jtdavey:
Motivation:

While swift-nio-ssl supports mTLS on the server, there's no api available to get the verified X509 certificate after the handshake completes.

Additionally, the existing API for NIOSSLCertificate doesn't have an API to access the issuer and subject names.

Result:

These changes add new APIs. Specifically:
- nioSSL_peerCertificate() to the channel extension. Similar to retrieving the tls version, this gets the verified peer X509 certificate after the handshake completes.
- subjectName extension to NIOSSLCertificate. Grabs the subjectName from the BoringSSL X509 object as an array of SSLCertificateName
- issuerName extension to the NIOSSLCertificate. Grabs the issuerName from the BoringSSL X509 object as an array of SSLCertificateName

### DIFF
--- a/Sources/NIOSSL/NIOSSLHandler.swift
+++ b/Sources/NIOSSL/NIOSSLHandler.swift
@@ -810,6 +810,10 @@ extension NIOSSLHandler {
     public var tlsVersion: TLSVersion? {
         self.connection.getTLSVersionForConnection()
     }
+
+    public var peerCertificate: NIOSSLCertificate? {
+        self.connection.getPeerCertificate()
+    }
 }
 
 extension Channel {
@@ -819,6 +823,13 @@ extension Channel {
             $0.tlsVersion
         }
     }
+
+    public func nioSSL_peerCertificate() -> EventLoopFuture<NIOSSLCertificate?> {
+        self.pipeline.handler(type: NIOSSLHandler.self).map {
+            $0.peerCertificate
+        }
+    }
+
 }
 
 extension ChannelPipeline.SynchronousOperations {
@@ -826,6 +837,11 @@ extension ChannelPipeline.SynchronousOperations {
     public func nioSSL_tlsVersion() throws -> TLSVersion? {
         let handler = try self.handler(type: NIOSSLHandler.self)
         return handler.tlsVersion
+    }
+
+    public func nioSSL_peerCertificate() throws -> NIOSSLCertificate? {
+        let handler = try self.handler(type: NIOSSLHandler.self)
+        return handler.peerCertificate
     }
 }
 
--- a/Sources/NIOSSL/NIOSSLHandler.swift
+++ b/Sources/NIOSSL/NIOSSLHandler.swift
@@ -811,6 +811,9 @@ extension NIOSSLHandler {
         self.connection.getTLSVersionForConnection()
     }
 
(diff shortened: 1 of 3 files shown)


Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -819,6 +823,13 @@ extension Channel {
             $0.tlsVersion
         }
     }
+
+    public func nioSSL_peerCertificate() -> EventLoopFuture<NIOSSLCertificate?> {
```
Lukasa, line 827: Similarly, can we add a doc comment to this?

Sources/NIOSSL/NIOSSLHandler.swift
```diff
@@ -827,6 +838,11 @@ extension ChannelPipeline.SynchronousOperations {
         let handler = try self.handler(type: NIOSSLHandler.self)
         return handler.tlsVersion
     }
+
+    public func nioSSL_peerCertificate() throws -> NIOSSLCertificate? {
```
Lukasa, line 842: Similarly, can we add a doc comment to this?

Sources/NIOSSL/SSLCertificateName.swift
```diff
@@ -0,0 +1,81 @@
+@_implementationOnly import CNIOBoringSSL
```
Lukasa, line 1: This file will need a license header.

Sources/NIOSSL/SSLCertificateName.swift
```diff
@@ -0,0 +1,81 @@
...
+
+public struct SSLCertificateNameType: Equatable, Hashable, Sendable {
```
Lukasa, line 3: Can we add doc comments to all the public API in this file?

Sources/NIOSSL/SSLCertificateName.swift
```diff
@@ -0,0 +1,81 @@
...
+    public var nid: Int32
```
Lukasa, line 4: I suspect this can be `internal`: the values are an implementation detail of BoringSSL and not particularly meaningful for anyone else.

Sources/NIOSSL/SSLCertificateName.swift
```diff
@@ -0,0 +1,81 @@
...
+
+            var encodedName: UnsafeMutablePointer<UInt8>? = nil
+            let stringLength = CNIOBoringSSL_ASN1_STRING_to_UTF8(&encodedName, data)
+
+            guard let namePtr = encodedName else {
+                continue
+            }
+            let arr = UnsafeBufferPointer(start: namePtr, count: Int(stringLength))
+            let nameString = String(decoding: arr, as: UTF8.self)
+            CNIOBoringSSL_OPENSSL_free(namePtr)
```
Lukasa, line 52: Mild nit: it's often nicer to stick this in a `defer` immediately after the `guard`. That way we know we can't possibly forget this in a refactor.

Sources/NIOSSL/SSLCertificateName.swift
Lukasa: Oh sorry, I didn't mean the whole type, just the `nid` property.

Lukasa (approved):
Awesome, this LGTM. Thanks @jtdavey! :sparkles:

PR #540, opened 2025-06-11
Add `toDerBytes` to `SSLPrivateKey`

gjcairo:
This PR adds new API (`SSLPrivateKey/toDerBytes()`) to get the DER bytes for a private key.
This is only supported for native (EVP) keys; if using a custom key, an empty array of bytes will be returned instead.

### DIFF
--- a/Sources/NIOSSL/CustomPrivateKey.swift
+++ b/Sources/NIOSSL/CustomPrivateKey.swift
@@ -81,7 +81,7 @@ public protocol NIOSSLCustomPrivateKey: _NIOPreconcurrencySendable {
 }
 
 extension NIOSSLCustomPrivateKey {
-    public var derBytes: [UInt8] { [] }
+    @inlinable public var derBytes: [UInt8] { [] }
 }
 
 /// This is a type-erased wrapper that can be used to encapsulate a NIOSSLCustomPrivateKey and provide it with


Sources/NIOSSL/SSLPrivateKey.swift
```diff
@@ -404,6 +404,19 @@ extension NIOSSLPrivateKey {
             return customKey.signatureAlgorithms
         }
     }
+
+    /// Extracts the bytes of this private key in DER format.
+    /// - Important: If the key is not an `EVP_PKEY`, an empty array will be returned.
+    /// - Returns: The DER-encoded bytes for this `EVP_PKEY` private key.
+    /// - Throws: If an error occurred while serializing the private key.
+    public func toDERBytes() throws -> [UInt8] {
```
Lukasa, line 412: Can we add tests for this?

Sources/NIOSSL/SSLPrivateKey.swift
```diff
@@ -404,6 +404,19 @@ extension NIOSSLPrivateKey {
...
+        switch self.representation {
+        case .native(let evpKey):
+            return try Self.withUnsafeDERBuffer(of: evpKey) { Array($0) }
+        case .custom(let custom):
+            return []
```
Lukasa, line 417: We should allow custom private keys to handle this case, but just default them to returning the empty array.

Sources/NIOSSL/SSLPrivateKey.swift
gjcairo: Welp, I just realise the latest push had failed - tests exist now :D

Sources/NIOSSL/SSLPrivateKey.swift
gjcairo: Also having second thoughts about the API: do you think `toDERBytes()` is more appropriate than having a `derBytes` computed property?

Sources/NIOSSL/CustomPrivateKey.swift
```diff
@@ -75,12 +80,14 @@ public protocol NIOSSLCustomPrivateKey: _NIOPreconcurrencySendable {
     func decrypt(channel: Channel, data: ByteBuffer) -> EventLoopFuture<ByteBuffer>
 }
 
+extension NIOSSLCustomPrivateKey {
+    var derBytes: [UInt8] { [] }
```
Lukasa, line 84: I'm pretty sure this has to be public, and it escaped only because we have a `@testable` import in the test file.

Sources/NIOSSL/CustomPrivateKey.swift
```diff
@@ -75,12 +80,14 @@ public protocol NIOSSLCustomPrivateKey: _NIOPreconcurrencySendable {
     func decrypt(channel: Channel, data: ByteBuffer) -> EventLoopFuture<ByteBuffer>
 }
 
+extension NIOSSLCustomPrivateKey {
+    public var derBytes: [UInt8] { [] }
```
Lukasa, line 84: This can be `@inlinable` too.

Lukasa (approved):
Very nice, thanks @gjcairo.

PR #544, opened 2025-07-03
Revert "Enable certificate verification if trustRoots are set to non-`nil`"

rnro:
Reverts apple/swift-nio-ssl#543

This may have resulted in a behavior change which would adversely affect downstream, we should back it out whilst we investigate.


Lukasa (approved):
Thanks Rick.

PR #545, opened 2025-07-04
Add mTLSConfiguration option for TLSConfiguration

### LINKED ISSUE #542: Provide a `TLSConfiguration.makeServerWithMTLSConfiguration` or similar to make it less dangerous to configure mTLS with NIO
Configuring mTLS servers with NIOSSL is a little dangerous and weird. Here is what you're meant to do:

```
        var tlsConfiguration = TLSConfiguration.makeServerConfiguration(
            certificateChain: serverChain
        )
        if requireMTLS {
            tlsConfiguration.trustRoots = caChainToCheckClients
            tlsConfiguration.certificateVerification = .noHostnameVerification // THIS IS IMPORTANT(!!)
        }
        return tlsConfiguration
```

The crucial bit here is `tlsConfiguration.certificateVerification = .noHostnameVerification` and this fails _open_. So if you forget that line, then it defaults to `.noVerification` which just doesn't do mTLS at all. And `.fullVerification` is also no good because otherwise it tries to check the client's host name.

So I believe there's only one correct construction and that's the one above. This should be made much easier for the user.

Furthermore, it might make sense to raise from `.noValidation` to `.noHostnameVerification` if the user sets `trustRoots`, why else would they set it?

Lukasa:
Motivation:

When configuring for mTLS, there is really only one right way to do that. This adds that behaviour.

Modifications:

- Add mTLSConfiguration option
- Add tests

Result:

Better behaviour for mTLS.
Resolves #542


glbrntt (approved):
Nice one

PR #554, opened 2025-08-14
Updated test certificates in NIOSSLTestHelpers

aryan-25:
Motivation:

Some tests are failing in `TLSConfigurationTest` due to the certificates in `NIOSSLTestHelpers` having expired.

Modifications:

Updated `NIOSSLTestHelpers` with regenerated certificates.

Result:

Tests no longer fail.


Lukasa (approved):
Nice one, thanks @aryan-25.

PR #557, opened 2025-09-16
Make sslContextCallback overrides replace the existing chain, rather than adding to it

hamzahrmalik:
Currently, the certs returned by the sslContextCallback are added to the context, rather than replacing the certs already in the context.

This means that, if a callback is used to return a new chain to be used, the old and new chains are both presented (but only the new leaf). This should not cause any problem, but it is inefficient

Change: Clear the additional certs before adding the new ones


Lukasa (approved):
Very nice, thanks @hamzahrmalik!

PR #560, opened 2025-09-30
Use `Context.environment` in `Package.swift` instead of importing `Foundation`

gwynne:
#### Motivation:

Importing `Foundation` in `Package.swift` is undesirable; nowhere else
in the actual `NIOSSL` package is `Foundation` used, it appears only in
tests and examples. This removes the usage from the manifest altogether.
There is no compatibility concern; `Context.environment` has been
available since SwiftPM 5.6.

#### Modifications:

Replaced `ProcessInfo.processInfo.environment` with
`Context.environment`.

#### Result:

There is no functional impact. `Package.swift` no longer imports
`Foundation`.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -15,10 +15,6 @@
 
 import PackageDescription
 
-// Used only for environment variables, does not make its way
-// into the product code.
-import class Foundation.ProcessInfo
-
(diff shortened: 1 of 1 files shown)


glbrntt (approved):
Awesome, thanks Gwynne!

PR #567, opened 2025-11-04
Add support for Windows

Joannis:
Fix compilation of NIOSSL on Windows
Note: This code adds some things we might not want to keep around.
- OPENSSL_NO_ASM flag which I neededto get it to work
- Windows directory listing using FindFirstFileA (not properly tested)
- mlock -> VirtualLock on Windows (doesn't yet check if locking was successful)


Joannis:
Still needs to set up Windows CI, but if possible I'd like to ensure the tests still succeed on existing platforms

Sources/CNIOBoringSSL/crypto/thread_win.cc
```diff
@@ -115,18 +115,18 @@ static void NTAPI thread_local_destructor(PVOID module, DWORD reason,
 //
 // Force a reference to _tls_used to make the linker create the TLS directory
 // if it's not already there. (E.g. if __declspec(thread) is not used). Force
-// a reference to p_thread_callback_boringssl to prevent whole program
+// a reference to CNIOBoringSSL_p_thread_callback_boringssl to prevent whole program
```
Lukasa: We aren't taking any changes to BoringSSL itself. They shouldn't be necessary.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -919,16 +924,35 @@ internal class DirectoryContents: Sequence, IteratorProtocol {
...
+            let name = withUnsafePointer(to: &fileData.cFileName) { ptr in
+                // Pointers to homogeneous tuples in Swift are always bound to both the tuple type and the element type,
+                // so the assumption below is safe.
+                let elementPointer = UnsafeRawPointer(ptr).assumingMemoryBound(to: CChar.self)
+                return String(cString: elementPointer)
+            }
+            FindNextFileA(dir, &fileData)
+            return self.path + name
+        }
+        #else
```
Lukasa, line 1000: Why does this need an `#if` around the whole block? It'd be nicer to factor the calls out to helpers that can have a small `#if`.

Lukasa (changes requested):

odrobnik:
@Lukasa Maybe give it another look now? We’ve worked in your review comments.

Budoman:
Independent Windows CI datapoints for this PR, from a clean windows-2022 GitHub runner (Swift 6.3.3), two runs.

**Run 1, this branch as-is:** builds with zero errors (main scores 1,633 errors on the same runner, so this PR takes Windows from unbuildable to clean). The test suite is green across the unit suites (ByteBufferBIO, CertificateVerification, ClientSNI, CustomPrivateKey, IdentityVerification, ALPN, and the rest). Five failures, all in NIOSSLIntegrationTest and all in the close/shutdown family:
`testCloseModeOutputServerAndClient`, `testCloseModeOutputTriggersFlush`, `testMultipleCloseOutput`, `testReceivingGibberishAfterAttemptingToClose`, `testSubsequentWritesFailAfterCloseModeOutput` (errors: "Already closed", "uncleanShutdown", one XCTAssertNotNil).

**Run 2, same branch with swift-nio overridden to apple/swift-nio#3704's head** (the Windows data-path backpressure fix): identical result, same five failures. So these are not downstream of the known WSAEWOULDBLOCK data-path bug; they look like Windows half-close/shutdown semantics in their own right.

Happy to share the workflow or raw logs, and to re-run against any revision if that helps get this landed. Even as-is, this PR takes swift-nio-ssl from 1,633 errors to a clean build with a green core suite on Windows.

Budoman:
Follow-up on the five failures: root-caused, and none of them is this PR's fault.

Four are a swift-nio defect. `SelectorWSAPoll` maps `POLLHUP` to `.reset`, but WSAPoll has no `POLLRDHUP`, so an orderly peer FIN and an abortive RST arrive identically and both take the terminal reset path: the selector can never emit `.readEOF`, which makes `allowRemoteHalfClosure` unimplementable on Windows, and a live half-closed channel gets force-closed. Instrumented on windows-2022: 8 POLLHUP deliveries, 4 of them force-closing an open, active channel. One-line fix, mapping `POLLHUP` to `.readEOF` so `recv` becomes the authority (0 = orderly FIN honouring half-closure; `WSAECONNRESET`/`WSAECONNABORTED` = genuine abort): `Budoman/swift-nio@windows-half-closure` (7fc88bf, stacked on apple/swift-nio#3704). Validated: the four close-mode tests pass, the full swift-nio-ssl suite is green, and swift-nio's own `testHalfCloseOwnOutput` flips from failing to passing, with zero new failures and macOS unaffected.

The fifth (`testReceivingGibberishAfterAttemptingToClose`) is a test-only issue: its regex requires a forward slash before BoringSSL's `__FILE__` path, which is backslash-separated on Windows. Test-only fix, one hunk: `Budoman/swift-nio-ssl@windows-close-mode` (803d461, based on `jo/windows-support`), free to lift straight into this branch.

With both applied, this PR's suite is fully green on windows-2022 / Swift 6.3.3. One structural note: Windows half-closure currently has no upstream CI guard because `StreamChannelTest` is skipped wholesale (`NIOPosix` has no `PipeChannel` on Windows), which is why the selector defect could ship undetected. We are opening the swift-nio fix as a PR so it can be reviewed properly.

Budoman:
One more independent Windows datapoint from assembling a real server on top of this branch: a duplicate-symbol collision between CNIOBoringSSL and swift-crypto's CCryptoBoringSSL that will hit anyone linking both into one Windows binary (which any TLS server using swift-crypto for application crypto will do).

**Symptom** (lld, windows-2022, Swift 6.3.3):

```
lld-link: error: duplicate symbol: p_thread_callback_boringssl
>>> defined at .build\checkouts\swift-crypto\Sources\CCryptoBoringSSL\crypto\thread_win.cc:157
>>> defined at .build\checkouts\swift-nio-ssl\Sources\CNIOBoringSSL\crypto\thread_win.cc:159
```

**Root cause:** both projects vendor BoringSSL behind a symbol prefix, but `p_thread_callback_boringssl` in `thread_win.cc` escapes both prefix schemes. It is declared `extern "C"` with deliberately external linkage (the `.CRT$XLC` TLS-callback registration plus a `/INCLUDE:` linker pragma depend on that), and the prefix headers do not cover it; the comment in the file even anticipates this: "in the prefixed build, |p_thread_callback_boringssl| may be a macro", but no such macro is defined in either project. The two copies therefore collide at link time. macOS and Linux never see this because the symbol only exists under `_WIN32`.

**Fix that worked for us:** rename the symbol in one copy. We renamed CNIOBoringSSL's to `p_thread_callback_nio_boringssl` (all nine occurrences in `thread_win.cc`; the `/INCLUDE:` pragmas stringify the same identifier, so the rename carries through and the CRT registration stays intact): [Budoman/swift-nio-ssl@3baecb7](https://github.com/Budoman/swift-nio-ssl/commit/3baecb7707266162de3c9b73a981e558edf7e488). With that one file changed, our full multi-product build (server with NIOSSL TLS + swift-crypto, several CLI tools) completes cleanly on windows-2022.

A more principled home for the fix might be BoringSSL's prefixing mechanism itself (adding the symbol to the prefix list), which would fix it for every downstream vendored copy at once. Happy to turn either form into a PR if useful.

Budoman:
Windows datapoint on the fresh merge (head `22c1c18`), same clean windows-2022 runner + Swift 6.3.3 as the earlier runs.

**The library is still clean: `swift build` completes with zero errors.** The merge did not disturb the Windows build of NIOSSL itself.

**The test bundle no longer compiles on Windows**, from code the merge brought in from `main`:

```
Tests/NIOSSLTests/NIOSSLIntegrationTest.swift:1552, 1624
error: cannot find 'SOL_SOCKET' in scope
error: cannot find 'SO_REUSEADDR' in scope
error: type 'ChannelOptions' has no member 'socket'
```

Two sites, both `ServerBootstrap` setup using the raw POSIX constants. The portable spelling fixes both lines with no platform condition:

```swift
.serverChannelOption(ChannelOptions.socketOption(.so_reuseaddr), value: 1)
```

Worth knowing while iterating: a plain `swift test` on Windows reports only a bare `error: fatalError` for this, with no diagnostics; `swift build --build-tests -v` is what surfaces the real errors.

**One heads-up for the step after that.** The merge also brought `import Crypto` (plus `SwiftASN1`/`X509`) into the test helpers, so the test binary now links both CCryptoBoringSSL and CNIOBoringSSL. On Windows that hits the `p_thread_callback_boringssl` duplicate-symbol collision described in my previous comment, so the suite will stop at the linker even once the two sites above are fixed. The one-file rename ([Budoman/swift-nio-ssl@3baecb7](https://github.com/Budoman/swift-nio-ssl/commit/3baecb7707266162de3c9b73a981e558edf7e488)) resolves it; a prefix-list fix in BoringSSL upstream would be the more general home.

@Joannis happy to open small PRs against `jo/windows-support` for any or all of: the two socket-option lines, the BoringSSL symbol rename, and the earlier separator-agnostic file-path test fix. Say the word and they are yours within the day.

leovalen:
It would be really nice to see this merged. Seems like a low hanging fruit :)

PR #580, opened 2026-05-02
Reject partial wildcard SAN matches that produce IDNA A-label first label

kuranikaran:
### Motivation

`AnalysedCertificateHostname.init` already refuses to construct a wildcard matcher when the certificate's first label begins with the IDNA A-label prefix `xn--` (line 287). The symmetric case is not enforced: a wildcard SAN whose own first label is plain ASCII can still expand into a target hostname whose first label begins with `xn--`.

A SAN of the form `*-5wao1o.example.com` will currently match a target hostname `xn--rksmrgs-5wao1o.example.com` (the IDNA A-label form of `räksmörgås.example.com`). Both OpenSSL's `equal_wildcard` and CPython's `ssl._dnsname_match` reject that case explicitly; this library accepts it.

Verified by copying `AnalysedCertificateHostname` verbatim from `main` into a standalone executable and running:

```
[OK     ] Sanity: wildcard matches normal subdomain
           cert SAN: *.example.com
           target:   foo.example.com
           expected: true   got: true

[OK     ] Sanity: wildcard does NOT match different parent
           cert SAN: *.example.com
           target:   foo.evil.com
           expected: false   got: false

[OK     ] Sanity: cert that IS an A-label wildcard rejected at parse
           cert SAN: xn--*.example.com
           target:   xn--anything.example.com
           expected: false   got: false

[BROKEN ] VULN: partial wildcard expansion produces A-label target
           cert SAN: *-5wao1o.example.com
           target:   xn--rksmrgs-5wao1o.example.com
           expected: false   got: true
```

After the patch in this PR the BROKEN case becomes OK.

### Modifications

In `validMatchForName`, after splitting the target around its first period, return `false` if the target's first label begins with `xn--`. Mirrors the existing cert-side check at line 287 and the `caseInsensitiveElementsEqual(asciiIDNAIdentifier)` idiom already in use in this file.

### Result

Wildcard SANs no longer expand into IDNA A-label targets, matching the behaviour of OpenSSL and CPython. The existing `testDoesNotMatchIDNALabelWithWildcard` continues to cover the cert-side variant; the target-side variant is the new case this patch addresses.

### DIFF
--- a/Sources/NIOSSL/IdentityVerification.swift
+++ b/Sources/NIOSSL/IdentityVerification.swift
@@ -315,6 +315,10 @@ private struct AnalysedCertificateHostname {
             let (wildcardLabel, remainingComponents) = baseName.splitAroundIndex(firstPeriodIndex)
             let (targetFirstLabel, targetRemainingComponents) = target.splitAroundIndex(firstPeriodIndexForName)
 
+            guard !targetFirstLabel.prefix(4).caseInsensitiveElementsEqual(asciiIDNAIdentifier) else {
+                return false
+            }
+
             guard remainingComponents.caseInsensitiveElementsEqual(targetRemainingComponents) else {
                 // Wildcard is irrelevant, the remaining components don't match.
                 return false
--- a/Tests/NIOSSLTests/IdentityVerificationTest.swift
+++ b/Tests/NIOSSLTests/IdentityVerificationTest.swift
@@ -293,6 +293,41 @@ class IdentityVerificationTest: XCTestCase {
         XCTAssertTrue(matched)
     }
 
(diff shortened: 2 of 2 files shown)


Lukasa (changes requested):
Thanks for this! This is a good fix. Can you please add the necessary regression tests for this?

kuranikaran:
Added the regression test as a second commit. Cert SAN is the partial wildcard, target hostname is the IDNA A-label form. Asserts XCTAssertFalse on the matcher result. Fails on main, passes after the patch.

Lukasa (approved):
Nice change, thanks!

PR #582, opened 2026-05-14
Enable x25519_MLKEM768 by default

rafaelcepeda:
Default `TLSConfiguration.curves` is now [`.x25519_MLKEM768`, `.x25519`, `.secp384r1`], enabling post-quantum hybrid key exchange out of the box. Users can still set curves to nil to fall back to BoringSSL's built-in curve list, or override with any custom set.

Updated `quantum-secure-tls.md` to reflect that PQ is enabled by default and document how to customise or disable it.

### DIFF
--- a/Sources/NIOSSL/SSLContext.swift
+++ b/Sources/NIOSSL/SSLContext.swift
@@ -12,10 +12,9 @@
 //
 //===----------------------------------------------------------------------===//
 
-import NIOCore
-
 @_implementationOnly import CNIOBoringSSL
 @_implementationOnly import CNIOBoringSSLShims
+import NIOCore
 
 #if canImport(Darwin)
 import Darwin.C
--- a/Sources/NIOSSL/SSLContext.swift
+++ b/Sources/NIOSSL/SSLContext.swift
@@ -12,9 +12,10 @@
 //
 //===----------------------------------------------------------------------===//
 
+import NIOCore
+
 @_implementationOnly import CNIOBoringSSL
 @_implementationOnly import CNIOBoringSSLShims
-import NIOCore
 
 #if canImport(Darwin)
 import Darwin.C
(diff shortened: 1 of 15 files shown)


Sources/NIOSSL/SSLContext.swift
```diff
@@ -339,17 +347,19 @@ public final class NIOSSLContext {
...
-                    CNIOBoringSSL_SSL_CTX_set1_group_ids(context, algo.baseAddress, algo.count)
-                }
-            if returnCode != 1 {
-                let errorStack = BoringSSLError.buildErrorStack()
-                throw BoringSSLError.unknownError(errorStack)
+        // When the user has not expressed a preference, we supplement BoringSSL's
+        // default groups with x25519_MLKEM768 for post-quantum key exchange.
+        let groupIDs =
+            configuration.curves.map { $0.map { $0.rawValue } }
+            ?? [NIOTLSCurve.x25519_MLKEM768.rawValue] + boringSSLDefaultGroups
```
Lukasa: I think this statement may have gotten past my ability to cleanly follow 😵 . Might be worth factoring this out to a helper where we can be a bit more explicit.

rafaelcepeda: I've refactored this logic into a helper function.

Sources/NIOSSL/SSLContext.swift
```diff
@@ -33,6 +33,26 @@ import Android
...
+/// Returns the group IDs to configure on a context for the given curves preference.
+///
+/// When the user has expressed a preference (a non-nil array), it is honoured exactly.
+/// Otherwise, we supplement BoringSSL's default groups with x25519_MLKEM768 to enable
+/// post-quantum hybrid key exchange by default.
+private func resolveGroupIDs(for curves: [NIOTLSCurve]?) -> [UInt16] {
+    if let curves = curves {
+        return curves.map { $0.rawValue }
+    }
+    return [NIOTLSCurve.x25519_MLKEM768.rawValue] + boringSSLDefaultGroups
```
Lukasa: We may as well skip the `+` operation here and just store our preferences, and leave a comment noting that this is the BoringSSL list plus x25519_MLKEM768

Lukasa (approved):

PR #590, opened 2026-07-29
Don't crash when a peer completes an optionalVerification handshake without a certificate

rafaelcepeda:
## Motivation

  When a `NIOSSLHandler` is created with an `additionalPeerCertificateVerificationCallback`
  (via `_makeSSLServerHandler` / `_makeSSLClientHandler`) and the `TLSConfiguration` uses
  `.optionalVerification`, the peer is allowed to complete the handshake *without* presenting
  a certificate.

  In that situation `doHandshakeStep` reached the additional-verification branch, called
  `connection.getPeerCertificate()`, got `nil`, and hit a `preconditionFailure`, aborting the
  entire process.

  The constructor precondition did not catch this because `.optionalVerification` is
  `.none(validatePresentedCertificates: true)`, which is distinct from `.none` and therefore
  passes the `!= .none` check.

  ## Modifications

  - In `NIOSSLHandler.doHandshakeStep`, only invoke the
    `additionalPeerCertificateVerificationCallback` when the peer actually presented a
    certificate. The callback binding and the `getPeerCertificate()` binding are now combined
    into a single `if let ..., let ...`; if no certificate was presented the handshake
    completes normally without invoking the callback. This preserves the previous behaviour
    of only calling `getPeerCertificate()` when a callback is set (short-circuit ordering).
  - Removed the now-unreachable `preconditionFailure`.
  - Added an integration test that stands up a server with `.optionalVerification` and an
    `additionalPeerCertificateVerificationCallback`, connects a client that presents no
    certificate, and asserts the connection is handled without crashing the process.

  ## Result

  A peer that completes an `.optionalVerification` handshake without presenting a certificate
  no longer crashes the process. The additional verification callback still runs, unchanged,
  whenever a certificate *is* presented, so behaviour for the certificate-present path is
  unaffected. `.optionalVerification` + `additionalPeerCertificateVerificationCallback` is now
  a genuinely supported combination, matching what the constructor precondition already
  advertised.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .DS_Store
 /.build
+*/.build
 /Packages
 /*.xcodeproj
 Package.pins
@@ -9,3 +10,4 @@ Package.resolved
 DerivedData
 /.idea
 .swiftpm
+.vscode
--- a/Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
+++ b/Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
@@ -1521,7 +1521,7 @@ class NIOSSLIntegrationTest: XCTestCase {
         XCTAssertTrue(actualErrors.first is CustomUserError)
     }
 
-    // rdar://177743664 — A server built with `.optionalVerification` and an additional
+    // A server built with `.optionalVerification` and an additional
     // peer certificate verification callback must not crash when a client completes the
     // handshake without presenting a certificate. With the bug present, the server hits a
     // `preconditionFailure` in `doHandshakeStep` (getPeerCertificate() returns nil) and the
--- a/Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
+++ b/Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
@@ -1562,10 +1562,11 @@ class NIOSSLIntegrationTest: XCTestCase {
             XCTAssertNoThrow(try serverChannel.close().wait())
         }
 
+        let eventHandler = EventRecorderHandler<TLSUserEvent>()
         let clientChannel = try clientTLSChannel(
             context: clientCtx,
             preHandlers: [],
-            postHandlers: [],
+            postHandlers: [eventHandler],
             group: group,
             connectingTo: serverChannel.localAddress!,
             serverHostname: "localhost"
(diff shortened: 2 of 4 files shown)


Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
```diff
@@ -1521,6 +1521,69 @@ class NIOSSLIntegrationTest: XCTestCase {
         XCTAssertTrue(actualErrors.first is CustomUserError)
     }
 
+    // A server built with `.optionalVerification` and an additional
+    // peer certificate verification callback must not crash when a client completes the
+    // handshake without presenting a certificate. With the bug present, the server hits a
```
aryan-25, line 1526: I think the part explaining the current bug is irrelevant and will be confusing in the future. The same applies for the comment above `var optionalBuffer`.

Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
```diff
@@ -1521,6 +1522,141 @@ class NIOSSLIntegrationTest: XCTestCase {
...
+        )
+        defer {
+            XCTAssertNoThrow(try? clientChannel.close().wait())
+        }
+
+        // Drive the handshake to completion. With the bug present the server process aborts
+        // here; post-fix this must complete or fail gracefully without crashing the process.
+        var originalBuffer = clientChannel.allocator.buffer(capacity: 5)
+        originalBuffer.writeString("Hello")
+        _ = try? clientChannel.writeAndFlush(originalBuffer).wait()
```
aryan-25: We can more strongly assert the handshake was successful by doing this:

```suggestion
        let eventHandler = EventRecorderHandler<TLSUserEvent>()
        let clientChannel = try clientTLSChannel(
            context: clientCtx,
            preHandlers: [],
            postHandlers: [eventHandler],
            group: group,
            connectingTo: serverChannel.localAddress!,
            serverHostname: "localhost"
        )
        defer {
            XCTAssertNoThrow(try? clientChannel.close().wait())
        }

        var originalBuffer = clientChannel.allocator.buffer(capacity: 5)
        originalBuffer.writeString("Hello")
        let writeFuture = clientChannel.writeAndFlush(originalBuffer)
        writeFuture.whenComplete { _ in
            XCTAssertEqual(
                eventHandler.events[..<3],
                [.Registered, .Active, .UserEvent(.handshakeCompleted(negotiatedProtocol: nil))]
            )
        }
        try writeFuture.wait()
```

Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
```diff
@@ -1521,6 +1522,141 @@ class NIOSSLIntegrationTest: XCTestCase {
...
+
+        // The client can flush 0.5-RTT data before the server's rejection alert arrives, so the
+        // write is best-effort. The reliable signal is that the server rejects the certificate-less
+        // client and tears the connection down.
+        var originalBuffer = clientChannel.allocator.buffer(capacity: 5)
+        originalBuffer.writeString("Hello")
+        _ = try? clientChannel.writeAndFlush(originalBuffer).wait()
+
+        // Wait for the connection to be torn down as a result of the failed handshake.
+        try clientChannel.closeFuture.wait()
```
aryan-25: Waiting for just the close promise is sufficient here.

```suggestion
        // Wait for the connection to be torn down as a result of the failed handshake.
        try clientChannel.closeFuture.wait()
```

.gitignore
aryan-25, line 1: Seems like this was an accidental change.

Tests/NIOSSLTests/NIOSSLIntegrationTest.swift
```diff
@@ -1521,6 +1522,141 @@ class NIOSSLIntegrationTest: XCTestCase {
         XCTAssertTrue(actualErrors.first is CustomUserError)
     }
 
+    //
```
aryan-25: Seems like this was an accidental change.
```suggestion
```

aryan-25 (approved):
Thank you!

PR #591, opened 2026-08-05
Expose SSL_set_enforce_rsa_key_usage via TLSConfiguration

### LINKED ISSUE #589: Expose SSL_set_enforce_rsa_key_usage via TLSConfiguration (allow relaxing RSA leaf keyUsage check on the client)
### Motivation

When acting as a TLS client against a server whose RSA leaf certificate has a `keyUsage` extension without the bit required by the negotiated cipher, BoringSSL aborts the handshake with
`OPENSSL_internal:KEY_USAGE_BIT_INCORRECT`. On the TLS 1.2 ECDHE_RSA path this is `digitalSignature`; on TLS 1.3 the `digitalSignature` requirement is unconditional.

This bites interop with widely-deployed servers that ship RSA certificates carrying only `keyEncipherment` (+ `dataEncipherment`) - notably Windows RDP's auto-generated self-signed certificates. Crucially,
`certificateVerification = .none` does **not** relax this check: keyUsage enforcement lives on the cipher/cert-consistency path, not the trust-verification path (the header even notes the flag "will not be read until after the certificate verification callback"). So there is currently no way, from NIOSSL's public API, to complete such a handshake.

BoringSSL already provides the exact escape hatch: `SSL_set_enforce_rsa_key_usage(ssl, 0)`, which relaxes the check for RSA leaves when the client negotiates TLS 1.2 or below (it's a no-op / always-enforced
otherwise). It's a per-`SSL` setting with no `SSL_CTX` equivalent, and NIOSSL exposes no hook that surfaces the raw `SSL *` (`NIOSSLContextCallback` only provides SNI values and a cert/key override), so
applications can't reach it today.

### Proposal

Add an opt-in `TLSConfiguration` property, defaulting to the current strict behavior:

```swift
/// Whether the keyUsage extension of RSA leaf certificates is enforced for
/// consistency with the negotiated TLS usage, when acting as a client at
/// TLS 1.2 or below. Mirrors BoringSSL's `SSL_set_enforce_rsa_key_usage`.
/// Defaults to `true` (enforced).
public var enforceRSAKeyUsage: Bool = true
```

Applied per connection in `SSLConnection.init` on the client path:

```swift
if !configuration.enforceRSAKeyUsage {
    CNIOBoringSSL_SSL_set_enforce_rsa_key_usage(self.ssl, 0)
}
```

This preserves normal forward-secret cipher negotiation for the connection (unlike the alternative of pinning static-RSA cipher suites) and only relaxes the leaf keyUsage check when explicitly requested.

### Alternatives considered

- Pinning `cipherSuites` to `kRSA` so the check requires `keyEncipherment` instead - degrades the live connection to a non-forward-secret cipher and fails outright against servers that have disabled static-RSA
suites.
- Using Network.framework via swift-nio-transport-services - needless internal rewrite, would be nice to avoid if we can help it.

### Environment

- swift-nio-ssl 2.37.1 (BoringSSL vendored therein)

Happy to open a PR if the direction is acceptable.

---

tl;dr: I have an RDP proto library that uses swift-nio-ssl that has compatibility problems with Windows. Please expose this BoringSSL tunable so downstream consumers dont have to use ugly workarounds. Please and thank you!

adityasingh2400:
BoringSSL rejects a handshake with `KEY_USAGE_BIT_INCORRECT` when a client negotiating TLS 1.2 or below gets an RSA leaf whose `keyUsage` extension is inconsistent with the negotiated cipher, and that check is not part of trust evaluation, so `certificateVerification = .none` does not relax it. `SSL_set_enforce_rsa_key_usage` is the documented escape hatch, but it is a per-`SSL` setting with no `SSL_CTX` equivalent and NIOSSL never surfaced the raw `SSL *`, so there was no way to reach it from the public API.

This adds `TLSConfiguration.enforceRSAKeyUsage`, defaulting to `true` so existing behaviour is unchanged, and applies it in `SSLConnection.init` where the per-connection `SSL` object is available. The docs spell out that it should only be turned off for a specific interoperability reason, since it means accepting a peer that is using its key contrary to its own certificate. The property is also folded into `bestEffortEquals` and `bestEffortHash`.

Verified with four new tests in `TLSConfigurationTest`, backed by a `keyUsage` parameter added to the `generateSelfSignedCert` helper: an RSA leaf carrying only `keyEncipherment` fails the handshake with `KEY_USAGE_BIT_INCORRECT` by default, succeeds once the client sets `enforceRSAKeyUsage = false`, still fails when only the server opts out since the setting is client-only in BoringSSL, and configs differing solely in this flag compare unequal. The success test fails with `KEY_USAGE_BIT_INCORRECT` when the `SSLConnection` change is reverted.

Fixes #589

### DIFF
--- a/Sources/NIOSSL/SSLConnection.swift
+++ b/Sources/NIOSSL/SSLConnection.swift
@@ -77,6 +77,13 @@ internal final class SSLConnection {
         CNIOBoringSSL_SSL_set_ex_data(self.ssl, sslConnectionExDataIndex, pointerToSelf)
 
         self.setRenegotiationSupport(self.parentContext.configuration.renegotiationSupport)
+
+        // This is a per-connection setting with no SSL_CTX equivalent, so it has to be applied here.
+        // BoringSSL only consults it when we are a client negotiating TLS 1.2 or below, and ignores
+        // it otherwise.
+        if !self.parentContext.configuration.enforceRSAKeyUsage {
+            CNIOBoringSSL_SSL_set_enforce_rsa_key_usage(self.ssl, 0)
+        }
     }
 
     deinit {
--- a/Sources/NIOSSL/TLSConfiguration.swift
+++ b/Sources/NIOSSL/TLSConfiguration.swift
@@ -455,6 +455,20 @@ public struct TLSConfiguration {
     /// This instructs the client which identities can be used by evaluating what CA the identity certificate was issued from.
     public var sendCANameList: Bool
 
+    /// Whether the `keyUsage` extension of an RSA leaf certificate is required to be consistent with the
+    /// negotiated TLS usage. Defaults to `true`.
+    ///
+    /// This only has an effect when acting as a client negotiating TLS 1.2 or below, where the check is
+    /// otherwise performed against the cipher suite in use. In all other cases the check is always enforced
+    /// and this property is ignored. Note that setting ``certificateVerification`` to
+    /// ``CertificateVerification/none`` does not relax this check, as it is not part of trust evaluation.
+    ///
+    /// - Warning: Only set this to `false` when a specific interoperability problem requires it, for example
+    ///     talking to a server whose RSA certificate carries only `keyEncipherment` while the handshake
+    ///     negotiates a cipher suite that needs `digitalSignature`. Disabling the check means the peer is
+    ///     using its key in a way its own certificate says it should not be used for.
+    public var enforceRSAKeyUsage: Bool = true
+
     private init(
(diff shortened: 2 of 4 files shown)


davidben:
Note that BoringSSL has since removed this API, after a very lengthy deprecation process.
https://boringssl.googlesource.com/boringssl.git/+/39dc2564ec06a323d1905b689b08452152a41b70

I'm surprised to hear that Windows RDP certificates get this wrong. Did you file a bug with Microsoft?

PR #594, opened 2026-08-11
NIOSSLContext: fall back to BoringSSL default verify paths on Linux

hpr:
On Linux/FreeBSD, `platformDefaultConfiguration` locates the system trust store via a hardcoded list of FHS paths (`/etc/ssl/certs/...`, `/etc/pki/...`). On systems that do not follow this layout (for example minimal container images or installations using non-standard prefixes), neither path exists, `SSL_CTX_load_verify_locations(NULL, NULL)` fails, and `BoringSSLError.unknownError([])` is thrown — with no way for users to point NIOSSL at their trust store.

This change falls back to `SSL_CTX_set_default_verify_paths`, which uses BoringSSL's compiled-in `OPENSSLDIR` defaults and, importantly, the standard `SSL_CERT_FILE`/`SSL_CERT_DIR` environment variables — the same mechanism OpenSSL users expect.

Systems with the standard FHS layout are unaffected: the heuristic load succeeds first and the fallback never runs.

### DIFF
--- a/Tests/NIOSSLTests/SSLContextTests.swift
+++ b/Tests/NIOSSLTests/SSLContextTests.swift
@@ -12,8 +12,6 @@
 //
 //===----------------------------------------------------------------------===//
 
-import CNIOBoringSSL
-import Foundation
 import NIOCore
 import NIOEmbedded
 import NIOPosix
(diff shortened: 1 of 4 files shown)


Lukasa (commented):
Thanks for this: can you accompany this change with a test?

hpr:
> Thanks for this: can you accompany this change with a test?

Made it testable and added test in https://github.com/apple/swift-nio-ssl/commit/7183a17a29fcc4811a9ad5d8612dcbe747692ae3, thanks

Tests/NIOSSLTests/SSLContextTests.swift
```diff
@@ -206,4 +208,48 @@ final class SSLContextTest: XCTestCase {
...
+        defer {
+            unsetenv("SSL_CERT_FILE")
+            unsetenv("SSL_CERT_DIR")
+        }
+
+        let sslContext = CNIOBoringSSL_SSL_CTX_new(CNIOBoringSSL_TLS_client_method())!
+        defer { CNIOBoringSSL_SSL_CTX_free(sslContext) }
+
+        XCTAssertNoThrow(
+            try NIOSSLContext.platformDefaultConfiguration(
```
Lukasa: I think rather than punching a hole in the API so we can test just this bit in isolation, it might be nicer to confirm that the API itself works as designed. That is, keep the setup the same and then try an in-memory TLS handshake. It should complete successfully.

hpr: @Lukasa Done and passing on https://github.com/hpr/swift-nio-ssl/actions/runs/32087988802, thanks

PR #598, opened 2026-08-15
Avoid unconditional plaintext read buffer reallocations

### LINKED ISSUE #227: Unconditional reallocation on the read path
On the read path in `SSLConnection.readDataFromNetwork` we try to avoid the number of reallocations we perform by using `ByteBuffer.writeWithUnsafeMutableBytes(minimumWritableBytes:)`. We pass `SSL_MAX_RECORD_SIZE` (16kB) to the minimum size, ensuring we will always be able to write a complete record.

This is reasonable in theory, but it behaves awkwardly when we combine it with the fact that `NIOSSLHandler.plaintextReadBuffer` is always allocated with an initial size of `SSL_MAX_RECORD_SIZE`. The reason this is awkward is that we call `SSLConnection.readDataFromNetwork` in a loop until we get `.incomplete`. This means that the first time we read any record, the next looped call will ask for 16kB of free space. We won't have it (we just wrote into the buffer!) and so we'll realloc.

In practice, this forces an unconditional realloc at some stage. Not great!

There are two options to resolve this. The first is to change the capacity reservation. We could try to use `SSL_pending` for this. However, we need to do some investigation before we do that, because `SSL_pending` doesn't read from the transport, and so it may not always be able to tell us how big this needs to be.

The other option, if the above doesn't work, is simply to immediately allocate the buffer at `2 * SSL_MAX_RECORD_SIZE`. In practice this will avoid this unconditional resize, and in most allocators this will also be proxied directly to `mmap`. That means that if we never end up using the trailing 16kB, nothing bad will happen, as the page will never be faulted in.

stephenlclarke:
### Motivation

Fixes #227.

`NIOSSLHandler` initially reserves one maximum TLS record, while the read loop immediately asks for another record's worth of writable space after consuming the first. Reserve two records up front so the normal loop does not force an unconditional buffer growth.

The test-only capacity accessor follows the existing `ByteBufferBIO` test seam pattern and verifies the reservation at handler installation.

### Testing

- Negative control on current `main`: expected 32768 bytes, received 16384
- Focused regression: passed
- Full macOS suite: 325 tests passed
- Production build with the Swift 6.3 CI warning, dependency-import, and explicit-Sendable flags: passed
- Release benchmark package build: passed
- Renegotiation integration test: passed
- `git diff --check`: passed
- Commit signature: verified

### Maintainer note

This is a patch-level performance fix. The fork author cannot apply the required `🔨 semver/patch` label to the upstream PR, so a maintainer will need to add it.

### DIFF
--- a/Sources/NIOSSL/NIOSSLHandler.swift
+++ b/Sources/NIOSSL/NIOSSLHandler.swift
@@ -68,6 +68,10 @@ public class NIOSSLHandler: ChannelInboundHandler, ChannelOutboundHandler, Remov
         self.storedContext?.channel
     }
 
+    internal var _testOnly_plaintextReadBufferCapacity: Int? {
+        self.plaintextReadBuffer?.capacity
+    }
+
     internal init(
         connection: SSLConnection,
         shutdownTimeout: TimeAmount,
@@ -95,7 +99,7 @@ public class NIOSSLHandler: ChannelInboundHandler, ChannelOutboundHandler, Remov
         self.connection.parentHandler = self
         self.connection.eventLoop = context.eventLoop
 
-        self.plaintextReadBuffer = context.channel.allocator.buffer(capacity: SSL_MAX_RECORD_SIZE)
+        self.plaintextReadBuffer = context.channel.allocator.buffer(capacity: 2 * SSL_MAX_RECORD_SIZE)
         // If this channel is already active, immediately begin handshaking.
         if context.channel.isActive {
             doHandshakeStep(context: context)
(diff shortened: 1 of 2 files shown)


stephenlclarke:
Refresh 2026-08-26: rebased the single signed fix onto current main (8d8f7844). New head: 4e51b87e. The exact NIOSSLIntegrationTest.testPlaintextReadBufferCanHoldTwoRecords regression passed, and git diff --check passed. Build output contained only existing upstream Swift compiler warnings.

stephenlclarke:
Refresh 2026-09-05: rebased the single signed fix onto current Apple `main` (`03827c1a9fdb`), including upstream TLS handshake and configuration fixes. The new head is `884b18fc96d8feae3ca63a7e662b8ea0eff31235`; `swift test --jobs 4 --filter NIOSSLIntegrationTest.testPlaintextReadBufferCanHoldTwoRecords` passed 1/1 and `git diff --check` passed. The PR is mergeable; fork-originated CI still requires Apple approval.

Lukasa (approved):
Very nice, thank you!

PR #600, opened 2026-08-27
Fix comparing TLSConfiguration objects with some closures

Lukasa:
Motivation:

TLSConfiguration aims to produce an equality comparison even though it contains closures. Unfortunately, the way it does this is subject to some bugs. Specifically, we pass the closure directly to the generic method `withUnsafeByes` which can in some cases force the compiler to emit a reabstraction thunk. This happens at the point of comparison, and so we can end up at a point where a TLSConfiguration object compares unequal _to itself_, which is not great.

Modifications:

Defend against the reabstraction thunks by temporarily storing the closures inside a struct, which can be passed to the generic method without the reabstraction thunks.

Adds a helper for comparisons as well to make this more likely to get correct implementations going forward.

Result:

Correct comparisons.

### DIFF
--- a/Sources/NIOSSL/TLSConfiguration.swift
+++ b/Sources/NIOSSL/TLSConfiguration.swift
@@ -521,6 +521,57 @@ public struct TLSConfiguration {
 extension TLSConfiguration: Sendable {}
 
 // MARK: BestEffortHashable
+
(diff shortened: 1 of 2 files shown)


Lukasa:
macOS failures are fixed in #601.

glbrntt (approved):

PR #602, opened 2026-08-31
Accept every c_rehash collision suffix, not just .0

avalyset:
`_isRehashFormat` recognises only a one-digit collision suffix, so `openssl rehash` links from `.10` upward are not seen as c_rehash links at all.

### The three barriers

On `main` the name check is:

```swift
guard let lastPathComponent = utf8PathSplitView.last,
    lastPathComponent.count == 10
else { return false }
let filenameParts = lastPathComponent.split(separator: UInt8(ascii: "."))

guard filenameParts.count == 2,
    let filename = filenameParts.first,
    let fileExtension = filenameParts.last,
    fileExtension.count == 1,
    filename.count == 8,
    filename.allSatisfy({ $0.isHexDigit }),
    fileExtension.first == UInt8(ascii: "0")
else { return false }
```

Three things reject a valid link independently:

1. `lastPathComponent.count == 10` — rejects any name longer than `HHHHHHHH.D` before the extension is examined at all.
2. `fileExtension.count == 1` — rejects any suffix of more than one digit.
3. `fileExtension.first == UInt8(ascii: "0")` — rejects every suffix that does not begin with `0`, which for the ids OpenSSL actually writes means everything except `.0`.

Relaxing any one of them is not enough on its own.

### Consequence

The filter lives in `loadVerifyLocations(_:context:sendCANames:)`, which calls `_isRehashFormat` over the directory contents when `sendCANames` is set, loads each surviving link as a PEM certificate, and passes it to `addCACertificateNameToList` so its CA name reaches the client CA list. (`addRootCertificate` is a different function; it adds a certificate to the trust store and does not walk the directory.)

A CA that lost the hash race is therefore skipped, so its name is absent from the CertificateRequest, and a client holding a certificate from that CA has no reason to send it. The failure is quiet: nothing errors, the handshake just proceeds without the client certificate.

### The change

- `fileExtension.allSatisfy({ $0.isDecimalDigit })` replaces the `== "0"` comparison, mirroring the `isHexDigit` check on the line above. `isDecimalDigit` is added to the same `UTF8.CodeUnit` extension as `isHexDigit`, in the same shape, reusing the existing `asciiZero`/`asciiNine` constants.
- `fileExtension.count == 1` becomes `!fileExtension.isEmpty`. Emptiness still has to be excluded explicitly, because `allSatisfy` is vacuously true on an empty collection.
- `lastPathComponent.count == 10` is removed rather than relaxed to `>= 10`. It is redundant once the parts are checked: `filenameParts.count == 2`, `filename.count == 8` and a non-empty all-digits extension already pin the name to 8 + 1 + digits. A second, looser length check would only be another thing to keep in sync.
- The split passes `omittingEmptySubsequences: false`. This is load-bearing. With the default behaviour `Collection.split` discards empty pieces, so `.7f44456a.0`, `7f44456a..0` and `7f44456a.0.` collapse to the same two parts as `7f44456a.0`. The removed `count == 10` check had been rejecting exactly those three by accident; the explicit split replaces that accident with an intentional rule, and the reasoning is recorded at the call site.

No upper bound on the number of digits is imposed. The reasoning, and the OpenSSL sources it rests on, are in a comment on this PR.

### Tests

`testRehashFormat` already asserts a non-digit extension (`.z`) is rejected; `.a` is added next to it, since a hex-but-not-decimal character is the case most likely to regress if someone reaches for `isHexDigit` here.

`testRehashFormatAcceptsCollisionSuffixes` creates real symlinks and covers:

- accepted: 0, 1, 9, 10, 123, 255.
- rejected: `.7f44456a.0`, `7f44456a..0`, `7f44456a.0.`, `7f44456a.`, `7f44456a.0.1` and `7f44456a.r0`.

Every assertion carries a message saying why that name is in the list. The rejected names are real symlinks to a real certificate, so the name check is the only thing that can reject them. Only the first three depend on the explicit split; the comment in the test says what rejects the other three, so nobody reads the split as decorative.

The existing `getRehashFilename` helper is used unchanged. It formats with `%08lx.%d` and already produces `.10`, `.123` and `.255` correctly.

One existing assertion changes meaning. `testRehashFormat` builds `7f44456a.1` for its "in rehash format, but not a symlink" case, and on `main` that passes for the wrong reason: the name is rejected at the `"0"` comparison, before the symlink check it is meant to exercise. I confirmed this by passing a nonexistent path through the function, where reaching `lstat` is observable as a throw:

| Source | `_isRehashFormat("/etc/ssl/certs/7f44456a.1")` |
| --- | --- |
| `main` | returns `false`, does not throw — rejected at the name check |
| this PR | throws — the name is accepted and execution reaches the symlink check |

After this change the assertion exercises what it was written for.

### Verification

Run in `swift:6.2-noble`, the image the Linux 6.2 leg of the shared matrix names (`apple/swift-nio`, `scripts/generate_matrix.sh` line 78):

```
docker run --rm -v "$PWD":/src -w /src swift:6.2-noble swift test --scratch-path .build-linux
```

| Run | Result |
| --- | --- |
| `apple:main` (03827c1), unmodified | `Executed 339 tests, with 0 failures (0 unexpected)` |
| PR head before this commit, unmodified | `Executed 340 tests, with 0 failures (0 unexpected)` |
| With this change | `Executed 340 tests, with 0 failures (0 unexpected)` |
| With this change, the matrix's 6.2 argument overrides | `Executed 340 tests, with 0 failures (0 unexpected)` |
| New test against `apple:main` source | 5 failures: `.1`, `.9`, `.10`, `.123`, `.255` |
| New test against this branch's first commit | 3 failures: `.10`, `.123`, `.255` |
| New test against a variant keeping the default split | 3 failures: `.7f44456a.0`, `7f44456a..0`, `7f44456a.0.` |

The last three are negative controls, showing the new assertions have teeth in both directions. `swift format lint` reports only the pre-existing `OrderedImports` warnings, unchanged from the baseline.

The `.10` case was reported by @stephenlclarke; I reproduced it here before changing anything.

---

The code in this pull request was written with Claude Fable 5.1. The verification above and this description were written with Claude Opus 5. I have reviewed and tested the change, and I am responsible for its content.

### DIFF
--- a/Sources/NIOSSL/SSLContext.swift
+++ b/Sources/NIOSSL/SSLContext.swift
@@ -782,25 +782,37 @@ extension NIOSSLContext {
     /// Takes a path and determines if the file at this path is of c_rehash format .
     internal static func _isRehashFormat(path: String) throws -> Bool {
         // Check if the element’s name matches the c_rehash symlink name format.
-        // The links created are of the form HHHHHHHH.D, where each H is a hexadecimal character and D is a single decimal digit.
+        // `openssl rehash` writes its links with `snprintf(..., "%08x.%s%d", ...)` (apps/rehash.c):
+        // eight hex characters, a period, a type infix that is empty for certificates and "r" for
+        // CRLs, and a decimal collision id that starts at 0 and counts up for every further
+        // certificate sharing the same subject name hash. The id is a plain %d, not one character,
+        // so ".10" and beyond are ordinary names and must be recognised here.
         let utf8PathView = path.utf8
         let utf8PathSplitView = utf8PathView.split(separator: UInt8(ascii: "/"))
 
-        // Make sure the path is at least 10 units long
-        guard let lastPathComponent = utf8PathSplitView.last,
-            lastPathComponent.count == 10
-        else { return false }
-        // Split into filename parts HHHHHHHH.D -> [[HHHHHHHH], [D]]
-        let filenameParts = lastPathComponent.split(separator: UInt8(ascii: "."))
+        guard let lastPathComponent = utf8PathSplitView.last else { return false }
 
-        // Double check that the extension did not fail to cast to an integer.
-        // Make sure that the filename is an 8 character hex based file name.
+        // Split into filename parts HHHHHHHH.D -> [[HHHHHHHH], [D]].
+        //
+        // The split must keep empty subsequences. With the default (omitting) behaviour a name
+        // such as "7f44456a..0", ".7f44456a.0" or "7f44456a.0." collapses to the same two parts
+        // as "7f44456a.0" and would be accepted, even though it has a period in a position
+        // `openssl rehash` never puts one. Keeping the empty pieces makes every stray period show
+        // up in the part count or as an empty part, so the checks below can reject it.
+        let filenameParts = lastPathComponent.split(
+            separator: UInt8(ascii: "."),
+            omittingEmptySubsequences: false
+        )
+
+        // Exactly one period, an 8 character hex filename before it and at least one decimal digit
+        // after it. No separate length check on the whole name is needed: those three conditions
+        // already pin it to 8 + 1 + digits.
         guard filenameParts.count == 2,
             let filename = filenameParts.first,
             let fileExtension = filenameParts.last,
-            fileExtension.count == 1,
             filename.count == 8,
             filename.allSatisfy({ $0.isHexDigit }),
+            !fileExtension.isEmpty,
             fileExtension.allSatisfy({ $0.isDecimalDigit })
         else { return false }
 
@@ -815,7 +827,8 @@ extension NIOSSLContext {
         #endif
 
         // Return true at this point because the file format is considered to be in rehash format and a symlink.
-        // Rehash format being "%08lx.%d" or HHHHHHHH.D
+        // Rehash format being "%08x.%s%d" with an empty infix, or HHHHHHHH.D where D is one or
+        // more decimal digits.
         return true
     }
 }
--- a/Sources/NIOSSL/SSLContext.swift
+++ b/Sources/NIOSSL/SSLContext.swift
@@ -801,7 +801,7 @@ extension NIOSSLContext {
             fileExtension.count == 1,
             filename.count == 8,
             filename.allSatisfy({ $0.isHexDigit }),
-            fileExtension.first == UInt8(ascii: "0")
+            fileExtension.allSatisfy({ $0.isDecimalDigit })
         else { return false }
 
         // Check if the element is a symlink. If it is not, return false.
@@ -964,7 +964,8 @@ internal class DirectoryContents: Sequence, IteratorProtocol {
     }
 }
 
-// Used as part of the `_isRehashFormat` format to determine if the filename is a hexadecimal filename.
+// Used as part of the `_isRehashFormat` format to determine whether the filename is a hexadecimal
+// filename and whether its extension is a decimal digit.
 extension UTF8.CodeUnit {
     private static let asciiZero = UInt8(ascii: "0")
     private static let asciiNine = UInt8(ascii: "9")
@@ -978,6 +979,15 @@ extension UTF8.CodeUnit {
         case (.asciiZero)...(.asciiNine),
             (.asciiLowercaseA)...(.asciiLowercaseF),
             (.asciiUppercaseA)...(.asciiUppercaseF):
+            return true
+        default:
+            return false
+        }
+    }
+
+    var isDecimalDigit: Bool {
+        switch self {
+        case (.asciiZero)...(.asciiNine):
             return true
         default:
             return false
--- a/Tests/NIOSSLTests/TLSConfigurationTest.swift
+++ b/Tests/NIOSSLTests/TLSConfigurationTest.swift
@@ -1045,10 +1045,16 @@ class TLSConfigurationTest: XCTestCase {
     }
 
     func testRehashFormatAcceptsCollisionSuffixes() throws {
-        // `openssl rehash` names its links "%08lx.%d": when several certificates share a subject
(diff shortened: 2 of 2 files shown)


stephenlclarke:
I found one more edge case while testing this with Xcode. `openssl rehash` can continue beyond `.9`, but the remaining `fileExtension.count == 1` check rejects `.10` and above.

I’ve put a small follow-up on `upstream/rehash-collision-suffixes`, based directly on this PR’s current head:
https://github.com/stephenlclarke/swift-nio-ssl/commit/b27024773ee2852ae7ec9eb6b2f0cc451dc78469

It covers `.10` and `.123`, while continuing to reject empty and malformed suffixes. The focused test passes.

Please feel free to cherry-pick the commit or take the branch. I haven’t opened a competing PR.

avalyset:
Thank you Steve — you are right, and the fix as it stood was incomplete. I reproduced the failure here before changing anything: on the previous head, adding `10` to the suffix list in `testRehashFormatAcceptsCollisionSuffixes` fails with

```
error: XCTAssertTrue failed - /tmp/.../56c899cd.10 is a valid c_rehash link and should be recognised
```

I have pushed a second commit to this branch rather than cherry-picking yours, so that the reasoning and the verification are my own, and I have credited you in the commit message.

First, a correction to my own earlier description of this PR, which I have now edited. I had quoted openssl-rehash(1) as the authority. That was wrong. `doc/man1/openssl-rehash.pod.in` lines 51-52 say `D` is "a single decimal digit", which describes the behaviour this PR removes rather than the behaviour OpenSSL has. The man page and the implementation disagree, so the implementation is the anchor. I had also named the wrong function: the directory walk is in `loadVerifyLocations(_:context:sendCANames:)`, not `addRootCertificate`.

Line numbers below are `openssl/openssl` at `1935d215b5c6bc8619342e6271cfd52837517112` (master, 2026-09-05), except where a release branch is named.

**Upper bound.** For `apps/rehash.c` it is 255. `MAX_COLLISIONS` is 256 (line 48), `add_entry` refuses a new entry once a bucket holds that many (line 158, `if (bp->num_needed >= MAX_COLLISIONS) {`), and `do_dir` allocates the next id out of a bitmask sized `MAX_COLLISIONS / 8` (line 362), scanning upward at lines 442-443:

```c
                    while (bit_isset(idmask, nextid))
                        nextid++;
```

That is not the only writer in the field. The older perl `tools/c_rehash.in` is still shipped on the 3.0 and 3.5 branches, and its suffix loop has no cap at all — lines 231-243 of `tools/c_rehash.in` on `openssl-3.5` (identical on `openssl-3.0`):

```perl
    my $suffix = 0;
    # Search for an unused hash filename
    my $crlmark = $is_cert ? "" : "r";
    while(exists $hashlist{"$hash.$crlmark$suffix"}) {
```

with `$suffix++` at line 241 and nothing bounding it. That is why I did not put a ceiling in the predicate: a hard limit of 255 would be correct for `openssl rehash` and wrong for a directory that `c_rehash` populated. Accepting any number of digits is the only rule that covers both.

**Leading zeros.** OpenSSL does not write them; all three link-name calls use `%d`. Each spans more than one line — lines 435-436:

```c
                    snprintf(buf, buflen, "%08x.%s%d", bp->hash,
                        suffixes[bp->type], ep->old_id);
```

and lines 445-447 (a new link) and 467-469 (a link being removed, with `ep->old_id` in place of `nextid`) have the same shape with the directory prefixed:

```c
                    snprintf(buf, buflen, "%s%s%08x.%s%d",
                        dirname, pathsep, bp->hash,
                        suffixes[bp->type], nextid);
```

On the read side, lines 220-222 are the whole test:

```c
    id = strtoul(&filename[i], &endptr, 10);
    if (*endptr != '\0')
        return -1;
```

The only thing rejected there is a non-digit remainder. `.010` has none, so OpenSSL's own reader takes it. An all-digits check therefore matches what OpenSSL reads and is a strict superset of what it writes. I did not add a no-leading-zero rule, which would be stricter than OpenSSL itself.

**The `.rD` suffix stays rejected.** Line 104 is `static const char *suffixes[] = { "", "r" };` — an empty infix for certificates, `r` for CRLs — so `HHHHHHHH.rD` is a CRL link, described at lines 57-58 of `openssl-rehash.pod.in`. This is worth an explicit negative case because of where the accepted names go. In `loadVerifyLocations`, every name this predicate accepts is passed to `NIOSSLCertificate(_file:format: .pem)` at line 725, inside the loop, with no `catch` around it, and `loadVerifyLocations` itself is `throws`. Whatever that initialiser does with a CRL file, the result is not a skipped entry: it either yields a certificate or propagates out of the whole call. I added `7f44456a.r0` so the predicate keeps that name out deliberately rather than by accident.

One measurement worth recording, since we both landed on `omittingEmptySubsequences: false` and it is easy for a later reader to take it for a style choice. I built a variant that drops the length checks but keeps the default split, and ran the suite against it: `.7f44456a.0`, `7f44456a..0` and `7f44456a.0.` are all accepted. The `count == 10` check had been rejecting those by accident. It is exactly those three and no others — `7f44456a.` is caught by the `!isEmpty` check, `7f44456a.0.1` yields three parts under either behaviour, and `7f44456a.r0` fails `isDecimalDigit`. The test comment now says so, so the split is not mistaken for belt-and-braces.

I also dropped `lastPathComponent.count` entirely rather than relaxing it to `>= 10`. Once `filenameParts.count == 2`, `filename.count == 8` and a non-empty all-digits extension all hold, the name is already pinned to 8 + 1 + digits, so the length check has nothing left to add.

Full suite in `swift:6.2-noble`: 340 tests, 0 failures, before and after, and also with the matrix's 6.2 argument overrides.

---

The code in this pull request was written with Claude Fable 5.1, and this comment with Claude Opus 5. I have reviewed and tested the change, and I am responsible for its content.

Lukasa (approved):
Very nice change, thank you!

PR #603, opened 2026-09-02
Close the channel when input closes during the TLS handshake

Sam-Amin:
### Motivation

`NIOSSLHandler` treats input closure during the handshake as a fatal handshake
failure: `userInboundInputClosedTriggered` synthesises
`NIOSSLError.handshakeFailed(.sslError([.eofDuringHandshake]))`, fires it, and
forwards `ChannelEvent.inputClosed`. Unlike every other fatal handshake failure
it then returns without closing the channel and without leaving `.handshaking`.

When the channel has `allowRemoteHalfClosure` set, nothing else closes it
either — NIO deliberately does not close on FIN, and a pipeline still waiting on
ALPN may contain no handler that owns the closure. The descriptor stays in
`CLOSE_WAIT` for the lifetime of the process, and a graceful shutdown that waits
on all child channels never completes.

Compare `doHandshakeStep`'s `.failed` branch, which fires the same
`handshakeFailed` error and then calls `channelClose`. Both are fatal; only one
closed.

A related hole sits underneath. `channelClose` and `channelInactive` complete
`shutdownPromise` and `closePromise` but never `closeOutputPromise`. A
`close(mode: .output)` issued while handshaking is buffered with its promise
stored there, and `discardBufferedActions` only fails `.write` promises, so that
promise was never completed: the caller hangs, or trips NIO's leaked-promise
`fatalError`. Closing the channel on this path makes it reliably reachable.

### Modifications

- In `userInboundInputClosedTriggered`, mark `.handshaking` and
  `.additionalVerification` as fatal and call `channelClose` after firing the
  error and forwarding `inputClosed`. The post-handshake ragged-EOF branch is
  unchanged: half closure is legitimate once the connection is up.
- Complete `closeOutputPromise` in `channelClose` and `channelInactive`,
  alongside the promises they already fail.
- Added tests for input closure during the handshake (`EmbeddedChannel` and real
  sockets), during additional peer certificate verification — including
  completing the verification callback after the close — and for the buffered
  `close(mode: .output)` promise. A control test asserts that input closure
  after the handshake still fires `uncleanShutdown`, forwards `inputClosed`, and
  leaves the channel open.

### Result

A peer that connects and half-closes before completing the handshake no longer
leaks a descriptor. Post-handshake half closure is unaffected.

Two observable changes for callers on the mid-handshake path: buffered writes
now fail with `ChannelError.ioOnClosedChannel` rather than the TLS error, and
`errorCaught` fires once rather than twice. Both match the existing
`doHandshakeStep(.failed)` behaviour.

### DIFF
--- a/Sources/NIOSSL/NIOSSLHandler.swift
+++ b/Sources/NIOSSL/NIOSSLHandler.swift
@@ -158,10 +158,13 @@ public class NIOSSLHandler: ChannelInboundHandler, ChannelOutboundHandler, Remov
         }
         let shutdownPromise = self.shutdownPromise
         self.shutdownPromise = nil
+        let closeOutputPromise = self.closeOutputPromise
+        self.closeOutputPromise = nil
         let closePromise = self.closePromise
         self.closePromise = nil
 
         shutdownPromise?.fail(channelError)
+        closeOutputPromise?.fail(channelError)
         closePromise?.fail(channelError)
         context.fireErrorCaught(channelError)
         discardBufferedActions(reason: channelError)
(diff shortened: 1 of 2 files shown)


Lukasa (approved):
Very nice, thanks Sam.

PR #604, opened 2026-09-05
fix: close before forwarding fatal input closure

stephenlclarke:
## Motivation

Follow up #603 by making its fatal input-closure path safe against synchronous downstream reentrancy.

While an additional certificate verification future is pending, `userInboundInputClosedTriggered` forwards `ChannelEvent.inputClosed` before `channelClose` changes the handler state. A downstream handler can synchronously complete that future from the callback, re-enter the handshake while the state still says `.additionalVerification`, and attempt to reactivate a connection whose input has already closed.

## Modifications

- Mark fatal handshake input closure as `.closed` before firing downstream error and input-closed events.
- Strengthen the existing additional-verification regression so a downstream input-closed handler completes the verification future synchronously.
- Preserve the post-handshake ragged-EOF and remote-half-closure behavior from #603.

## Testing

`swift test --filter NIOSSLIntegrationTest.testInputClosedDuringAdditionalVerificationClosesTheChannel`

Result: 1 test passed. The commit is signed and based directly on the merge commit for #603 (`03827c1a9fdb2b6b00a4e93ede8861520263af8c`).

### DIFF
--- a/Sources/NIOSSL/NIOSSLHandler.swift
+++ b/Sources/NIOSSL/NIOSSLHandler.swift
@@ -249,6 +249,13 @@ public class NIOSSLHandler: ChannelInboundHandler, ChannelOutboundHandler, Remov
             fatal = false
         }
 
+        if fatal {
+            // Mark the state terminal before invoking downstream handlers. Those handlers may
+            // synchronously complete an outstanding additional-verification future, which must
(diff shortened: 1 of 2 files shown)


Lukasa (approved):
Thanks for this!

PR #605, opened 2026-09-05
fix: send bad certificate alert for Darwin trust failures

stephenlclarke:
## Motivation

Darwin's default Security.framework trust verifier reports a rejected chain to BoringSSL through the custom verification callback. The callback returned `ssl_verify_invalid` without setting `out_alert`, leaving the alert byte unspecified instead of sending the standard `bad_certificate` alert.

Public custom verification callbacks must retain BoringSSL's existing default alert selection, while the internal Darwin trust manager should provide an explicit, deterministic alert.

## Modifications

- Pass BoringSSL's `out_alert` pointer through to `CustomVerifyManager.process`.
- Allow only internal verification callbacks to configure a fixed failure alert.
- Configure the Darwin default trust manager with `SSL_AD_BAD_CERTIFICATE` through a small C shim.
- Add focused coverage for internal failure, public callback compatibility, success, pending verification, and Darwin default configuration.

## Testing

`swift test --filter CustomVerifyManagerTests`

Result: 6 tests passed. No public API is changed; public verification callbacks continue to leave BoringSSL's alert selection untouched.

### DIFF
--- a/Sources/NIOSSL/SSLCallbacks.swift
+++ b/Sources/NIOSSL/SSLCallbacks.swift
@@ -452,20 +452,29 @@ internal struct CustomVerifyManager {
 
     private var result: PendingResult = .notStarted
 
+    /// A fixed TLS alert for an internal verifier's rejected result.
+    ///
+    /// Public verification callbacks deliberately leave this unset so BoringSSL retains its
+    /// documented default alert selection.
+    let failureAlert: UInt8?
+
     /// Contains the metadata that the callback returned. As such, this property will *only* contain a value if
     /// `self.result` is `.complete` (and if the callback promise returns metadata).
     var verificationMetadata: VerificationMetadata?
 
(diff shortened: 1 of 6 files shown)


Sources/NIOSSL/SSLCallbacks.swift
```diff
@@ -486,7 +495,10 @@ extension CustomVerifyManager {
 }
 
 extension CustomVerifyManager {
-    mutating func process(on connection: SSLConnection) -> ssl_verify_result_t {
+    mutating func process(
+        on connection: SSLConnection,
+        outAlert: UnsafeMutablePointer<UInt8>?
```
Lukasa, line 500: Let's avoid passing a pointer in here: it moves memory unsafety around in an unfortunate way. Instead, let's amend the return value here to return a tuple and have the second value be the optional we need.

